PR #14, opened 2017-10-23
 Fix dead links, add small extra info, change other small things

Danktuary:
Some last minute changes to fix dead links and spruce things up a bit.


Lewdcario:
channe

PR #21, opened 2017-10-25
Sweep

Drahcirius:
Use em dash over regular dash.
Changed some places where `can` makes more sense than `may`.
`i.e.` -> `e.g.` where appropriate
Removed a section.
Minor rewording.


guide/command-handling/dynamic-commands.md
```diff
@@ -55,7 +55,7 @@ catch (error) {
 }
 ```
 
-If there isn't a command with that name, `return`. If there is, `.get()` the command and call it `.execute()` method while passing in our `message` and `args` variables as the method arguments. In case something goes wrong, log the error and report back to the member to let them know.
+If there isn't a command with that name, we exit early. If there is, `.get()` the command and call its `.execute()` method while passing in our `message` and `args` variables as the method arguments. In case something goes wrong, log the error and report back to the member to let them know.
```
Lewdcario: we exit early -> you exit early
our -> your

guide/creating-your-bot/README.md
```diff
@@ -12,7 +12,7 @@ It's suggested that you save the file as `index.js`, but you may name it whateve
 
 ### Logging in to Discord
 
-Once you've created a new file, do a quick check to see if you have everything setup properly. Copy & paste the following code into your file and save it again. Don't worry if you don't understand it right away - it'll be explained a bit more in depth after this.
+Once you've created a new file, do a quick check to see if you have everything setup properly. Copy & paste the following code into your file and save it. Don't worry if you don't understand it right away—it'll be explained a bit more in depth after this.
```
Lewdcario: space after `away-` i think

guide/creating-your-bot/configuration-files.md
```diff
@@ -46,37 +46,3 @@ As previously mentioned, you'll probably want to store more than just your token
...
-// index.js
-const responseArray = require('./random-responses.json');
-console.log(responseArray); // ["Hi!", "Hey!", "What's up?", "How are you?"]
-```
-
-This makes our code a bit more clean and easy to understand.
-
-## Resulting code
-
-If you want to compare your code to the code we've constructed so far, you can review it over on the GitHub repository [here](https://github.com/Danktuary/Making-Bots-with-Discord.js/tree/master/code_samples/creating-your-bot/configuration-files).
```
Danktuary: The "Resulting code" bit should stay

PR #31, opened 2017-11-01
Declarative imports!

aemino:
Declarative imports are a good idea.  They make your code good and declarative.


Danktuary:
I think you might've misunderstood what I explained before, or I might've not worded it perfectly. This guide is meant to start off very "new to Node/ES6 but not entirely new to JavaScript" friendly, and using destructuring right off the bat is something I want to avoid, because it may confuse newcomers.

Take a look at this part of the guide, for example:


That'd be a good entry point to start using destructuring. However, in places where it only uses `new Discord.Client()`, I don't see a huge benefit of using destructuring in such places. If someone else does, that's fine by me, but that's just my opinion.

PR #33, opened 2017-11-02
 added Async/Await examples and explaination also added difference between Users and GuildMembers in faq

Dev-Yukine:
Added Async/Await and Promise explaination and examples aswell as the difference between Users and GuildMembers in faq (you maybe want to look over the grammar 👀 )

### DIFF
--- a/guide/additional-info/async-await.md
+++ b/guide/additional-info/async-await.md
@@ -116,7 +116,7 @@ client.on('message', async message => {
 			await message.react('🇨');
 		} 
 		catch(error) {
-			//handle failure of any Promise rejection inside here
+			// handle failure of any Promise rejection inside here
 		}
 	}
 });
@@ -134,7 +134,7 @@ client.on('message', message => {
 		message.channel.send('this message will be delete')
 		.then(sentMessage => sentMessage.delete(10000))
 		.catch(error => {
-			//handle error
+			// handle error
 		});
 	}
 });
@@ -149,7 +149,7 @@ client.on('message', async message => {
 			const sentMessage = await message.channel.send('this message will be delete');
 			sentMessage.delete(10000);
 		}catch(error){
-			//handle error
+			// handle error
 		}
 	}
 });
--- a/guide/additional-info/async-await.md
+++ b/guide/additional-info/async-await.md
@@ -1,45 +1,43 @@
 # Understanding async/await
 
-If you aren't very familiar with EcmaScript 2017, you may not know about async/await. It's a useful way to handle Promises in a synchronous way instead stacking .then() callback. Also, it looks cleaner.
+If you aren't very familiar with ECMAScript 2017, you may not know about async/await. It's a useful way to handle Promises in a synchronous manner instead stacking ".then()" callbacks. Also, it looks cleaner and is increases readability overall.
 
 ## How do Promises work?
 
-Before we can get into async/await, you should know what Promises are and how they work because async/await is just a way to handle Promises (if you know what Promises are and how to deal with them you can skip this part). Promises are a way to handle asynchronous tasks in Javascript, they are the newer alternative to callbacks. A Promise has a lot of simularities to a progress bar; Promises represent an ongoing process that has not yet finished. A good example for that is a request to a server (e.g discord.js sends requests to Discord's API).
+Before we can get into async/await, you should know what Promises are and how they work, because async/await is just a way to handle Promises. if you know "what Promises are and how to deal with them" you can skip this part. 
 
-A Promise can have 3 states:
+Promises are a way to handle asynchronous tasks in Javascript; they are the newer alternative to callbacks. A Promise has a lot of similarities to a progress bar; Promises represent an ongoing process that has not yet finished. A good example for that is a request to a server (e.g discord.js sends requests to Discord's API).
 
-* pending
-* fulfilled
-* rejected
+A Promise can have 3 states pending, fulfilled, and rejected
 
-The state, **pending**, means that the Promise still is ongoing and nether fulfilled nor rejected.
-The state, **fulfilled**, means that the Promise is done and was executed without any errors.
-The state, **rejected**, means that the Promise encountered an error and could not be executed correctly.
+The state **pending** means that the Promise still is ongoing and nether fulfilled nor rejected.
+The state **resolved** means that the Promise is done and was executed without any errors.
+The state **rejected** means that the Promise encountered an error and could not be executed correctly.
 
-One important thing to know is that a Promise can only have 1 state; it can never be pending and fulfilled, rejected and fulfilled, or pending and rejected. You may be asking, "How would that look in code?". Here is a small example:
-(ES6 code is being used in this example; if you do not know what that is, you should read up on that [here](/additional-info/es6-syntax).)
+One important thing to know is that a Promise can only have one state at a time; it can never be pending and fulfilled, rejected and fulfilled, or pending and rejected. You may be asking, "How would that look in code?". Here is a small example:
+<p class="tip">ES6 code is being used in this example; if you do not know what that is, you should read up on that [here](/additional-info/es6-syntax).</p>
 
 ```js
 function asyncTask() {
 	return Promise((resolve, reject) => {
 		setTimeout(resolve("Task is done"), 2000)
-	})
+	});
 }
 
 asyncTask().then(value => {
-	// asyncTask is complete and has not encountered any errors
-	// The resolved value will be the String "Task is done"
+	// "asyncTask" is complete and has not encountered any errors
+	// the resolved value will be the string "Task is done"
 }).catch(error => {
-	// asyncTask encountered an error
+	// "asyncTask" encountered an error
 	// the error will be an Error Object
-})
+});
 ```
 
-In this scenario the asyncTask returns a Promise and we attach a .then() function and a .catch() function to it. The .then() function will trigger if the Promise was fulfilled and the .catch() function if the Promise was rejected. But with our function we resolve the Promise after 2 seconds with the String "Task is Done", so the .catch() function will never be executed.
+In this scenario the "asyncTask" returns a Promise and we attach a ".then()" function and a ".catch()" function to it. The ".then()" function will trigger if the Promise was fulfilled and the ".catch()" function if the Promise was rejected. But with our function we resolve the Promise after 2 seconds with the String "Task is Done", so the .catch() function will never be executed.
 
 ## How to implement async/await
 
-After knowing how Promises works and what they are for, let's look at an example in which we handle multiple Promises. Let's say you wanted to react to a message in a certain order with letters (regional indicators). For this example, I will take the basic template for a discord.js bot with some es6 adjustments.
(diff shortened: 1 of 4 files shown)


guide/additional-info/async-await.md
```diff
@@ -0,0 +1,158 @@
...
+client.on('message', message => {
+	if (message.content === `${prefix}react`) {
+		message.react('🇦');
+		message.react('🇧');
+		message.react('🇨');
+	}
+});
+```
+
+But since all of these react methods are started at the same time, it would just be a race which server request finished first, so there would be no guarantee that it would react in the order you wanted it to. In order to make sure it reacts in order (a, b, c), we need to use the .then() callback from the Promises that these methods return. As a result the code we want would mostly look like this:
```
Danktuary: * "... a race which server request finished first ..." -> "... a race to which server request finished first ...".

guide/additional-info/async-await.md
```diff
@@ -0,0 +1,158 @@
...
+				message.react('🇨').catch(error => {
+				//handle failure of the third react
+			})
+		}).catch(error => {
+			//handle failure of the second react
+		})
+	}).catch(error => {
+		//handle failure of the first react
+	})
+	}
```
Danktuary: Indentation here is a bit messed up.

guide/additional-info/async-await.md
```diff
@@ -0,0 +1,158 @@
...
+client.on('message', async message => {
+	if (message.content === `${prefix}react`) {
+		await message.react('🇦');
+		await message.react('🇧');
+		await message.react('🇨');
+	}
+});
+```
+
+That would mostly be the same code with async/await (note that you can only use the await keyword inside an async function!), but how do we catch Promise rejections now since we won't use .catch() anymore? That is also a useful feature with async/await; the error will be thrown if you await it so you can just wrap the awaited Promises inside a try/catch and you are 👌.
```
Danktuary: "... and you are 👌" -> "... and you're good to go".

guide/additional-info/async-await.md
```diff
@@ -0,0 +1,158 @@
...
+That would mostly be the same code with async/await (note that you can only use the await keyword inside an async function!), but how do we catch Promise rejections now since we won't use .catch() anymore? That is also a useful feature with async/await; the error will be thrown if you await it so you can just wrap the awaited Promises inside a try/catch and you are 👌. 
+
+```js
+client.on('message', async message => {
+	if (message.content === `${prefix}react`) {
+		try{
+			await message.react('🇦');
+			await message.react('🇧');
+			await message.react('🇨');
+		}catch(error) {
```
Danktuary: Needs spaces after the `try` braces, and `catch(error)` needs to go on its own line, for brace-style consistency with the rest of the guide.

guide/additional-info/async-await.md
```diff
@@ -0,0 +1,158 @@
...
+			//handle failure of any Promise rejection inside here
+		}
+	}
+});
+```
+
+A nice side effect is that you don't have redundant catch blocks now since you use a multi catch because you wrapped all Promises inside the try block. 
+
+So you may be asking, "How would I get the value the Promise resolved with?" Well, let's look at an example where you want to delete a sent message.
```
Danktuary: Remove the comma, add a period after the quoted question.

guide/additional-info/async-await.md
```diff
@@ -0,0 +1,158 @@
...
+		message.channel.send('this message will be delete')
+		.then(sentMessage => sentMessage.delete(10000))
+		.catch(error => {
+			//handle error
+		});
+	}
+});
+```
+
+The return value of a .send() is a Promise what resolves with the sent Message object, but how would the same code with Async/Await look like?
```
Danktuary: Uncapitalize Async/Await.

guide/creating-your-bot/common-questions.md
```diff
@@ -99,4 +99,8 @@ user.send('<content>');
 });
 ```
 
+### What is the difference between a User and a GuildMember?
+
+A lot of users get confused what is the difference between a User and a GuildMember the simple answer is that an User represent a Discord User Global and a GuildMember represent a Discord User on a specific server that means only GuildMembers can have Permissions, Roles and Nicknames as example because all these things are server bound information that could be different on every server that User is in.
```
Danktuary: * Needs a period before "the simple answer".
* "a Discord User Global" should be "a global Discord user"
* Uncapitalize "User" in "... represent a Discord User".
* Needs a period before "the means only GuildMembers ...".
* Uncapitalize "Permissions, Roles and Nicknames" since we're not talking about the classes but rather the actual nouns, and add a comma before the "and".
* "... as example" -> "... , for example, ..."
* "server bound" -> "server-bound"
* Uncapitalize "User" in that last sentence".

Danktuary: Another thing is that this is placed awkwardly; it's right in-between the `<p class="tip">` tag and text related to `.awaitMessages()`. This should be placed after that.

guide/additional-info/async-await.md
```diff
@@ -0,0 +1,163 @@
+# Understanding async/await
+
+If you aren't very familiar with ECMAScript 2017, you may not know about async/await. It's a useful way to handle Promises in a synchronous manner instead stacking `.then()` callbacks. Also, it looks cleaner and is increases readability overall.
```
Danktuary: Typo at "is increases readability" (should be "it").

guide/additional-info/async-await.md
```diff
@@ -0,0 +1,163 @@
...
+
+## How do Promises work?
+
+Before we can get into async/await, you should know what Promises are and how they work, because async/await is just a way to handle Promises. if you know "what Promises are and how to deal with them" you can skip this part.
```
Danktuary: Second sentence doesn't need the quotes, and the beginning of the sentence needs to be capitalized. A comma before "you can skip this part" would make it less awkward to read, I think.

guide/additional-info/async-await.md
```diff
@@ -0,0 +1,163 @@
...
+
+If you aren't very familiar with ECMAScript 2017, you may not know about async/await. It's a useful way to handle Promises in a synchronous manner instead stacking `.then()` callbacks. Also, it looks cleaner and is increases readability overall.
+
+## How do Promises work?
+
+Before we can get into async/await, you should know what Promises are and how they work, because async/await is just a way to handle Promises. if you know "what Promises are and how to deal with them" you can skip this part. 
+
+Promises are a way to handle asynchronous tasks in Javascript; they are the newer alternative to callbacks. A Promise has a lot of similarities to a progress bar; Promises represent an ongoing process that has not yet finished. A good example for that is a request to a server (e.g discord.js sends requests to Discord's API).
+
+A Promise can have 3 states pending, fulfilled, and rejected
```
Danktuary: Needs a semicolon before listing the 3 states.

guide/additional-info/async-await.md
```diff
@@ -0,0 +1,163 @@
...
+
+The state **pending** means that the Promise still is ongoing and nether fulfilled nor rejected.
+The state **resolved** means that the Promise is done and was executed without any errors.
+The state **rejected** means that the Promise encountered an error and could not be executed correctly.
```
Danktuary: "The **{state-name}** state", not "The state **{state-name}**".

guide/additional-info/async-await.md
```diff
@@ -0,0 +1,163 @@
...
+
+One important thing to know is that a Promise can only have one state at a time; it can never be pending and fulfilled, rejected and fulfilled, or pending and rejected. You may be asking, "How would that look in code?". Here is a small example:
+<p class="tip">ES6 code is being used in this example; if you do not know what that is, you should read up on that [here](/additional-info/es6-syntax).</p>
```
Danktuary: * Remove the comma after "You may be asking" 
* Add a linebreak before the `<p>` tag, for guide-wide consistency.
* I think a period instead of a semicolon (inside the `<p>` tag text) would read better.

guide/additional-info/async-await.md
```diff
@@ -0,0 +1,163 @@
...
+
+```js
+function asyncTask() {
+	return Promise((resolve, reject) => {
```
Danktuary: Missing the `new` before `Promise`; this is currently returning `TypeError: undefined is not a promise`. However, you're also not using `reject` anywhere in here, which I feel like makes this a bit redundant.

Mostly just a suggestion, but why not try to create a real world example instead of pseudo code? I've noticed that real world examples help people learn more instead of just giving them the syntax and hoping for the best.

guide/additional-info/async-await.md
```diff
@@ -0,0 +1,163 @@
...
+asyncTask().then(value => {
+	// `asyncTask` is complete and has not encountered any errors
+	// the resolved value will be the string "Task is done"
+}).catch(error => {
+	// `asyncTask` encountered an error
+	// the error will be an Error Object
+});
+```
+
+In this scenario the "asyncTask" returns a Promise and we attach a `.then()` function and a `.catch()` function to it. The `.then()` function will trigger if the Promise was fulfilled and the `.catch()` function if the Promise was rejected. But with our function we resolve the Promise after 2 seconds with the String "Task is Done", so the `.catch()` function will never be executed.
```
Danktuary: Quotes should be backticks. The first sentence sounds a bit weird to me though. Maybe just remove the part after "`asyncTask` returns a Promise", since you talk about `.then()` and `.catch()` right after.

guide/additional-info/async-await.md
```diff
@@ -0,0 +1,163 @@
...
+client.on('message', message => {
+	if (message.content === `${prefix}react`) {
+		// code inside here
+	}
+});
+
+client.login('tokeninhere');
+```
+
+So now we need to put the code in. If you don't know how NODE.JS asynchronous execution works, you would probably try something like this:
```
Danktuary: "Node.js", not "NODE.JS"

guide/additional-info/async-await.md
```diff
@@ -0,0 +1,162 @@
...
+		catch(error) {
+			// handle failure of any Promise rejection inside here
+		}
+	}
+});
+```
+
+A nice side effect is that you don't have redundant catch blocks now since you use a multi catch because you wrapped all Promises inside the try block. 
+
+So you may be asking, "How would I get the value the Promise resolved with?"
```
Danktuary: Missing a period.

guide/additional-info/async-await.md
```diff
@@ -0,0 +1,162 @@
...
+		message.channel.send('this message will be delete')
+			.then(sentMessage => sentMessage.delete(10000))
+			.catch(error => {
+				// handle error
+			});
+	}
+});
+```
+
+The return value of a .send() is a Promise what resolves with the sent Message object, but how would the same code with async/await look like?
```
Danktuary: Missing backticks at `.send()`.

guide/additional-info/async-await.md
```diff
@@ -0,0 +1,162 @@
...
+
+```js
+client.on('message', async message => {
+	if (message.content === `${prefix}delete`) {
+		try{
```
Danktuary: Needs spacing between the `try` and the brace.

PR #39, opened 2017-11-07
added theory information to async/await aswell as replaced the words function in combination with .then() and .catch() because that are methods

Dev-Yukine:
yea pretty much an enhancement to the async/await page (you want to look over my grammar again btw)

### DIFF
--- a/guide/additional-info/async-await.md
+++ b/guide/additional-info/async-await.md
@@ -39,8 +39,8 @@ In this scenario the `deleteMessages` function returns a Promise. The `.then()` 
 
 ## How to implement async/await
 
-### Theory information
-The following Information are important to know before starting working with async/await. You can only use the await keyword inside a function what is marked as async (you call the async keyword before the function name, function keyword or the parameter when you use an arrow function) a simple example would be 
+### Theory
+The following Information are important to know before starting working with async/await. You can only use the await keyword inside a function that is declared as async (you put the async keyword before the  function keyword or the parameter when you use an arrow function) a simple example would be:
 ```js
 async function markedAsAsync() {
 	// code
@@ -52,11 +52,11 @@ async () => {
 	// code
 }
 ```
-also an important thing to know is that a function marked as async will always return a Promise. In addition to this if you return something inside a async function this value will be the resolved value and if you throw an error inside an async function it will reject this Promise with the error object. 
+Also an important thing to know, is that a function declared as async will always return a Promise. In addition to this, if you return something inside a async function the returned promise will resolve with that value and if you throw an error inside an async function it will reject the Promise with the error object. 
 
 
 ### Execution with discord.js code
-After knowing how Promises works and what they are for aswell as about the theory information, let's look at an example in which we handle multiple Promises. Let's say you wanted to react to a message in a certain order with letters (regional indicators). For this example, I will take the basic template for a discord.js bot with some ES6 adjustments.
+After knowing how Promises works and what they are for as well as about the theory, let's look at an example in which we'll handle multiple Promises. Let's say you want to react with letters (regional indicators) in a certain order. For this example, you will take the basic template for a discord.js bot with some ES6 adjustments.
 
 ```js
 const Discord = require('discord.js');
--- a/guide/additional-info/async-await.md
+++ b/guide/additional-info/async-await.md
@@ -40,15 +40,19 @@ In this scenario, the `deleteMessages` function returns a Promise. The `.then()`
 ## How to implement async/await
 
 ### Theory
-The following information is important to know before working with async/await. You can only use the `await` keyword inside a function that is declared as `async` (you put the async keyword before the function keyword or before the parameters when using a callback function). 
+
+The following information is important to know before working with async/await. You can only use the `await` keyword inside a function that is declared as `async` (you put the `async` keyword before the `function` keyword or before the parameters when using a callback function). 
 
 A simple example would be:
+
 ```js
 async function declaredAsAsync() {
 	// code
 }
-``` 
-or 
+```
+
+or
+
 ```js 
 const declaredAsAsync = async () => {
 	// code
@@ -63,9 +67,10 @@ client.on('event', async (first, last) => {
 })
 ```
 
-An important thing to know is that a function declared as async will always return a Promise. In addition to this, if you return something, the Promise will resolve with that value, and if you throw an error, it will reject the Promise with that error.
+An important thing to know is that a function declared as `async` will always return a Promise. In addition to this, if you return something, the Promise will resolve with that value, and if you throw an error, it will reject the Promise with that error.
 
 ### Execution with discord.js code
+
 After knowing how Promises work and what they are for, as well as about the theory, let's look at an example in which we'll handle multiple Promises. Let's say you want to react with letters (regional indicators) in a certain order. For this example, you will take the basic template for a discord.js bot with some ES6 adjustments.
 
 ```js
--- a/guide/additional-info/async-await.md
+++ b/guide/additional-info/async-await.md
@@ -40,7 +40,7 @@ In this scenario the `deleteMessages` function returns a Promise. The `.then()` 
 ## How to implement async/await
 
 ### Theory
-The following Information are important to know before starting working with async/await. You can only use the await keyword inside a function that is declared as async (you put the async keyword before the function keyword or the parameter when you use an arrow function) a simple example would be:
+The following information are important to know before starting working with async/await. You can only use the await keyword inside a function that is declared as async (you put the async keyword before the function keyword or the parameter when you use an arrow function) a simple example would be:
 ```js
 async function markedAsAsync() {
 	// code
@@ -52,11 +52,11 @@ async () => {
 	// code
 }
 ```
-Also an important thing to know, is that a function declared as async will always return a Promise. In addition to this, if you return something the Promise will resolve with that value and if you throw an error it will reject the Promise with this error. 
+Also an important thing to know, is that a function declared as async will always return a Promise. In addition to this, if you return something the Promise will resolve with that value and if you throw an error it will reject the Promise with that error. 
 
 
 ### Execution with discord.js code
-After knowing how Promises works and what they are for as well as about the theory, let's look at an example in which we'll handle multiple Promises. Let's say you want to react with letters (regional indicators) in a certain order. For this example, you will take the basic template for a discord.js bot with some ES6 adjustments.
+After knowing how Promises work and what they are for, as well as about the theory, let's look at an example in which we'll handle multiple Promises. Let's say you want to react with letters (regional indicators) in a certain order. For this example, you will take the basic template for a discord.js bot with some ES6 adjustments.
 
 ```js
 const Discord = require('discord.js');
--- a/guide/additional-info/async-await.md
+++ b/guide/additional-info/async-await.md
@@ -35,11 +35,28 @@ deleteMessages(5).then(value => {
 });
 ```
 
-In this scenario the `deleteMessages` function returns a Promise. The `.then()` function will trigger if the Promise was resolved and the `.catch()` function if the Promise was rejected. But with our function we resolve the Promise after 2 seconds with the String "Everything is fine", so the `.catch()` function will never be executed.
+In this scenario the `deleteMessages` function returns a Promise. The `.then()` method will trigger if the Promise was resolved and the `.catch()` method if the Promise was rejected. But with our function we resolve the Promise after 2 seconds with the String "Everything is fine", so the `.catch()` method will never be executed.
 
 ## How to implement async/await
 
-After knowing how Promises works and what they are for, let's look at an example in which we handle multiple Promises. Let's say you wanted to react to a message in a certain order with letters (regional indicators). For this example, I will take the basic template for a discord.js bot with some ES6 adjustments.
+### Theory information
+The following Information are important to know before starting working with async/await. You can only use the await keyword inside a function what is marked as async (you call the async keyword before the function name, function keyword or the parameter when you use an arrow function) a simple example would be 
+```js
+async function markedAsAsync() {
+	// code
+}
+``` 
+or 
+```js 
+async () => {
+	// code
+}
+```
+also an important thing to know is that a function marked as async will always return a Promise. In addition to this if you return something inside a async function this value will be the resolved value and if you throw an error inside an async function it will reject this Promise with the error object. 
+
+
+### Execution with discord.js code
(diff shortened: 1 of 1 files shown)


guide/additional-info/async-await.md
```diff
@@ -35,11 +35,28 @@ deleteMessages(5).then(value => {
...
+```js 
+async () => {
+	// code
+}
+```
+also an important thing to know is that a function marked as async will always return a Promise. In addition to this if you return something inside a async function this value will be the resolved value and if you throw an error inside an async function it will reject this Promise with the error object. 
+
+
+### Execution with discord.js code
+After knowing how Promises works and what they are for aswell as about the theory information, let's look at an example in which we handle multiple Promises. Let's say you wanted to react to a message in a certain order with letters (regional indicators). For this example, I will take the basic template for a discord.js bot with some ES6 adjustments.
```
SpaceEEC: - `aswell` -> `as well`
- same here, `information` in `theory information` feels a bit redundant
- should be `we will handle` (or `we'll handle`)
- `you want to react`
- `to react with letters (regional indicators) in a certain order`, feels nicer this way
- `I will` -> `you will`. (For examples second person, for general stuff third person.)

guide/additional-info/async-await.md
```diff
@@ -35,11 +35,28 @@ deleteMessages(5).then(value => {
 });
 ```
 
-In this scenario the `deleteMessages` function returns a Promise. The `.then()` function will trigger if the Promise was resolved and the `.catch()` function if the Promise was rejected. But with our function we resolve the Promise after 2 seconds with the String "Everything is fine", so the `.catch()` function will never be executed.
+In this scenario the `deleteMessages` function returns a Promise. The `.then()` method will trigger if the Promise was resolved and the `.catch()` method if the Promise was rejected. But with our function we resolve the Promise after 2 seconds with the String "Everything is fine", so the `.catch()` method will never be executed.
 
 ## How to implement async/await
 
-After knowing how Promises works and what they are for, let's look at an example in which we handle multiple Promises. Let's say you wanted to react to a message in a certain order with letters (regional indicators). For this example, I will take the basic template for a discord.js bot with some ES6 adjustments.
+### Theory information
```
SpaceEEC: `information` in `Theory information` feels a bit redundant

guide/additional-info/async-await.md
```diff
@@ -35,11 +35,28 @@ deleteMessages(5).then(value => {
...
+	// code
+}
+``` 
+or 
+```js 
+async () => {
+	// code
+}
+```
+also an important thing to know is that a function marked as async will always return a Promise. In addition to this if you return something inside a async function this value will be the resolved value and if you throw an error inside an async function it will reject this Promise with the error object.
```
SpaceEEC: - maybe capitalize `also`?
- I feel like there should be a comma after `to know`, this could be an optional or even wrong one though
- ^ for `In addition to this`, however this is an optional one (Just a personal opinion, you can let that out)
- `inside a_n_ async function`
- `if you return something ... the (returned) promise will resolve with the returned value`. `returned` feels a bit redundant, not sure whether this is actually needed.
- `if you throw an error ... it will reject _the_ Promise with the _thrown_ error.` ~~(You can technically throw anything although it should be an Error)~~
- maybe also a comma after `throw an error inside an async function`?

guide/additional-info/async-await.md
```diff
@@ -35,11 +35,28 @@ deleteMessages(5).then(value => {
...
+In this scenario the `deleteMessages` function returns a Promise. The `.then()` method will trigger if the Promise was resolved and the `.catch()` method if the Promise was rejected. But with our function we resolve the Promise after 2 seconds with the String "Everything is fine", so the `.catch()` method will never be executed.
 
 ## How to implement async/await
 
-After knowing how Promises works and what they are for, let's look at an example in which we handle multiple Promises. Let's say you wanted to react to a message in a certain order with letters (regional indicators). For this example, I will take the basic template for a discord.js bot with some ES6 adjustments.
+### Theory information
+The following Information are important to know before starting working with async/await. You can only use the await keyword inside a function what is marked as async (you call the async keyword before the function name, function keyword or the parameter when you use an arrow function) a simple example would be
```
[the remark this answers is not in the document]
Dev-Yukine: that infront of a function name was supposed to be for methods but youi are right, it sounds ugly

guide/additional-info/async-await.md
Dev-Yukine: because this guide shows best practise i would only covering throwing Errors inside a asnyc function because you should always reject with an actuall error object

guide/additional-info/async-await.md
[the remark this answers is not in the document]
SpaceEEC: Oh, in front of methods! Seems worth to be mentioned.

guide/additional-info/async-await.md
SpaceEEC: Correct, thus strikethrough :P

guide/additional-info/async-await.md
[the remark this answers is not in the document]
Dev-Yukine: hmm will add that again tomorrow 👀

guide/additional-info/async-await.md
```diff
@@ -35,11 +35,28 @@ deleteMessages(5).then(value => {
...
+In this scenario the `deleteMessages` function returns a Promise. The `.then()` method will trigger if the Promise was resolved and the `.catch()` method if the Promise was rejected. But with our function we resolve the Promise after 2 seconds with the String "Everything is fine", so the `.catch()` method will never be executed.
 
 ## How to implement async/await
 
-After knowing how Promises works and what they are for, let's look at an example in which we handle multiple Promises. Let's say you wanted to react to a message in a certain order with letters (regional indicators). For this example, I will take the basic template for a discord.js bot with some ES6 adjustments.
+### Theory
+The following Information are important to know before starting working with async/await. You can only use the await keyword inside a function that is declared as async (you put the async keyword before the  function keyword or the parameter when you use an arrow function) a simple example would be:
```
SpaceEEC: You have two spaces before the word "function" in the parentheses.

guide/additional-info/async-await.md
```diff
@@ -35,11 +35,28 @@ deleteMessages(5).then(value => {
...
+	// code
+}
+``` 
+or 
+```js 
+async () => {
+	// code
+}
+```
+Also an important thing to know, is that a function declared as async will always return a Promise. In addition to this, if you return something inside a async function the returned promise will resolve with that value and if you throw an error inside an async function it will reject the Promise with the error object.
```
SpaceEEC: `inside a_n_ async function` (missing n)

The `inside an async function` when you throw an error feels redundant, it should be clear that the async function is meant.
Same for `reject the Promise with the error object`, this could be `reject with the thrown error.` or something similar. (Especially `object` feels wrong here)

guide/additional-info/async-await.md
```diff
@@ -35,11 +35,28 @@ deleteMessages(5).then(value => {
...
+In this scenario the `deleteMessages` function returns a Promise. The `.then()` method will trigger if the Promise was resolved and the `.catch()` method if the Promise was rejected. But with our function we resolve the Promise after 2 seconds with the String "Everything is fine", so the `.catch()` method will never be executed.
 
 ## How to implement async/await
 
-After knowing how Promises works and what they are for, let's look at an example in which we handle multiple Promises. Let's say you wanted to react to a message in a certain order with letters (regional indicators). For this example, I will take the basic template for a discord.js bot with some ES6 adjustments.
+### Theory
+The following Information are important to know before starting working with async/await. You can only use the await keyword inside a function that is declared as async (you put the async keyword before the function keyword or the parameter when you use an arrow function) a simple example would be:
```
SpaceEEC: `Information` should be in lowercase.

guide/additional-info/async-await.md
```diff
@@ -35,11 +35,28 @@ deleteMessages(5).then(value => {
...
+	// code
+}
+``` 
+or 
+```js 
+async () => {
+	// code
+}
+```
+Also an important thing to know, is that a function declared as async will always return a Promise. In addition to this, if you return something the Promise will resolve with that value and if you throw an error it will reject the Promise with this error.
```
SpaceEEC: `... reject the Promise with _that_ error.`, to be consistent with the resolve part.

guide/additional-info/async-await.md
```diff
@@ -35,11 +35,28 @@ deleteMessages(5).then(value => {
...
+```js 
+async () => {
+	// code
+}
+```
+Also an important thing to know, is that a function declared as async will always return a Promise. In addition to this, if you return something the Promise will resolve with that value and if you throw an error it will reject the Promise with this error. 
+
+
+### Execution with discord.js code
+After knowing how Promises works and what they are for as well as about the theory, let's look at an example in which we'll handle multiple Promises. Let's say you want to react with letters (regional indicators) in a certain order. For this example, you will take the basic template for a discord.js bot with some ES6 adjustments.
```
SpaceEEC: - `how Promises work`, no `s`
- maybe a comma after `what they are for`? (I'm not that confident in my comma placements, this just reads a bit confusing without one)

guide/additional-info/async-await.md
```diff
@@ -35,11 +35,28 @@ deleteMessages(5).then(value => {
 });
 ```
 
-In this scenario the `deleteMessages` function returns a Promise. The `.then()` function will trigger if the Promise was resolved and the `.catch()` function if the Promise was rejected. But with our function we resolve the Promise after 2 seconds with the String "Everything is fine", so the `.catch()` function will never be executed.
+In this scenario the `deleteMessages` function returns a Promise. The `.then()` method will trigger if the Promise was resolved and the `.catch()` method if the Promise was rejected. But with our function we resolve the Promise after 2 seconds with the String "Everything is fine", so the `.catch()` method will never be executed.
```
Danktuary: Phrases that need a comma after them:
* "In this scenario"
* "... if the Promise was resolved"
* "But with our  function ..."

guide/additional-info/async-await.md
```diff
@@ -35,11 +35,28 @@ deleteMessages(5).then(value => {
...
 
 ## How to implement async/await
 
-After knowing how Promises works and what they are for, let's look at an example in which we handle multiple Promises. Let's say you wanted to react to a message in a certain order with letters (regional indicators). For this example, I will take the basic template for a discord.js bot with some ES6 adjustments.
+### Theory
+The following information are important to know before starting working with async/await. You can only use the await keyword inside a function that is declared as async (you put the async keyword before the function keyword or the parameter when you use an arrow function) a simple example would be:
```
Danktuary: * "The following information are important" -> "The following information is important"
* Backticks around "await" in "You can only use the await keyword ..."
* Backticks around "async" in "you put the async keyword before ..."
* "before the function keyword" should be "before the function name".
* "or the parameter when you use an arrow function" isn't entirely correct. I think a more accurate description would be "or before the parameters when using a callback function".
* That sentence needs a period, making "a simple example would be:" its own sentence.

guide/additional-info/async-await.md
```diff
@@ -35,11 +35,28 @@ deleteMessages(5).then(value => {
...
+```js
+async function markedAsAsync() {
+	// code
+}
+``` 
+or 
+```js 
+async () => {
```
Danktuary: This isn't entirely valid. I get the point you're trying to make, but this may confuse some people. Instead, store this in a variable, or use some kind of `client.on('event', async (first, last) => ...)` example.

guide/additional-info/async-await.md
```diff
@@ -35,11 +35,28 @@ deleteMessages(5).then(value => {
...
+	// code
+}
+```
+Also an important thing to know, is that a function declared as async will always return a Promise. In addition to this, if you return something the Promise will resolve with that value and if you throw an error it will reject the Promise with that error.
```
Danktuary: * The comma in the first sentence doesn't need to be there.
* Using "Also" and "In addition to" in concurrent sentences is a bit repetitive. You can remove the "Also" entirely.
* That last sentence could benefit from a couple of commas. The way I think would be ideal to read it would be:
```` 
In addition to this, if you return something, the Promise will resolve with that value, and if you throw an error, it will reject the Promise with that error.
```

guide/additional-info/async-await.md
```diff
@@ -35,25 +35,35 @@ deleteMessages(5).then(value => {
...
+In this scenario, the `deleteMessages` function returns a Promise. The `.then()` method will trigger if the Promise was resolved, and the `.catch()` method if the Promise was rejected. But with our function, we resolve the Promise after 2 seconds with the String "Everything is fine", so the `.catch()` method will never be executed.
 
 ## How to implement async/await
 
 ### Theory
-The following information are important to know before starting working with async/await. You can only use the await keyword inside a function that is declared as async (you put the async keyword before the function keyword or the parameter when you use an arrow function) a simple example would be:
+The following information is important to know before starting working with async/await. You can only use the `await` keyword inside a function that is declared as `async` (you put the async keyword before the function keyword or before the parameters when using a callback function).
```
Danktuary: * "before starting working with async/await" sounds a bit awkward. I think just "before working with async/await" would be alright.
* You forgot to put backticks around "async" inside the phrase in parenthesis.

guide/additional-info/async-await.md
```diff
@@ -35,25 +35,35 @@ deleteMessages(5).then(value => {
...
 ## How to implement async/await
 
 ### Theory
-The following information are important to know before starting working with async/await. You can only use the await keyword inside a function that is declared as async (you put the async keyword before the function keyword or the parameter when you use an arrow function) a simple example would be:
+The following information is important to know before starting working with async/await. You can only use the `await` keyword inside a function that is declared as `async` (you put the async keyword before the function keyword or before the parameters when using a callback function). 
+
+A simple example would be:
 ```js
-async function markedAsAsync() {
+async function declaredAsync() {
```
Danktuary: You name it `declaredAsync` here but then `declaredAsAsync` below.

guide/additional-info/async-await.md
```diff
@@ -35,25 +35,35 @@ deleteMessages(5).then(value => {
...
 or 
 ```js 
-async () => {
+const declaredAsAsync = async () => {
 	// code
 }
 ```
-Also an important thing to know, is that a function declared as async will always return a Promise. In addition to this, if you return something the Promise will resolve with that value and if you throw an error it will reject the Promise with that error. 
 
+you can use that aswell if you use the Arrow function as a event listener function.
```
Danktuary: * Beginning of the sentence needs to be capitalized.
* "aswell" -> "as well".
* "Arrow" shouldn't be capitalized.
* "as a event listener" -> "as an event listener".

PR #40, opened 2017-11-11
Various phrasing, stylistic and other adjustments

robflop:
As the title says -- various awkward phrasing occurrences have been adjusted, stylistics changes such as spacing out objects or capitalization, consistency changes and more.

Edit: Also as discussed on Discord so this doesn't get lost: 

killing / restarting a shard needs clarification ( <http://discordjs.guide/#/sharding/additional-information> )

PR #42, opened 2017-11-16
Update code samples

robflop:
Forgot to update code samples along with #40. This is them.


code_samples/sequelize/currency/app.js
```diff
@@ -102,10 +102,10 @@ client.on('message', async msg => {
 
 		return msg.channel.send('```' +
```
Danktuary: This as a whole should be converted to `message.channel.send(stuffHere, { code: true })` (both here and in the guide page).

code_samples/sequelize/currency/dbInit.js
```diff
@@ -8,10 +8,10 @@ const sequelize = new Sequelize('database', 'username', 'password', {
 });
 
 const CurrencyShop = sequelize.import('models/CurrencyShop');
-sequelize.import('models/Users')
+sequelize.import('models/Users');
 sequelize.import('models/UserItems');
 
-sequelize.sync().then(async () => {
+sequelize.sync().then(async() => {
```
Danktuary: Same as above; the space should be there.

code_samples/sequelize/tags/sequelize.js
```diff
@@ -38,7 +38,7 @@ client.once('ready', () => {
 	Tags.sync();
 });
 
-client.on('message', async (msg) => {
+client.on('message', async(msg) => {
```
Danktuary: Space here and no parenthesis around `msg`. I think it'd also be a good idea to change all instances of `msg` to `message`, since that's what we use everywhere else in the guide.

code_samples/sharding/getting-started/index.js
```diff
@@ -1,4 +1,8 @@
 const { ShardingManager } = require('discord.js');
-const manager = new ShardingManager('./bot.js', { totalShards: 'auto', token: 'your-token-goes-here' });
+const manager = new ShardingManager('./bot.js', {
+	totalShards: 'auto',
```
Danktuary: I found out earlier that `totalShards` has a default value of `auto`, so this is rendered as redundant and can be removed, as well as putting the object on a single line again. The guide page codeblock and paragraph(s) should be updated to reflect this as well.

code_samples/sequelize/currency/app.js
```diff
@@ -26,7 +26,7 @@ Reflect.defineProperty(currency, 'getBalance', {
 	},
 });
 
-client.once('ready', async () => {
+client.once('ready', async() => {
```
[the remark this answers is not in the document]
robflop: The eslint file needs to be updated then, I adjusted it based on the errors it was giving out.

code_samples/sequelize/currency/dbInit.js
robflop: Same as above for this as well, eslint file dictated it.

code_samples/sequelize/tags/sequelize.js
robflop: Once again, eslint file dictates - but changing the variable names, sure.

code_samples/sequelize/currency/app.js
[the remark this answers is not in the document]
Danktuary: Then I guess update the rule here and on the linter page for this. Apparently there's options for anonymous, named, and one specifically for async arrow functions. https://eslint.org/docs/rules/space-before-function-paren

Set anonymous and named to never, and async arrow function to always.

guide/sequelize/currency.md
```diff
@@ -314,16 +314,16 @@ We use `.add()` for both removing and adding currency. Since we already check if
 
 ```js
 const item = await CurrencyShop.findOne({ where: { name: { $like: commandArgs } } });
```
Danktuary: As mentioned in Discord, this should be `$iLike`. It seems like the code sample is already using `$iLike`, so just update it here and in the text below.

code_samples/sequelize/currency/app.js
```diff
@@ -0,0 +1,114 @@
...
+	}
+	else if (command === 'leaderboard') {
+
+		return message.channel.send(
+			Array.from(currency.entries()).sort((a, b) => b[1].balance - a[1].balance)
+				.filter((u) => client.users.get(u[0]))
+				.slice(0, 10)
+				.map((v, k) => `(${k + 1}) ${(client.users.get(v[0]).tag)}: ${v[1].balance}💰`)
+				.join('\n'),
+			{ code: true });
```
Danktuary: * This code is slightly outdated from the actual guide page
* The final parenthesis (the one matching the `.send(`'s opening paren) should be put on the next line, both here and on the guide page.

code_samples/sequelize/currency/app.js
```diff
@@ -0,0 +1,114 @@
...
+
+		message.channel.send(`You've bought a ${item.name}`);
+
+	}
+	else if (command === 'shop') {
+
+		const items = await CurrencyShop.findAll();
+		return message.channel.send('```' +
+			items.map(i => `${i.name}: ${i.cost}💰`).join('\n')
+		+ '```');
```
Danktuary: This should also be using the `code: true` message option (can be put on a single line).

Danktuary:
Also, if you don't mind, could you add this to the to-do list somewhere?

```md
* Killing / restarting a shard [needs clarification](http://discordjs.guide/#/sharding/additional-information)
* Update `Reflect.defineProperty()` and prototyped methods to use `.findOrCreate()`/`.findOrBuild()` instead of manual checks
* Fix code style inconsistency in the Sequelize pages
* Fix wording style inconsistency in the Sequelize pages (e.g. using "we" instead of "you")
```

Probably under a section titled "Fixes" or something.

PR #45, opened 2017-11-26
fixes stuff Gus and Appel pointed out (in async/await)

Dev-Yukine:
because they yelled at me for that 😭

### DIFF
--- a/guide/additional-info/async-await.md
+++ b/guide/additional-info/async-await.md
@@ -1,6 +1,6 @@
 # Understanding async/await
 
-If you aren't very familiar with ECMAScript 2017, you may not know about async/await. It's a useful way to handle Promises in a hoisted manner. Also, it looks cleaner, it increases readability overall and it is slightly faster.
+If you aren't very familiar with ECMAScript 2017, you may not know about async/await. It's a useful way to handle Promises in a hoisted manner. In addition, it looks cleaner, increases overall readability, and is slightly faster.
 
 ## How do Promises work?
 
@@ -35,7 +35,7 @@ deleteMessages(5).then(value => {
 });
 ```
 
-In this scenario, the `deleteMessages` function returns a Promise. The `.then()` method will trigger if the Promise was resolved, and the `.catch()` method if the Promise was rejected. But with our function, we resolve the Promise after 2 seconds with the String "Deleted 10 messages.", so the `.catch()` method will never be executed.
+In this scenario, the `deleteMessages` function returns a Promise. The `.then()` method will trigger if the Promise was resolved, and the `.catch()` method if the Promise was rejected. But with our function, we resolve the Promise after 2 seconds with the String "Deleted 10 messages.", so the `.catch()` method will never be executed. You can also pass the `.catch()` function as the second parameter of `.then()`.
 
 ## How to implement async/await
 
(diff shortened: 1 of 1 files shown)


guide/additional-info/async-await.md
```diff
@@ -109,24 +109,18 @@ But since all of these react methods are started at the same time, it would just
...
 client.on('message', message => {
 	if (message.content === `${prefix}react`) {
-		message.react('🇦').then(() => {
-			message.react('🇧').then(() => {
-				message.react('🇨').catch(error => {
-					// handle failure of the third react
-				});
+		message.react('🇦')            
+			.then(() => message.react('B'))
+            .then(() => message.react('C'))
```
Drahcirius: Tabs instead of spaces please

guide/additional-info/async-await.md
```diff
@@ -109,24 +109,18 @@ But since all of these react methods are started at the same time, it would just
 ```js
 	if (message.content === `${prefix}react`) {
-		message.react('🇦').then(() => {
-			message.react('🇧').then(() => {
-				message.react('🇨').catch(error => {
-					// handle failure of the third react
-				});
+		message.react('🇦')            
+			.then(() => message.react('B'))
```
Drahcirius: These are the same smallcaps letters you use everywhere else.

Dev-Yukine: what do you mean by that? 🤔

Drahcirius: 🇧 vs B

guide/additional-info/async-await.md
```diff
@@ -20,9 +20,9 @@ One important thing to know is that a Promise can only have one state at a time;
 
 ```js
 function deleteMessages(amount) {
-	return new Promise((resolve, reject) => {
-		if (amount > 10) return reject(new Error('You can\'t delete more than 10 Messages at a time.'));
-		resolve('Deleted 10 messages.');
+	return new Promise((resolve) => {
+		if (amount > 10) throw new Error('You can\'t delete more than 10 Messages at a time.');
+		setTimeout(() => resolve('Deleted 10 messages.'), 2000)
```
appellation: this message is still incongruent with what the text below says it resolves with

guide/additional-info/async-await.md
```diff
@@ -109,24 +109,17 @@ But since all of these react methods are started at the same time, it would just
...
-		})
-		.catch(error => {
-			// handle failure of the first react
-		});
 	}
 });
 ```
 
-This code looks really messy and has redundant catch methods, so let's look how the same code would look with async/await.
+This is an special form of resolving mutiple Promises because we use the `.then()` callback to chain resolving Promises with each other and if one of the Promises gets rejected the `.catch()` does fire. So let's look how the same code would look with async/await.
```
appellation: 1. I dunno how "special" promise chains are.
2. Technically the `catch` callback is called, not fired.
3. Might be useful to link to documentation on promise chains (https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/then#Chaining maybe?)

guide/additional-info/async-await.md
```diff
@@ -1,6 +1,6 @@
 # Understanding async/await
 
-If you aren't very familiar with ECMAScript 2017, you may not know about async/await. It's a useful way to handle Promises in a synchronous manner instead stacking `.then()` callbacks. Also, it looks cleaner and it increases readability overall.
+If you aren't very familiar with ECMAScript 2017, you may not know about async/await. It's a useful way to handle Promises in a hoisted manner. Also, it looks cleaner, it increases readability overall and it is slightly faster.
```
Danktuary: The last sounds a bit repetitive, when describing the code with "it"/"it is". I'd suggest changing it to "In addition, it looks cleaner, increases overall readability, and is slightly faster."

guide/additional-info/async-await.md
```diff
@@ -109,24 +109,17 @@ But since all of these react methods are started at the same time, it would just
...
-		})
-		.catch(error => {
-			// handle failure of the first react
-		});
 	}
 });
 ```
 
-This code looks really messy and has redundant catch methods, so let's look how the same code would look with async/await.
+In this piece of code we [chain resolve](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/then#Chaining) Promises with each other and if one of the Promises gets rejected the function we passed to `.catch()` get called. So let's look how the same code would look with async/await.
```
Danktuary: * A comma after "In this piece of code".
* A comma after "Promises with each other" and "gets rejected".
* "So let's look how" -> "So let's look at how".

PR #50, opened 2017-12-12
ES5 example not actually in ES5

Danktuary:
* Fixed an ES5 example not actually being written in ES5 (along with missing a semicolon).
* Fixed a grammar error.
* Added a small example of destructuring with variable renaming.
* Move the template literals section above the arrow functions section, because the arrow function examples use template literals in the examples, so it doesn't make much sense to display that before they're explained.

### DIFF
--- a/guide/additional-info/es6-syntax.md
+++ b/guide/additional-info/es6-syntax.md
@@ -37,6 +37,85 @@ If you haven't noticed, this piece of code is actually already using a bit of ES
 
 As for the code above, there are a few places where things can be done better. Let's look at them.
 
+### Template literals
+
+If you check the code above, it's currently doing things like `prefix + 'name'` and `'Your username: ' + message.author.username`, which is perfectly valid. It is a bit hard to read, though, it's not too fun to constantly type out. Fortunately, there's a better alternative.
+
+```js
+// ES5 version, as we currently have it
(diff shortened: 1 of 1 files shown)


Danktuary:
Could you elaborate on what you mean? I'm a bit confused. And would that fit well on a page that's solely for explaining ES6 features?

Dev-Yukine:
well when using template literals with not string variables (Numbers, Objects etc.) it make usage od the toString() method to convert that to a string and that means if you overwrite an existing prototypes/create your own toString() method on lets say a class then it will display whatever this method return. and you correct maybe that is not quit needed while explaining ES6 but thats a nice thing to note to mention.

Danktuary:
That wouldn't really fit in this page (or currently any page in the guide); it's more so just a JS thing and has been around since ES5 (or possibly before), so I'll leave that out for the time being.

Drahcirius:
Do you think there would be some value to clarify that template strings doesn't replace concatenation, but rather, it's string interpolation, which is common in other languages? It's not specific to new JS developers, but if they're familiar with other languages, that might be the term that they would recognize. Perhaps mention they're similar to heredocs as well, since JS didn't handle strings that spanned multiple lines, but template strings allows that?

Danktuary:
Fair point; I'll squeeze that in somewhere.

PR #51, opened 2017-12-30
Add a page explaining reactions

Danktuary:
This new page covers things like how to react with both unicode and custom emojis, how to ensure order when reacting, awaiting reactions (kindof), and how to listen for reactions on uncached messages via the `raw` event.
I've also renamed the "FAQ" sidebar section title to "Popular Topics", because "Common Questions" right under a "FAQ" section seems redundant, and I couldn't find a better place to put this page.


guide/additional-info/reactions.md
```diff
@@ -0,0 +1,246 @@
...
+![Emojis reacting in correct order](/assets/img/1IWSJ1C.png)
+
+<p class="tip">If you aren't familiar with Promises or `async`/`await`, you can read more about them on [MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise) or [our guide page on async/await]((/additional-info/async-await)!</p>
+
+## Awaiting reactions
+
+A common use case for reactions in commands is having a user confirm or deny an action, or creating a poll system. Luckily, we actually [already have a guide page that covers this](/creating-your-bot/collectors)! Check out that page if you want a more in-depth explanation. Otherwise, here's a basic example for reference:
+
+```js
+message.react('👍').then(() => message.react('👎'));
```
Lewdcario: Perhaps suggest as an optimization trick, to use `Promise.all([promises])`, if they do not care about order (and I guess have more than two promises to put). Like in the thumbsup/down reaction sections, I do not think one would care about the order. Doesn't matter if you include this or not, just a thought.

PR #52, opened 2017-12-31
Add a page for misc examples

Danktuary:
Most of these taken from the old #examples channel on the d.js server, but this can be used as a place to store more random examples that don't need a full page of explaining later on.


guide/additional-info/miscellaneous-examples.md
```diff
@@ -0,0 +1,161 @@
...
+		});
+	}
+});
+
+client.login('your-token-goes-here');
+```
+
+### Catching UnhandledPromiseRejectionWarnings
+
+If you've ever seen something labled `UnhandledPromiseRejectionWarnings` in your console, it can sometimes be a bit vague as to why it occured. In addition to that, not handling promise rejections is currently deprecated and will eventually be removed. Once that happens, any unhandled promise rejections will close your app with a non-zero exit code.
```
Lewdcario: `it can sometimes be a bit vague as to why it occured` -> `the reason of occurrence can be vague` or something like this

guide/additional-info/miscellaneous-examples.md
```diff
@@ -0,0 +1,161 @@
...
+
+message.channel.send({ embed: exampleEmbed });
+```
+
+##### RichEmbed builder notes
+
+* If you're on discord.js v11.2 or above, you can simply do `message.channel.send(exampleEmbed)`.
+* For `.setColor()`, it accepts a base 10 integer, HEX color as a string, or an array of RGB values.
+* There are also specific strings you can use with `.setColor()`, e.g. `DARK_RED` or `RANDOM`. A list of them can be found [here](https://discord.js.org/#/docs/main/stable/typedef/ColorResolvable).
+* `.addBlankField()` is just shorthand for `.addField('\u200b', '\u200b')`. If you want to make it inline as well, pass in `true` as the first argument.
```
Lewdcario: You could possibly note that the formatting might be not ideal. Perhaps suggest the invisible thumbnail workaround. Or not if you want. Just a thought.

guide/creating-your-bot/common-questions.md
```diff
@@ -39,6 +39,14 @@ if (member.roles.exists('name', '<role name>')) {
 }
 ```
 
+### How do I limit a command to a single user?
+
+```js
+if (message.author.id === '<id>') {
```
Lewdcario: I somewhat recommend using the inverse as there's less chance of bracket worries for beginners.

guide/additional-info/miscellaneous-examples.md
Lewdcario: Also, I think it is just `unhandledPromiseRejection` *warnings*, like, you listen for that event. I think that changes things, not sure.

guide/creating-your-bot/common-questions.md
Danktuary: I'm on the fence about this simple because of how the question is asked and how the code should reflect the question, and vice versa, and also because of how simple it looks. Consistency is a good point, but I think that would apply more to the pages with full explanations, rather than these snippets.

guide/additional-info/miscellaneous-examples.md
Danktuary: Something that just now came to mind: since this section contains a bit too much info to be considered a "quick" example, once the embeds page is created, we can strip the info from this page and move it there, and then leave a message saying "Hey, check out this page for more info on embeds" or something along those lines.

PR #53, opened 2018-01-01
The Picture On The Guide Is Outdated I Guess

lloti:


Danktuary:
While you're at it, you could update the image right after that, since there's a new fancy "Generate OAuth2 URL" button.

PR #59, opened 2018-01-14
[W.I.P] Add raw event guide

Darqam:
So this is very much just a rough draft of this page.

I think a few examples with reaction add, member update, and message delete could be nice as well. Just to show the link between event emitting+receiving, and handling possible 'incomplete' events.

Also my english is generally bad, so feel free to correct any terrible english.

### DIFF
--- a/guide/additional-info/raw-event.md
+++ b/guide/additional-info/raw-event.md
@@ -35,3 +35,5 @@ In many cases, this event won't be necessary for bot coders. The options to fetc
 Similarly, one could want to track a guildMember update event, even if the details of the update are not available. 
 
 In some cases, the raw event will contain sufficient information to be able to cleanly emit a regular d.js event, in other cases this might not be the case (taking the example of message reaction add versus guild member update).
+
+### A few examples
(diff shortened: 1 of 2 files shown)


guide/additional-info/raw-event.md
```diff
@@ -0,0 +1,39 @@
...
+
+Since the raw event doesn't depend on the djs cache to trigger, this event will occur on every trigger, whether the object is cached by d.js or not (for example the member, or message cache). 
+
+Using this raw event thus allows you to not have to worry about the d.js cache being incomplete for your code to work.
+
+### The bad
+
+What makes this event great is also what brings it down, the lack of the d.js cache. If we continue with the example of GUILD_MEMBER_UPDATE, using the raw event, we will get ONLY the new guild member object. That is, it is possible to know that _something_ changed, but not what changed. If the guild member object had been in the d.js cache, we would have had access to the oldMember and newMember object.
+
+Now before the question is asked, no it is not possible to cache the oldMember object as the raw event is triggered. Once that event is triggered, the guildMember has already been updated, the old information, if not cached, is gone.
```
Danktuary: This part isn't exactly true. The user could create their own caching system and compare the new data they get with the data they've stored.

PR #66, opened 2018-01-23
Fix responsive theme toggle fade issue

Danktuary:
There's this really weird issue that only occurs when viewing the site on mobile devices (or a browser with the window width at <= 768px). If you open the sidebar and click the theme toggle button, the sidebar and theme toggle buttons will fade colors instead of immediately changing like the rest of the site. Tiny PR but this fixes that.


1Computer1:
revolutionary

Drahcirius:
We should add this to the entire site @Danktuary , the fade-in looks smooth.

PR #70, opened 2018-02-03

harrycode:


Danktuary:
Sorry, but I'm gonna have to decline this. There are many things in this that don't fit with the rest of the guide, such as:

1. It doesn't provide any description of what the snippet does.
2. It assumes this would need to go in a different file, and doesn't follow this guide's command handling syntax.
3. It doesn't comply with the ESLint config file we use.

But the biggest thing would be that it's simply too niche; I don't think many users who read the guide could benefit from this, which is the guide's goal.

Thanks anyway, though!

PR #74, opened 2018-02-06
Change color of .inserted 

RuyiLi:
The old color (in my opinion) was a bit too light, giving it the feel of a comment; a darker color would help improve its clarity.

Old:
image

New:
image

### DIFF
--- a/guide/assets/css/main.css
+++ b/guide/assets/css/main.css
@@ -150,7 +150,7 @@ body.close #theme-toggle-button {
 }
 
 .token.inserted {
-	color: #42b983;
+	color: #00ab58;
 	border-width: 0;
 	transition: color 0.2s ease-in-out;
 }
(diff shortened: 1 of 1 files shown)


Danktuary:
I'm not fond of the new color tbh. It looks too lime-y of a green; doesn't fit well with the rest of the guide. Also doesn't too great on the dark theme ([old](https://i.imgur.com/zG7ftGG.png) vs [new](https://i.imgur.com/NpuKWMG.png)).

If you wanna find a different shade of green to use, that'd be okay, but otherwise I'd just leave it alone.
(As a side note, this would need to be applied to other codeblocks used around the guide and not just diff codeblocks.)

RuyiLi:
Just thought it seemed a little pale, so I went for a more vibrant color. I guess it doesn't really matter in the grand scheme of things, so I'll close this for now.

Also, didn't know there was a dark theme. Neat.

PR #75, opened 2018-02-06
Add bot restart and user tagging examples

Danktuary:
2 simple examples that are asked almost daily in the support channels, as well as some `<p class="tip">` tags to accompany them with relevant, useful info. Also updated the reaction adding example, since you can just directly insert an emoji ID if you have that, instead of using `client.emojis.get('<id>')`.

### DIFF
--- a/guide/popular-topics/common-questions.md
+++ b/guide/popular-topics/common-questions.md
@@ -101,8 +101,9 @@ user.send('<content>');
 ### How do I tag a certain user in a message?
 
 ```js
-const user = <client>.users.get('<id>');
+const user = <message>.mentions.users.first();
 <message>.channel.send(`Hi, ${user}.`);
+<message>.channel.send('Hi, <@user id>.');
 ```
 
 <p class="tip">If you want to tag the user who sent the message, you can use `<message>.reply()`. For example: `<message>.reply('hi.')` would result in `@User, hi.`. If you want to insert the tag elsewhere, you can store `<message>.author` as your `user` variable and use the original example.</p>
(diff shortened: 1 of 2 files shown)


guide/popular-topics/common-questions.md
```diff
@@ -98,6 +98,17 @@ user.send('<content>');
 
 <p class="tip">If you want to DM the user who sent the message, you can use `<message>.author.send()`.</p>
 
+### How do I tag a certain user in a message?
+
+```js
+const user = <client>.users.get('<id>');
+<message>.channel.send(`Hi, ${user}.`);
```
Lewdcario: Suggest to use <@id> as well.

guide/popular-topics/common-questions.md
```diff
@@ -121,12 +132,22 @@ user.send('<content>');
...
+process.exit();
+```
+
+<p class="tip">`process.exit()` will only kill your Node process, but when using [PM2](http://pm2.keymetrics.io/), it will restart the process whenever it gets killed. You can read our guide on PM2 [here](/improving-dev-environment/pm2).</p>
+
+<p class="warning">Be sure to [limit this to your own ID](/popular-topics/common-questions?id=how-do-i-limit-a-command-to-a-single-user) so that other users can't restart your bot!</p>
+
 ### What is the difference between a User and a GuildMember?
 
 A lot of users get confused as to what the difference between Users and GuildMembers is. The simple answer is that a User represent a global Discord user and a GuildMember represents a Discord user on a specific server. That means only GuildMembers can have permissions, roles, and nicknames, for example, because all these things are server-bound information that could be different on every server that user is in.
```
Lewdcario: `because all these things` -> `because all of these things`
`every` -> `any`

guide/popular-topics/common-questions.md
Danktuary: I'll change the first example to a `message.mentions.users.first()` snippet then, so that it's not as repetitive.

PR #76, opened 2018-02-08
Add linting support for JS files and JS codeblocks in .md files

Danktuary:
Found this [pretty cool eslint plugin](https://github.com/eslint/eslint-plugin-markdown) that lints JS codeblocks inside .md files, which is perfect for the guide. Since it requires ESLint, I've added a `package.json` file, appropriately moved around the `.eslintrc.json` files, added some `package.json` scripts, and satisfied the linter as best as possible.

There's still the debate of choosing to remove the `<message>`-like stuff on the FAQ page, so that we can get those codeblocks linted as well instead of skipping over them. Personally, I'd prefer them without the `<>` either way, but it's not a huge deal; I'd be okay with keeping it as is.

After this, we can set up Travis and make sure all incoming PRs get linted as well. 😄


guide/popular-topics/reactions.md
```diff
@@ -140,7 +140,7 @@ if (message.content === '!fruits') {
 	Promise.all([
 		message.react('🍎'),
 		message.react('🍊'),
-		message.react('🍇')
+		message.react('🍇'),
```
iCrawl: This is definitely not okay.

Gawdl3y: Agreed. This is utterly appalling, and quite frankly, disgusting.

Danktuary: Begone, Thots

PR #77, opened 2018-02-09
Add array destructuring section to es6 page

Danktuary:
Since destructuring isn't limited to only objects, I've updated the destructuring section on the ES6 Syntax page to include an explanation/examples for array destructuing, as well as updated the object destructuring section a bit.

### DIFF
--- a/guide/additional-info/es6-syntax.md
+++ b/guide/additional-info/es6-syntax.md
@@ -174,23 +174,29 @@ We won't be covering the lexical `this` scope with arrow functions in here, but 
 
 ### Destructuring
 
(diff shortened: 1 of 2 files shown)


guide/additional-info/es6-syntax.md
```diff
@@ -174,23 +174,29 @@ We won't be covering the lexical `this` scope with arrow functions in here, but
 
 ### Destructuring
 
-Destructuring is an easy way to extract properties from an object. If you've never seen this syntax before, it can be a bit confusing. But it's actually very easy to understand! Here, take a look:
+Destructuring is an easy way to extract items from an object or array. If you've never seen the syntax for it before, it can be a bit confusing, but it's actually very easy to understand once explained!
+
+#### Object destructuring
+
+Here's an common example where destructuring would come in handy:
```
lloti: an common

PR #87, opened 2018-03-19
Update mention prefix example text

Darqam:
Cause people be lazy


lloti:
Was removing the top sentence intentional?

Danktuary:
@lloti It was discussed with me beforehand and I told him it's fine. It's not essential to the example, and I may also move this elsewhere in the future.

PR #88, opened 2018-03-20
update REST guide

devdutchy:
random.cat is now aws.random.cat


Danktuary:
Missing a link reference on line 71 of the rest-api.md page (can't highlight it due to being on mobile atm). Thanks for this!

PR #89, opened 2018-03-24
Update raw-event.js

Wodki:
Added an additional check in case the channel is undefined

### DIFF
--- a/code_samples/popular-topics/reactions/raw-event.js
+++ b/code_samples/popular-topics/reactions/raw-event.js
@@ -11,7 +11,7 @@ client.on('raw', async event => {
 	const { d: data } = event;
 	const channel = client.channels.get(data.channel_id);
 
-	if (channel.messages.has(data.message_id)) return;
+	if (!channel || channel.messages.has(data.message_id)) return;
 
 	const user = client.users.get(data.user_id);
 	const message = await channel.fetchMessage(data.message_id);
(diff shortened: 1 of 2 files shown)


Wodki:
Okay, did. Sorry it took me a while, I hadn't check github.

Danktuary:
Hey @Wodki, thanks for this! I went to the gist and noticed @Boxroom left a comment. I never noticed before because you don't get notifications for gist comments, but I noticed they ran into the same issue as you and came up with a different solution, i.e.:

```js
if (typeof client.channels.get(data.channel_id) === 'undefined') {
	await client.users.get(data.user_id).createDM();
}
```

I think that, instead of returning if no channel was found, creating the DM channel would be a cleaner solution because it would allow reaction fetching via the raw event to work in DMs as well. Could you update your PR to use this instead? 😄 Instead of the code above, I would recommend this instead:

```diff
- const channel = client.channels.get(data.channel_id);
+ const channel = client.channels.get(data.channel_id) || await client.users.get(data.user_id).createDM();
```

And of course, update the reactions.md file to this as well. I noticed that you made a separate PR for that, but you can just make a commit to your [patch-1](https://github.com/Wodki/guide/tree/patch-1) branch on your fork so that it gets added to this PR, since it's all related. Thanks again!

Wodki:
Okay, I think that should do it. Thanks for helping me out to do this properly, I had actually never used the GitHub interface before to make pull requests etc. If you need anything else please tell me :)

Danktuary:
No problem at all, and thank you again! 😄

PR #93, opened 2018-04-25
fixing a typo and explaining possible errors in the music bot example

gloss-water:
I saw some people talking about having trouble with the music bot example, and tried it out for myself on a fresh install.  I ran into a few issues when following the exact words of the tutorial, and tried to clear up the 2 errors that will probably pop up if someone's just copy-pasting the code and hasn't worked with discord.js before.  

The other thing I've seen firsthand from my own experience, as well as in conversations I lurked through in the discord server, is that this does not work in the Ubuntu subsystem for Windows 10. I'd like to put a clearer reason / explanation for why this doesn't work, but I don't have one.  I just know that playing audio this way even with a totally clean environment in the Ubuntu subsystem and following all the steps perfectly does not work.

Oh, and `ytdl-core` was called `ytdl-code`.

### DIFF
--- a/guide/popular-topics/miscellaneous-examples.md
+++ b/guide/popular-topics/miscellaneous-examples.md
@@ -21,13 +21,15 @@ npm install --save node-opus
 If you get an error that says 'FFMPEG not found', this can be resolved by installing ffmpeg.
 
 On Debian / Ubuntu:
+
 ```
 sudo apt-get install ffmpeg
 ```
 
 On Windows:
+
 ```
-npm install ffmpeg-binaries
+npm install ffmpeg-binaries --save
 ```
 
 Additionally, there have been reports that playing audio in this way from the Ubuntu subsystem offered by Windows 10 does not work.
--- a/guide/popular-topics/miscellaneous-examples.md
+++ b/guide/popular-topics/miscellaneous-examples.md
@@ -18,8 +18,14 @@ If you get an error that says 'OPUS_ENGINE_MISSING', you'll need to install one 
 npm install --save node-opus
 ```
 
-If you get an error that says 'FFMPEG not found', this can be resolved by installing the `ffmpeg-binaries` package.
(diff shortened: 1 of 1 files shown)


iCrawl:
This has most likely something to do with the underlying architecture as to how the subsystem interacts with windows and/or networking interface that windows provides.

The subsystem is a rather finicky situation that works well for development, but sometimes not. This also refers to how native Docker does not work in the subsystem for networking issues.

So my educated guess here would be pointing to the shared network the subsystem has with windows and the subsystem can not properly interact with it for anything more than basic interactions.

Drahcirius:
Can you add a note for people on debian/ubuntu that they can run `sudo apt-get install ffmpeg` instead of installing the ffmpeg-binaries package? I think for windows people they can also use chocolatey to install ffmpeg, but I don't have experience with that.

gloss-water:
Oh, that's a good idea, will do!

guide/popular-topics/miscellaneous-examples.md
```diff
@@ -4,12 +4,34 @@
...
+If you get an error that says 'OPUS_ENGINE_MISSING', you'll need to install one of the opus packages discord.js recommends.
+
+```
+npm install --save node-opus
+```
+
+If you get an error that says 'FFMPEG not found', this can be resolved by installing ffmpeg.
+
+On Debian / Ubuntu:
+```
```
Danktuary: Codeblocks and the sentences before them should have a blank line between

guide/popular-topics/miscellaneous-examples.md
```diff
@@ -4,12 +4,34 @@
...
+sudo apt-get install ffmpeg
+```
+
+On Windows:
+```
+npm install ffmpeg-binaries
```
Danktuary: Since Node 6 comes with npm 3.x by default, and for consistency, add the `--save` flag here like you did on the others

gloss-water:
Should be done 😄

PR #104, opened 2018-06-19
Reactions: Listenting on old Messages: account for MESSAGE_REACTION_REMOVE with empty usercache

almostSouji:
The current version of the guide simply returns if the message is cached.
If the message is cached but the usercache on the specified MessageReaction is not populated, the lib will not emmit the event, meaning we are left with a not emitted `MESSAGE_REACTION_REMOVE`.

This is is the case if the message gets cached by the guides raw event script.
Meaning on an uncached message the first reaction will properly emit,  but any following removal of uncached Reactions will not.
Accomplishing this required a bit of shifting in the guide

This PR attempts to fix this issue by letting the guides raw event script handle those cases.
Caveats: The only case breaking this approach is MESSAGE_REACTION_REMOVE on an uncached message inside DM

Which can not be handled due to guide/code-samples/popular-topics/reactions/raw-event.js 
requiring a guild in its constructor.

- [ ] Review for cleaner syntax

### DIFF
--- a/code-samples/popular-topics/reactions/raw-event.js
+++ b/code-samples/popular-topics/reactions/raw-event.js
@@ -16,11 +16,16 @@ client.on('raw', async event => {
 	const { d: data } = event;
 	const user = client.users.get(data.user_id);
 	const channel = client.channels.get(data.channel_id) || await user.createDM();
+	let message = channel.messages.get(data.message_id);
 
-	if (channel.messages.has(data.message_id)) return;
-
-	const message = await channel.fetchMessage(data.message_id);
 	const emojiKey = (data.emoji.id) ? `${data.emoji.name}:${data.emoji.id}` : data.emoji.name;
+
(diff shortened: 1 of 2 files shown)


queezly:
Reverted whatever nonsense that was and propose a new and much cleaner solution.

This as well as my original approach have one failcase:
MESSAGE_REACTION_REMOVE on an uncached message inside DM

This is due to the Emoji constructor https://github.com/discordjs/guide/blob/3ed769a9e71fdfee74e981e91305156f5c0a14a8/code-samples/popular-topics/reactions/raw-event.js#L25 requiring a guild

If you have a clean solution to also handle that case please let me know!

Danktuary:
After a lot of ~~procrastinating~~ discussion, I think it's best if we stop trying to cover all edge cases for this, since it is an unsupported/undocumented feature and a big enough hack as is. If people want to make it perfect, that's okay, but having us try to make it perfect will turn into more of a mess than it already is. 😅 Sorry again for the delay, and thanks for the time and effort you put into this!

PR #105, opened 2018-06-21
Reactions: More information.

gt-c:
Adds the loop method to the reaction race condition section.

### DIFF
--- a/guide/popular-topics/reactions.md
+++ b/guide/popular-topics/reactions.md
@@ -130,10 +130,10 @@ client.on('message', async message => {
 Lastly would be to use a loop.
 
 ```js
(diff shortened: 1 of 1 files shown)


queezly:
I don't really see this as its own standalone third solution, as it's more of an expansion on the async/await method shortening the code for multiple reactions

gt-c:
Good point. That commit should do it. :P

Danktuary:
Sorry for such a late reply!

I agree with Souji though; we already have 3 examples (and I'll probably be removing the `Promise.all()` example since it's not really relevant), and this one doesn't add anything too complex that the user reading shouldn't be able to figure out themselves. Sorry. 😅

PR #106, opened 2018-07-02
Guide: Role assignment on reaction

queezly:
A common asked question has been "How do i add a role to someone that reacts to the bots message"
This PR adds a guide and code sample on the topic categorized under Popular Topics > Reactions

### DIFF
--- a/guide/popular-topics/reactions.md
+++ b/guide/popular-topics/reactions.md
@@ -176,6 +176,33 @@ message.awaitReactions(filter, { max: 1, time: 60000, errors: ['time'] })
 	});
 ```
 
+## Assigning a Role on Reaction
+
+Another common use case for reactions is assigning roles to members that react to the bots message. To achieve our goal we will be utilizing a ReactionCollector. Since we  [already have a guide page that covers this](/popular-topics/collectors#reaction-collectors) we'll be focusing on the aspect of getting the GuildMember from the Collector.
+We will make use of `async`/`await` keywords for the following snippet to prevent callback madness, if you need further elaboration on this topic we [explain this in another part of the guide](/additional-info/async-await).
+
+<!-- eslint-skip -->
+```js
+// we send the initial message and react with the bot so members can easily add reactions
+const reactmessage = await message.channel.send('React with 👌 to get your role!'); 
+await reactmessage.react('👌'); 
+
+const filter = (reaction, user) => reaction.emoji.name === '👌' && !user.bot; 
+const collector = reactmessage.createReactionCollector(filter, { time: 15000 });
+
+collector.on('collect', async reaction => { 
+	const user = reaction.users.last(); // the last user that reacted aka the user that triggered the 'collect' event
+	const guild = reaction.message.guild; // the guild the reaction happened in
+	const member = guild.member(user) || await guild.fetchMember(user); // getting the member from the user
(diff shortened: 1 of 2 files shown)


Danktuary:
@Drahcirius brought up a good point about 2 weeks ago.
>I'm not sure about using a collector for role react though
>if they're going to use a command, why wouldn't the command just give them the role directly

>I can already see people making time 99999999999
>and having tons of collectors
>and every time their bot restarts it dies

It didn't hit me until that 2nd part, but some people might misunderstand how this works exactly. Coincidentally, someone in the support channels did that exact thing a few days after Drah mentioned this. Most people are looking for a more permanent solution, which would pretty much require using the regular `messageReactionAdd` event and `raw` event as well.

As for the PR, I'm not sure if you wanna rework the example or just leave it alone overall. I'd suggest a different example but I honestly can't think of one. Alternatively, if you're up for it and want to create a full page for that kind of stuff, that'd be welcome. If you're not, don't worry, I wouldn't blame you.

guide/popular-topics/reactions.md
```diff
@@ -176,6 +176,33 @@ message.awaitReactions(filter, { max: 1, time: 60000, errors: ['time'] })
...
+await reactmessage.react('👌'); 
+
+const filter = (reaction, user) => reaction.emoji.name === '👌' && !user.bot; 
+const collector = reactmessage.createReactionCollector(filter, { time: 15000 });
+
+collector.on('collect', async reaction => { 
+	const user = reaction.users.last(); // the last user that reacted aka the user that triggered the 'collect' event
+	const guild = reaction.message.guild; // the guild the reaction happened in
+	const member = guild.member(user) || await guild.fetchMember(user); // getting the member from the user
+	member.addRole("some-role-id"); // adding the role to the member
```
xDimGG: Quote consistency

guide/popular-topics/reactions.md
```diff
@@ -176,6 +176,33 @@ message.awaitReactions(filter, { max: 1, time: 60000, errors: ['time'] })
...
+	console.log(`Added the role to ${member.displayName}`);
+});
+```
+<p class="tip">On the masterbranch the `collect` event emits with the User that reacted as second parameter, so you don't need to obtain it as `reaction.users.last()`</p>
```
xDimGG: "masterbranch"

xDimGG: No reason to capitalize "user" here

guide/popular-topics/reactions.md
```diff
@@ -176,6 +176,33 @@ message.awaitReactions(filter, { max: 1, time: 60000, errors: ['time'] })
...
+
+If you want a static massage on the server people can react to to assign their roles you are better off using the `messageReactionAdd` event and comparing the `reaction.message.id` to your specified reactionmessage. This clientevent passes you the [MessageReaction object](https://discord.js.org/#/docs/main/stable/class/MessageReaction) and the user that reacted which you can utilize in the same way as shown above.
```
xDimGG: "massage"

xDimGG: "reactionmessage" "clientevent"

queezly:
Will close this, as the proposed idea needs a lot more rework and planning

Issue #97, opened 2018-05-09
Guide on Permissions


queezly:
Common question: how does discords permission system work
And yeah, it's hard to wrap your head around at first. I played around with it for quite some time since discords beta through various changes and the question comes up a lot in the supportchannels as well.
Long story short: i'm going to write a guide on it.

Planned coverage:
- What are permissions 
- What are overwrites
- Explicit Allow/Deny
- How all of that works with the different Tiers 
- How do Categories play into all of that

Questions I have for you:
1) Where do you think this belongs? Common Questions? Popular topics?
2) Anything else you'd like to see covered?
3) What should an example for this look like in your opinion?
4) Should i include role hierarchy/Admin/Owner overwrites or do you think that deserves its own paragraph?

Lewdcario:
1. popular topics
2. idunno, can just cover all of permissions... how they work with guild members... priority order... channels. can also make sure to cover implicit permissions that the api covers (like if you have no sview channel perms you can't send messages and anything else, similar with send messages).. also can cover how you cant deny role permissions, inheritance, etc. maybe also an example on how to combine permissions to a structure that can be resolved and get inserted into one of our methods. (like in the case where i want to combine x guild members and give them permissions to see a channel, but not the rest of the server)
3. just do what you'd do in code and we can review it. :)
4. all together in the main guide, but it can be on different pages as you see fit.

appellation:
Discord already has some help articles on this topic, so you can save yourself some effort by not overlapping.

https://support.discordapp.com/hc/en-us/articles/206029707
https://support.discordapp.com/hc/en-us/articles/206141927

queezly:
Change requests/amendments at #110

PR #110, opened 2018-07-18
Guide: Permissions and Overwrites

almostSouji:
Fulfillment of #97 
Adds guide and sample code for Discord.js Permissions and Overwrites
Feel free to point out any and all mistakes and request amendments/ make changes as you see fit

Another thing i forgot to mention when posting the PR:

> Discord already has some help articles on this topic, so you can save yourself some effort by not overlapping.

I tried this approach, however found it to be way more approachable to include this information in the guide as relevant so you don't have to interrupt the reading flow to check external sites, which can distract from the topic at hand and hinder understanding

general to do:
- [x] Codesample: fix [object Object] issue
- [x] Codesample: change `!modeveryone` to a more applicable solution
- [x] Codesample: add permission checks on bot
- [x] Codesample: change console.log to message.reply for immediate feedback
- [x] Codesample: check if role with name "Mod" already exists before creating
- [x] Guide: add "How to check for roles on the member" as simple bot "permission" system
- [x] Guide: add "How to remove a certain overwrite"
- [x] Review wording/grammar/spelling

make ready for 11.4:

- [x] Adapt code to be ready to merge for 11.4
- [x] Adapt overwrite sync
- [x] Adapt final permissions for roles
- [x] Review use of documentation links as 11.3-dev will no longer be accesisble on release


BigBrainAFK:
createmod should actually check if a role with the name already exists imo just to showcase how it should be done

code-samples/popular-topics/permissions/index.js
```diff
@@ -0,0 +1,82 @@
...
+				permissions &= ~everyoneOverwrites.deny;
+				permissions |= everyoneOverwrites.allow;
+			}
+			if (roleOverwrites) {
+				permissions &= ~roleOverwrites.deny;
+				permissions |= roleOverwrites.allow;
+			}
+			return new Discord.Permissions(permissions);
+		};
+		message.channel.send(getRoleFinalPermissions(message.channel, message.member.highestRole).serialize(), { code: 'js' });
```
SpaceEEC: Same as above here.

code-samples/popular-topics/permissions/index.js
```diff
@@ -0,0 +1,82 @@
...
+		},
+		{
+			id: message.client.user.id,
+			allow: ['VIEW_CHANNEL'],
+		}]);
+		console.log('Created a private channel');
+	}
+	else if (message.content === '!mypermissions') {
+		const finalPermissions = message.channel.permissionsFor(message.member);
+		message.channel.send(finalPermissions.serialize(), { code: 'js' });
```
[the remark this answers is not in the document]
queezly: I fixed the [objex Object] on the testbot but looks like i forgot to commit the changes on the guide, pardon

queezly:
Another thing i forgot to mention when posting the PR:

> Discord already has some help articles on this topic, so you can save yourself some effort by not overlapping.

I tried this approach, however found it to be way more approachable to include this information in the guide as relevant so you don't have to interrupt the reading flow to check external sites, which can distract from the topic at hand and hinder understanding

code-samples/popular-topics/permissions/index.js
```diff
@@ -0,0 +1,125 @@
...
+		return message.reply(`I need the permissions ${botPerms.join(', ')} for this demonstration to work properly`);
+	}
+
+	if (message.content === '!modeveryone') {
+		const everyonePerms = new Permissions(message.guild.defaultRole.permissions);
+		const newPerms = everyonePerms.add(['MANAGE_MESSAGES', 'KICK_MEMBERS']);
+		message.guild.defaultRole.setPermissions(newPerms.bitfield)
+			.then(() => message.reply('Added mod permissions to `@everyone`'))
+			.catch(console.error)
+		;
```
SpaceEEC: The semicolon here looks misplaced 👀

code-samples/popular-topics/permissions/index.js
```diff
@@ -0,0 +1,125 @@
...
+	}
+
+	else if (message.content === '!unmodeveryone') {
+		const everyonePerms = new Permissions(message.guild.defaultRole.permissions);
+		const newPerms = everyonePerms.remove(['MANAGE_MESSAGES', 'KICK_MEMBERS']);
+		message.guild.defaultRole.setPermissions(newPerms.bitfield)
+			.then(() => message.reply('Removed mod permissions from `@everyone`'))
+			.catch(console.error)
+		;
```
SpaceEEC: Same as above

code-samples/popular-topics/permissions/index.js
queezly: those little rascals must've escaped when I added it to the testbot and ported it over due to autofix

guide/popular-topics/permissions.md
```diff
@@ -0,0 +1,313 @@
...
+if (message.guild.me.permissionsIn(message.channel).has(['SEND_MESSAGES', 'VIEW_CHANNEL'])) {
+	return console.log('I can send a message here');
+}
+
+// Final permissions for a role
+if (channel.permissionsFor(myRole).has(['SEND_MESSAGES', 'VIEW_CHANNEL'])) {
+	channel.send('This role can send a message here');
+}
+```
+As you will generally want to allow `ADMINISTRATOR` to override you don't need to specify a second parameter here. It defaults to true.
```
bdistin: This line is the opposite of what it needs to be. permissionsFor will return Permissions.all if the member has an Administrator Role, however, you should never checkAdmin from the return of a permissionsFor, because the administrator permission does not apply from channel overwrites.

The above examples should also reflect has(permissions, false) / missing(permissions, false) ect.

queezly: After more testing and major brain damage from overthinking: That line should not exist at all neither this nor the other way.
Since permissionsFor() returns Permissions.all on global ADMINISTRATOR there is no way to check explicit permissions on the object returned by that method.

Instead added a warning to clarify that.
Further shuffled and restructured the guide a little in the next commit to get a better logical reading flow

Danktuary:
As mentioned on Discord, I think this should be split up into two different pages. The first half of the page explains a lot of technical in-depth stuff that people who first open the page may not be interested in. A lot of people are just interested in code samples and how to use permissions in d.js.
I don't think it's useless info, though. Some people might find it quite useful, so it'd be good to keep it on a separate page. The structure I'd suggest would be:

- Current page:
    - Intro
    - Hey, here's a link if you wanna know how all this stuff works exactly
    - Bottom half of the page with the actual code applications
- In-depth explanations page
    - Everything from "Discord's permission system" all the way up to "Missing permissions".

But overall looks great. 👍 Gonna leave a few review comments soon.

code-samples/popular-topics/permissions/index.js
```diff
@@ -0,0 +1,130 @@
+const util = require('util');
+const { Client, Permissions } = require('discord.js');
+const client = new Client();
+
+client.on('ready', () => {
+	console.log('Ready!');
+});
+
+client.on('message', message => {
```
Danktuary: The code style for this file is fairly inconsistent with the rest of the files in the guide. Check out [this file](https://github.com/discordjs/guide/blob/master/code-samples/creating-your-bot/commands-with-user-input/index.js) to see a decent example of what I mean and try to apply it to this file.

guide/popular-topics/permissions.md
```diff
@@ -0,0 +1,335 @@
...
+- It is trying to execute a forbidden action on the server owner.
+- It is trying to execute an action based on another unfulfilled factor (for example reserved for partnered guilds). 
+
+<tip>The `ADMINISTRATOR` permission being granted does not skip any hierarchical check!</tip>
+
+## Base permissions
+
+### Setting role permissions
+
+Permissions are set on roles, not the guild member itself. To change them you access the Role object (for example via `guild.roles`) and use the `.setPermissions()` method.
```
Danktuary: The "via `guild.roles`" part isn't exactly accurate, because you say "you access the Role object" when `guild.roles` returns a Collection. I think something like `guild.roles.random()` or `member.roles.first()` would be better .

guide/popular-topics/permissions.md
```diff
@@ -0,0 +1,335 @@
...
+Let's change the base permissions for @everyone so you can see it in action:
+
+```js
+// permission flag array
+guild.defaultRole.setPermissions(['SEND_MESSAGES', 'VIEW_CHANNEL']);
+
+// permission bit field in decimal representation
+guild.defaultRole.setPermissions(3072);
```
Danktuary: Unsure if including the bit field part is necessary, but I suppose some may find it useful. I think that, at the very least, a link to one of the [permissions calculator sites](https://discordapi.com/permissions.html) would be ideal. Probably in a `<tip>` somewhere above or below this example.

guide/popular-topics/permissions.md
```diff
@@ -0,0 +1,335 @@
...
+
+As you have likely already seen in your desktop client, channel overwrites have three states: 
+
+- explicit allow `true` (green ✓)
+- explicit deny `false` (red X) 
+- default `null` (gray /)
+
+### Adding overwrites
+
+To add a permission overwrite for a role or user you access the channel object and use the `.overwritePermissions()` method. The first parameter is the target of the overwrite, either a Role or User object or their respective resolvable, the second is a PermissionOverwriteOptions object.
```
Danktuary: Guild members have permissions, not users. d.js accepts a UserResolvable in this case, but it's quite the same thing.

guide/popular-topics/permissions.md
```diff
@@ -0,0 +1,335 @@
...
+
+Some methods and properties in Discord.js return permission decimals rather than a Permissions object, making it hard to manipulate or read them if you don't want to use bitwise operations.
+You can however pass these decimals to the Permissions constructor to convert them.
+
+```js
+const { Permissions } = require('discord.js');
+
+// constructing a Permissions Object from a permission number
+const permissions = new Permissions(268550160);
+// constructing a Permissions Object from an array of PermissionResolvables
```
Danktuary: Space between this comment and the line of code above it.

guide/popular-topics/permissions.md
```diff
@@ -0,0 +1,335 @@
...
+The `.has()` method takes two parameters. The first being either a permission number, single flag or an array of permission numbers and flags, the second being a boolean (true or false) indicating if you want to allow the `ADMINISTRATOR` permission to override (defaults to true).
+
+Let's say you want to know if the decimal bit field representation `268550160` has `MANAGE_CHANNELS` referenced:
+
+```js
+const { Permissions } = require('discord.js');
+
+const permissions = new Permissions(268550160);
+console.log(permissions.has('MANAGE_CHANNELS'));
+// output: true
```
Danktuary: These comments can go inline, e.g.:
```js
const permissions = new Permissions(268550160);
console.log(permissions.has('MANAGE_CHANNELS')); // true
```

Though, I think we have an eslint rule against this; not entirely sure. If so, just space them out instead.

guide/popular-topics/permissions.md
```diff
@@ -0,0 +1,335 @@
...
+// output : false
+```
+
+You can utilize these methods to adapt permissions or overwrites without touching the other flags. To achieve this you can get the existing permissions for a role, manipulating the bit field as described above and passing the changed bit field to `role.setPermissions()`.
+
+<tip>In the stable branch `role.permissions` returns a number which needs to be converted to a Permissions object for this to work as described here, we covered how to achieve this in the section "Converting permission numbers to Objects"</tip>
+
+## Resulting code
+
+If you want to check a working implementation for all methods and functionalities explained in this guide, you can look at a boilerplate bot over on the GitHub repository [here](https://github.com/discordjs/guide/tree/master/code-samples/popular-topics/permissions/).
```
Danktuary: This piece of text should be the same as all the other pages (because I'm probably gonna turn it all into a Vue component later, so it'll be changed at one point or another regardless.)

guide/popular-topics/permissions.md
almostSouji: removing bitfields from examples and adding a <tip> in terminology instead with "you can use permission decimals where we use flag literals" with the link to the dev portal.

The official permission calculator moved to the "Bot" page in the dev portal and I think we should stick with the official one.

guide/popular-topics/permissions.md
almostSouji: May i suggest generalizing the component a little, then?
```js
If you want to compare your code to the code we've constructed so far, you can review it over on the GitHub repository [here](https://github.com/discordjs/guide/tree/master/code-samples/...).
```
does not really apply in all cases, as this is not a guide with a set goal where you can just compare what you have with what's provided (it is the case with the other pages though) but more of a proof of concept and working example.

code-samples/popular-topics/permissions/index.js
almostSouji: Tried to apply, if you want it to look differently let me know

Danktuary: The blank lines between the if/else if statements should be removed

PR #113, opened 2018-07-20
[11.4] Move embeds to own popular topics

queezly:
**Edit:**
Changed the aim of this PR as discussed with Sanctuary:

Move Embeds in their own category in "popular questions", add missing examples for Attachments and make it ready to merge for 11.4

- [ ] Review wording/grammar/spelling
- [ ] Move content to own pop. topic
- [ ] Rework current file change to work with 11.4
- [ ] Attachments

### DIFF
--- a/guide/popular-topics/miscellaneous-examples.md
+++ b/guide/popular-topics/miscellaneous-examples.md
@@ -196,11 +196,11 @@ To solve this issue you can construct a new RichEmbed and apply the correlating 
 The following code sample results in a bot re-posting any and all embeds it reads:
 
 ```js
-const Discord = require('discord.js');
-const client = new Discord.Client();
+const { Client, RichEmbed } = require('discord.js');
+const client = new Client();
 
-const toRichEmbed = (messageEmbed) => {
-	const richEmbed = new Discord.RichEmbed();
+const toRichEmbed = messageEmbed => {
+	const richEmbed = new RichEmbed();
 	if (messageEmbed.title) {
 		richEmbed.setTitle(messageEmbed.title);
 	}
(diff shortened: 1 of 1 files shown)


lloti:
I think I just died a little on the inside

guide/popular-topics/miscellaneous-examples.md
```diff
@@ -185,6 +185,73 @@ Here's what the embed from code above would look like:
...
+If you ever tried to fetch the embed from an existing message and resend it (maybe in another channel) you will know that this is not easily possible.
+Received embeds are of the class [MessageEmbed](https://discord.js.org/#/docs/main/stable/class/MessageEmbed) while outgoing embeds have to be of the class [RichEmbed](https://discord.js.org/#/docs/main/stable/class/RichEmbed). 
+
+The RichEmbed class does have a constructor taking in embed data, this will however cause conflicts with the circular nature of MessageEmbeds. 
+To solve this issue you can construct a new RichEmbed and apply the correlating values of the received MessageEmbed as demonstrated below. 
+
+The following code sample results in a bot re-posting any and all embeds it reads:
+
+```js
+const Discord = require('discord.js');
```
Tylertron1998: const { RichEmbed } = require('discord.js');

guide/popular-topics/miscellaneous-examples.md
```diff
@@ -185,6 +185,73 @@ Here's what the embed from code above would look like:
...
+const client = new Discord.Client();
+
+const toRichEmbed = (messageEmbed) => {
```
Tylertron1998: Generally for lengthy functions you use a "fat function" (function toRichEmbed() { }) arrow function are used for much smaller functions or Higher-Order-Functions.

guide/popular-topics/miscellaneous-examples.md
iCrawl: `const { Client, RichEmbed } = require('discord.js');`

guide/popular-topics/miscellaneous-examples.md
iCrawl: While I agree that, personally, I would have used a function keyword here too it is pretty much a false statement to say that it's only used for "smaller functions" or higher-order functions.

Danktuary:
I think it's best for us to wait for 11.4 to get released so that we can avoid that custom function and just do something like:

```js
const fetchedMessage = await message.channel.fetchMessage(someID);
const [oldEmbed] = fetchedMessage.embeds;

message.channel.send(new Discord.RichEmbed(oldEmbed));
```

This is supposed to work in 11.3 but it has an issue with circular references. It should be fixed in 11.4 (already partially fixed in the 11.3-dev branch, just missing a fix for fields and embed author circular refs), so we can avoid this entire thing and stick with the simple solution above.

queezly:
Will open a new PR, since its purpose has changed drastically

PR #119, opened 2018-08-06
Add a page on notation

1Computer1:
People always get confused on these things so I thought a page explaining them would be nice. The types section is a big simplification but it should be enough for JS (probably already too much info).


Danktuary:
This is awesome, thanks so much! :smile: I have no complaints about it; just one small suggestion. I think it'd be useful if links to docs were used throughout some of the examples, or maybe even screenshots instead, e.g. under the `Class#foo` paragraph and under the angle brackets paragraph. This could help people remember all of this a bit better, I think.

Issue #121, opened 2018-08-07
Multiple versions


1Computer1:
Support for multiple versions would be nice. You could have one version for stable and one for master for example. The user could then choose with a dropdown menu.

Danktuary:
I'm a bit stuck on how to go about something like this. I had something in mind for this a while ago but never went through with it because we didn't really need it.

My idea was to create a Vue component that'd look something along the lines of this:
```vue
<!-- unsure of the component name, but just as an example -->
<branch-specific branch="stable">
	Here's some code that would only work on the *stable* branch.
</branch-specific>

<branch-specific branch="master">
	Here's some code that would only work on the *master* branch.
</branch-specific>

<!-- In case there's another branch other than stable and master -->
<branch-specific :branch="['stable', 'master']">
	Here's some code that would work on both the *stable* and *master* branches, but not any other branch.
</branch-specific>
```

Because 90% of the majority of pages and their contents will stay the same. In the cases of pages that only apply to one branch and not the other, we can do something like this.

```vue
<branch except="master">
	<p class="warning">This page only applies to the *master* branch. Click [here](#) to switch to it.</p>
</branch>
```

At the moment, I think this is the most decent approach, so that we don't have to have multiple files for the same page, just on different branches. You'd only need to change one file when makes edits, and we could easily update the contents of the page with Vue without having to redirect the user to another page, refresh the page, or do anything else complex.

However, if anybody else has some ideas, I'll gladly hear them out. 😄

iCrawl:
+1

There is absolutely no reason to duplicate pages or "switch" sites like we do on the docs site.
I think this is a great middleground to hit off on both the new "breaking" stuff and the old stable (which I'm sure quite a lot of people still use for the sake of stability).
There also shouldn't be many topics that require a "master only" tag.

Overall this should also make it easier for people to update from one version to another if the need ever occurs. Thus further reducing the thousands of articles and tags about "change X to Y on master".

NbOpposite:
I wonder, wouldn't a version-specific component make more sense?
Like, you specify semvers that this applies for the current topic rather than a git branch, as at some point, master will become version 12, and then, when we start work on the next major release, the docs with master-branch set will suddenly not apply to the master, but rather version 12?

Danktuary:
Good point, @NbOpposite. The component example I gave would stay roughly the same, but the dropdown menu could look something like this:

![Version chooser dropdown menu](https://i.imgur.com/ZOBkSO4.png)

[...]

PR #125, opened 2018-08-12
Add OAuth2 Page

rei2hu:
Add a page for getting an access token for OAuth2 through Discord.

This is a rough draft, feedback appreciated.


guide/oauth2/README.md
```diff
@@ -0,0 +1,200 @@
...
+const port = 53134;
+
+http.createServer((req, res) => {
+	let responseCode = 404;
+	let content = '404 Error';
+	if (req.url === '/') {
+		responseCode = 200;
+		content = fs.readFileSync('./index.html');
+	}
+	res.writeHead(responseCode, {
```
Danktuary: This piece of code + some other areas should be spaced out a bit more, for consistency/readability reasons. Here's a good example of what I mean: https://github.com/discordjs/guide/blob/master/guide/assets/js/dark-theme.js#L13-L31

In this case, just add some blank lines before/after the if statement and after the `.writeHead()` call.

guide/oauth2/README.md
```diff
@@ -0,0 +1,200 @@
...
+}).listen(port);
+```
+
+Right now, we've designated that the contents of an `index.html` file will be served to the user when they visit the root domain, so let's create an `index.html` file in the same directory with the following contents.
+
+```html
+<!DOCTYPE html>
+<div id='info'>
+    Hoi!
+</div>
```
Danktuary: Needs some `<head>` and `<body>` tags and tag attributes should use double quotes. The `<head>` tag can contain just a `<title>`.

guide/oauth2/README.md
```diff
@@ -0,0 +1,200 @@
...
+```
+
+You can start your server with `node index.js`. Once you start it, try connecting to [localhost](http://localhost:53134) and you should see Hoi!
```
Danktuary: I think the localhost link can just be a direct link instead of a masked link. Not a big deal tho.

guide/oauth2/README.md
```diff
@@ -0,0 +1,200 @@
...
+			const [, access_token, token_type] = match;
+			const xml = new XMLHttpRequest();
+			xml.onload = () => {
+				const { username, discriminator } = JSON.parse(xml.response);
+				document.getElementById('info').innerText += `, ${username}#${discriminator}`;
+			}
+			xml.open('GET', 'https://discordapp.com/api/users/@me');   
+			xml.setRequestHeader('authorization', token_type + ' ' + access_token);
+			xml.send();
+		} else {
```
Danktuary: if/else block style should be consistent with the one the guide uses, as in:
```js
if (true) {
	// ...
}
else {
	// ...
}
```

guide/oauth2/README.md
```diff
@@ -0,0 +1,200 @@
...
+<a id='login' style='display:none;' href='your oauth2 url here'>Identify Yourself</a>
+<script>
+	window.onload = () => {
+		const match = window.location.hash.match(/access_token=(.+?)&token_type=(.+?)(?:&|$)/);
+		if (match) {
+			const [, access_token, token_type] = match;
+			const xml = new XMLHttpRequest();
+			xml.onload = () => {
+				const { username, discriminator } = JSON.parse(xml.response);
+				document.getElementById('info').innerText += `, ${username}#${discriminator}`;
```
Danktuary: When you authorize the app, the text shows up as "Hoi!, User#1234", so I think you should either entirely replace the `.innerText` (to make it display "Hoi, User#1234!"), or append to it without the comma.

guide/oauth2/README.md
```diff
@@ -0,0 +1,200 @@
...
+	}
+	return randStr;
+}
+
+// ...
+
+// generate and store the string
+const randStr = generateRandomString();
+localStorage.setItem('stateParameter', randStr);
+oauthURL += `&state=${btoa(randStr)}`;
```
Danktuary: `oauthURL` is currently undefined as of right here. If someone tries to follow the guide, reading all the texts and using the code from the codeblocks and then get an error when they run the script, it seems a bit misleading/incomplete, which is why we try to avoid that in all the other pages. 

To avoid hardcoding the URL in a string or something, you can just do something like this for the next line:

```js
 document.getElementById('login').href += `&state=${btoa(randStr)}`;
```

guide/oauth2/README.md
```diff
@@ -0,0 +1,200 @@
...
+document.getElementById('login').href = oauthURL;
+```
+
+When you visit a url with a state parameter appended to it and then click `Authorize`, you'll notice that after being redirected, the url will also have the state parameter appended to it, which you should then check against what was stored.
+
+```js
+const stateParameter = localStorage.getItem('stateParameter');
+if (btoa(stateParameter) !== urlStateParameter) {
```
Danktuary: `urlStateParameter` is also undefined, but in this case you can just use a string summarizing what it's supposed to be

guide/oauth2/README.md
```diff
@@ -0,0 +1,200 @@
...
+		console.log(`The access code is: ${accessCode[1]}`);
+	}
+}
+if (url === '/') {
+	responseCode = 200;
+	content = fs.readFileSync('./index.html');
+}
+```
+
+Now, that you have to exchange this code with Discord for an access token. To do this, you need your `client_id` and `client_secret`. If you've forgotten them, head over to [your applications](https://discordapp.com/developers/applications) and get them. We'll be using `node-fetch` along with `form-data` to make our request to Discord; `node-fetch` should be bundled with the latest versions of Discord.js, and you can install `form-data` with `npm i form-data`.
```
Danktuary: Regarding: "`node-fetch` should be bundled with the latest versions of Discord.js"
This isn't exactly true because the guide bases its content based on the stable branch, so the user should still be told to install `node-fetch`.

This'll be a non-issue once #121 becomes a thing. Soon:tm:

guide/oauth2/README.md
```diff
@@ -0,0 +1,200 @@
+# OAuth2 
+
+OAuth2 enables application developers to build applications that utilize authentication and data from the Discord API.
```
Danktuary: The intro section is a bit bland. I think adding another sentence saying something along the lines of "It can be used to create things such as web dashboard to display user info, statistics, and allow extended functionality to your bot" would help interest the user.

PR #128, opened 2018-08-19
Add guide page about manually parsing Discord mentions

vzwGrey:


guide/_sidebar.md
```diff
@@ -13,6 +13,7 @@
 	* [Configuration files](/creating-your-bot/configuration-files)
 	* [Adding more commands](/creating-your-bot/adding-more-commands)
 	* [Commands with user input](/creating-your-bot/commands-with-user-input)
+	* [Parsing mention arguments](/creating-your-bot/parsing-mention-arguments)
```
Danktuary: I don't think this should be put under this section of the sidebar, because when people read the guide in order, they'll come to this page and think it's necessary for them to read when it's a bit more advanced than the beginner level feel of the rest of the pages before it.

It doesn't exactly fit under "Popular topics", but maybe we can take the "Common questions" and "Miscellaneous examples" sections, put them under a section called "Miscellaneous", and stick this under it?

guide/creating-your-bot/parsing-mention-arguments.md
```diff
@@ -0,0 +1,219 @@
...
+So, how do you actually use this new information for your bot?  
+Most of your code will not change, however instead of using `message.mentions` to find the mentioned users you will have to do it manually.  
+This may sound scary at first, but once you see the code you will see it is actually pretty simple.
+
+Say you already have a simple command handler like this:
+
+```js
+client.on('message', message => {
+	// Don not go any further if there is no command prefix.
+	if (!message.content.startsWith(config.prefix)) return;
```
Danktuary: I think explaining every single line of code, especially ones like this that are self-explanatory if you read them normally, is fairly redundant. Usually it reads better/feels less cluttered when you explain what the code is doing afterwards, outside of the codeblock. And even then, you can just give a general feel of it and not a line-by-line explanation.

Danktuary: Though, as another note, I think since this is basically just skeleton code, it's doesn't necessarily need any explanation. The code that needs to be explained is mainly the mention parsing.

guide/creating-your-bot/parsing-mention-arguments.md
```diff
@@ -0,0 +1,219 @@
...
+	const split = withoutPrefix.split(/ +/);
+	// The first element is the command, the rest are the arguments.
+	const command = split[0];
+	const args = split.slice(1);
+
+	// ....Some commands here....
+});
+```
+
+Now you can easily test the waters by upgrading the avatar command from last time.
```
Danktuary: "from last time" might be a bit unclear to those randomly landing on this page from somewhere else (e.g. a tag in the d.js server), so maybe a link linking to the previous command implementation would be nice.

guide/creating-your-bot/parsing-mention-arguments.md
```diff
@@ -0,0 +1,219 @@
...
+This is what we have so far. It is pretty simple, it will show the avatar of who used the command.
+
+```js
+if (command === 'avatar') {
+	// What to put here to use the mention?
```
Danktuary: No need for this comment

guide/creating-your-bot/parsing-mention-arguments.md
```diff
@@ -0,0 +1,219 @@
...
+```
+
+But how do you actually get the correct user now? Well, this requires a few simple steps.  
+Putting it into a function will make it easily reusable. We will use the name `getUserFromMention` here.
+
+```js
+function getUserFromMention(mention) {
+	// Check if `mention` contains a string.
+	// If not, just return undefined to signal there is no user to be found.
+	if (!mention) return undefined;
```
Danktuary: Same for this area; if you read the code, you'll understand what's going on faster than if you would read the comments.

Danktuary: Also not sure if we should be using `return undefined`, just `return`, or `return null`. For some reason I feel like null is more appropriate, but it's not a big deal either way.

guide/creating-your-bot/parsing-mention-arguments.md
```diff
@@ -0,0 +1,219 @@
...
+
+	// A user mention starts with <@ as well as a > at the end,
```
Danktuary: You explained this in one of the paragraphs above, so it doesn't need to be explained here again. Same for a few more comments in this codeblock section

guide/creating-your-bot/parsing-mention-arguments.md
```diff
@@ -0,0 +1,219 @@
...
+
+		// Now all that is left should be the ID,
+		// which can be used to get the user object.
+		const user = client.users.get(mention);
+		// The ID might be invalid, so check for that as well.
+		if (!user) {
+			// Just return undefined to indicate nothing was found.
+			return undefined;
+		}
+		else {
```
Danktuary: No need for the `else`. I think it can be simplified to just `if (user) return user;` and then maybe omit the `return undefined` entirely and let it hit the last line in the function.

Or maybe even just `return client.users.get(mention);`, because `.get()` returns undefined if nothing was found. In this case, if that were done, then a comment mentioning that would be okay, I think.

guide/creating-your-bot/parsing-mention-arguments.md
```diff
@@ -0,0 +1,219 @@
...
+		}
+	}
+
+	// Return undefined here as well.
+	// If the code reaches this point nothing was found.
+	return undefined;
+}
+```
+
+<tip>The `slice` method is used in a more advance way here. You can read the [MDN documentation](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/slice) for more info.</tip>
```
Danktuary: "`slice`" should be "`.slice()`", for consistency reasons.

guide/creating-your-bot/parsing-mention-arguments.md
```diff
@@ -0,0 +1,219 @@
...
+		// If he did try to turn that mention into a user object.
+		// The first argument should be the mention, so pass it to the function.
+		const user = getUserFromMention(args[0]);
+
+		// The function can return undefined, in which case notify
+		// the user that the argument was invalid.
+		if (!user) {
+			return message.reply('Please use a proper mention if you want to see someone else\'s avatar.');
+		}
+		else {
```
Danktuary: No need for `else` here or below this one, since there's a `return` in the if statement and nothing else below the `else` blocks.

guide/creating-your-bot/parsing-mention-arguments.md
```diff
@@ -0,0 +1,219 @@
...
+Lo and behold, it does work.
+
+So now, instead of using `message.mentions` you can use your new, fantastic function.
+This will allow you to add proper checks for all your args, so that you can tell when a command was used correctly and when it was used incorrectly.
+
+But this does not mark the end of the page. If you feel adventurous you can read on and learn how to use Regular Expressions to easily convert a mention into a user object in just two lines.
+
+### Using Regular Expressions
+
+Previously you learn how to use rudimentary string related functions to turn the special mention syntax Discord uses into a proper Discord.js User object.
```
Danktuary: "Discord.js" should be all lowercase.

guide/creating-your-bot/parsing-mention-arguments.md
```diff
@@ -0,0 +1,219 @@
...
+But using Regular Expressions (aka "RegEx" or "RegExp"), you can condense all that logic into a single line! Crazy, right?
+
+If you have never worked with Regular Expressions before, this might seem daunting. But in fact, you already have used regular expressions. Remember `withoutPrefix.split(/ +/);`? This little `/ +/` is actually a Regular Expression. The `/` on either side tell JavaScript where the Regular Expression begins and where it ends, the stuff inbetween is it is content. 
+
+<tip>For a more detailed explanation of please consult the [MDN's documentation](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/RegExp) about them.</tip>
```
Danktuary: Small wording error over here

guide/creating-your-bot/parsing-mention-arguments.md
```diff
@@ -0,0 +1,219 @@
...
+The RegEx you will use for user mentions will look like this: `/^<@!?(\d+)>$/`.
+Here is how the RegEx works:
+
+ 1. The `^` at the beginning and the `$` at the end mean that the mention has to take up the entire string.
+ 2. You have the typical `<@` and `>` at the beginning and end.
+ 3. The `?` after the `!` indicates that the `!` is optional.
+ 4. `\d+` means the RegEx will look for multiple digits, which will be the ID.
+ 5. The parentheses around the `\d+` create a capture group, which allows us to get the ID out of the mention.
+
+Using the `match` method on strings you can get the values of the capture group, i.e., the ID of the mention.
```
Danktuary: "`match`" should be "`.match()`" for consistency reason.

guide/creating-your-bot/parsing-mention-arguments.md
```diff
@@ -0,0 +1,219 @@
...
+Updating your `getUserFromMention` function to use RegEx gives you this:
+
+```js
+function getUserFromMention(mention) {
+	const matches = mention.match(/^<@!?(\d+)>$/);
+	// The id is the first and only match found by the RegEx.
+	// However the first element in the matches array will be the entire mention, not just the ID,
+	// so use index 1.
+	const id = matches[1];
+	// Now just get the User with the ID and return that.
```
Danktuary: I don't think this comment is entirely necessary, but the ones above it explaining the RegEx stuff are okay. 👍

PR #138, opened 2018-09-26
The command comparison now uses the config prefix

J-888:
Since this is the code sample for the config.json tutorial, it should support any prefix specified there


xDimGG:
You should use template literals instead of concatenating.

Danktuary:
The reason it wasn't already like this is because in the [next page](https://discordjs.guide/#/creating-your-bot/adding-more-commands) it talks about using the prefix from the config file near the top, so it shouldn't be changed in this code sample.

PR #142, opened 2018-10-01
add async functions to sharding examples

Lewdcario:
Came across this in a support channel the other day and figured it'd be nice to demonstrate.

### DIFF
--- a/guide/sharding/additional-information.md
+++ b/guide/sharding/additional-information.md
@@ -70,6 +70,38 @@ client.shard.broadcastEval(`this.${funcName}(${args});`);
 
 This would become `client.funcName(args)` once it gets through. This is handy if you, for example, have extended your client object with your own class and wish to call some of its methods manually.
 
+#### Asynchronous Functions
+
+There may be a time when you want to have your shard process an asynchronous function. Here's how you can do that!
+
+```js
+client.shard.broadcastEval(`
+	let channel = this.channels.get('id');
+	let msg;
+	if (channel) {
+		msg = channel.fetchMessage('id').then(m => m.id);
+	}
+	msg;
+`);
+```
+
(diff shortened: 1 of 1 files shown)


guide/sharding/additional-information.md
```diff
@@ -70,6 +70,38 @@ client.shard.broadcastEval(`this.${funcName}(${args});`);
...
+There may be a time when you want to have your shard process an asynchronous function. Here's how you can do that!
+
+```js
+client.shard.broadcastEval(`
+	let channel = this.channels.get('id');
+	let msg;
+	if (channel) {
+		msg = channel.fetchMessage('id').then(m => m.id);
+	}
+	msg;
```
Drahcirius: Missing return here. Also, couldn't this be `if (channel) return channel.fetchMessage('id').then(m => m.id);` ? Same with below snippet. Though I think if they're at the point of sharding they shouldn't really be handling promises like this, but using await/async only. Unless they're using lower node versions. But master branch doesn't even support that anymore, so maybe we should refrain from using promise-based examples in the more advanced sections?

guide/sharding/additional-information.md
```diff
@@ -70,6 +70,38 @@ client.shard.broadcastEval(`this.${funcName}(${args});`);
...
+
+This snippet allows you to return fetched messages outside of the `broadcastEval`, allowing you to know whether or not you were able to retrieve a message, for example. Remember, you aren't able to return entire objects outside. Now, what if we wanted to use `async/await` syntax inside?
+
+```js
+client.shard.broadcastEval(`
+	(async => {
+		let channel = this.channels.get('id');
+		let msg;
+		if (channel) {
+			msg = await channel.fetchMessage('id').then(m => m.id);
```
Drahcirius: If we're using await here, why not
```js
let msg = await channel.fetchMessage(<id>);
return msg.id;
```

PR #144, opened 2018-10-07
Add page about useful packages.

vzwGrey:


guide/miscellaneous/useful-packages.md
```diff
@@ -0,0 +1,189 @@
...
+
+### moment.js
+
+<tip>Official documentation: https://momentjs.com/docs/</tip>
+
+Moment is a powerful package for working with dates in JavaScript.  
+It allows you to quickly and easily format dates in any way you want or parse strings back into JavaScript Date objects.  
+There are even some extensions for it to allow you work with durations and more.
+
+For example if you want to wanted to ask your users to give the bot a date,
```
SpaceEEC: "if you want to wanted to ask"

guide/miscellaneous/useful-packages.md
```diff
@@ -0,0 +1,189 @@
...
+
+### Table of contents
+
+ 1. moment.js
+ 2. ms
+ 3. common-tags
+ 4. chalk
+ 5. winston
+
+### moment.js
```
iCrawl: Should be maybe switch this out for https://date-fns.org/ ?

Yes moment.js might be the more commonly used package, but we might as well point to one that has very similar syntax but performs better in almost all points.

robflop: I actually brought this up in private conversation and wanted to offer writing a section about possibly using dateFns instead, so i'm all for it.

Danktuary:
Also forgot to mention a sidebar link is gonna be needed (my fault; I told Grey not to add one a while ago 😅).

PR #145, opened 2018-10-07
Add missing semicolon in CONTRIBUTING.md.

dragonfire535:
^ Title. It bothered me.


Danktuary:
This is intentional, though. The very next sentence states:
> ESLint would error with `Parsing error: Unexpected token message` instead of letting you know that you're missing a semicolon.

It's there to give an example of when and why you'll need to use `<!-- eslint-skip -->`.

PR #148, opened 2018-10-18
minor edits to Adding Features, refactor cooldown section

appleJax:
**Changes:**
- minor edits to the "Adding features" page
- slight refactor of Cooldowns section to simplify it

>**PREVIEW:**

### Cooldowns

Spam is something you generally want to avoid - especially if one of your commands requires calls to other APIs, or takes a bit of time to build/send. This is also a very common feature bot developers want to integrate into their projects, so let's get started on that!

You'll be using the ping command to test this on. Add in the following bit of code:

```diff
module.exports = {
  name: 'ping',
+ cooldown: 5,
  execute(message) {
    message.channel.send('Pong.');
  },
};
```

This is the amount (in seconds) that the user will have to wait before being able to properly use that command again. You'll be using Collections again to store what you need.

In your main file, add in this line (preferably somewhere above your `ready` event):

```js
const cooldowns = new Discord.Collection();
```

Again in your main file, directly above the `try/catch`, add in the following:

```js
if (!cooldowns.has(command.name)) {
  cooldowns.set(command.name, new Discord.Collection());
}

const now = Date.now();
const timestamps = cooldowns.get(command.name);
const cooldownAmount = (command.cooldown || 3) * 1000;

if (timestamps.has(message.author.id)) {
  // ...
}
```

You check if the `cooldowns` Collection has the command set in it yet. If not, then add it in. Next, 3 variables are created:

1. A variable with the current timestamp.
2. A variable that `.get()`s the Collection for the triggered command.
3. A variable that gets the necessary cooldown amount. If you don't supply it in your command file, it'll default to 3. Afterwards, convert it to the proper amount of milliseconds.

After that, create a simple `if` statement to check if the `timestamps` Collection has the author ID in it yet.

Continuing with your current setup, this is the complete `if` statement:

```js
if (timestamps.has(message.author.id)) {
  const expirationTime = timestamps.get(message.author.id) + cooldownAmount;

  if (now < expirationTime) {
    const timeLeft = (expirationTime - now) / 1000;
    return message.reply(`please wait ${timeLeft.toFixed(1)} more second(s) before reusing the \`${command.name}\` command.`);
  }
}
```

Since the `timestamps` Collection has the author ID in it, you `.get()` it and then sum it up with the `cooldownAmount` variable, in order to get the correct expiration timestamp. You then check to see if it's actually expired or not.

If the `expirationTime` has not passed, you return a message letting the user know how much time is left until they can use that command again. As you will see shortly, the author ID should be deleted from the `timestamps` Collection upon expiration, but you should take this extra precaution to avoid potential UX problems.

Finally, if the `timestamps` Collection doesn't have the message author's ID (or if the author ID did not get deleted as planned), `.set()` the author ID with the current timestamp and create a `setTimeout()` to automatically delete it after the cooldown period has passed:

```js
timestamps.set(message.author.id, now);
setTimeout(() => timestamps.delete(message.author.id), cooldownAmount);
```

### DIFF
--- a/guide/command-handling/adding-features.md
+++ b/guide/command-handling/adding-features.md
@@ -23,9 +23,9 @@ The command handler you've been building so far doesn't do much aside from dynam
 In this short (but necessary) refactor, you:
 
 1. Renamed the original `command` variable to `commandName` (to be descriptive about what it actually is, and so that you can name one of our variables as `command` later).
-2. Changed the following if statement appropriately.
+2. Changed the following `if` statement appropriately.
 3. Made a variable named `command` which is the actual command object.
-4. Changed the line inside the try/catch statement to use the `command` variable instead.
+4. Changed the line inside the `try/catch` statement to use the `command` variable instead.
 
 Now you can start adding features!
 
@@ -50,7 +50,7 @@ module.exports = {
 };
 ```
 
-This is fine if you only have a few commands that require arguments. However, if you plan on making a lot of commands and don't want to copy & paste that if statement each time, it'd be a smart idea to change that check into something simpler.
+This is fine if you only have a few commands that require arguments. However, if you plan on making a lot of commands and don't want to copy & paste that `if` statement each time, it'd be a smart idea to change that check into something simpler.
 
 Here are the changes you'll be making:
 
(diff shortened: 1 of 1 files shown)


Drahcirius:
Can you change the file in the `/code-samples/command-handling/adding-features` folder to reflect these changes? Otherwise this looks good to me.

PR #151, opened 2018-10-26
Small style fix for `client.on('message')`

vzwGrey:
Since no one should actually have `client.on('message')` in their code I thought I'd just change it to ``client.on('message', ...)` so it's a bit more clear.


Danktuary:
I think `client.on('event', ...)` is an okay way to style it when referring to something in the user's code. It's just so that the reader has an idea of what we're talking about.

PR #153, opened 2018-10-27
Add fetchClientValues and combined stats command

Gawdl3y:
This adds info about the fetchClientProperties method, overhauls the broadcastEval example to actually need it, and combines the two for the resulting stats command.


code-samples/sharding/getting-started/bot.js
```diff
@@ -10,9 +10,16 @@ client.on('message', message => {
 	const command = args.shift().toLowerCase();
 
 	if (command === 'stats') {
-		return client.shard.broadcastEval('this.guilds.size')
+		const promises = [
+			client.shard.fetchClientProperties('guilds.size'),
```
SpaceEEC: You missed one. 👀

PR #156, opened 2018-11-09
Keyv

appellation:
Adds a page for per-guild prefixes using Keyv.

### DIFF
--- a/guide/_sidebar.md
+++ b/guide/_sidebar.md
@@ -36,7 +36,7 @@
 	* **Database**
 	* [Storing data with Sequelize](/sequelize/)
 	* [Making a currency system](/sequelize/currency)
-	* [Building a per-guild prefix system](/keyv)
+	* [Building a per-guild prefix system](/keyv/)
 
 	* **Frameworks**
 	* [Commando](/commando/)
--- a/guide/_sidebar.md
+++ b/guide/_sidebar.md
@@ -36,7 +36,7 @@
 	* **Database**
 	* [Storing data with Sequelize](/sequelize/)
 	* [Making a currency system](/sequelize/currency)
-	* [Building a per-guild prefix system](/keyv/)
+	* [Working with Keyv](/keyv/)
 
 	* **Frameworks**
 	* [Commando](/commando/)
--- a/guide/_sidebar.md
+++ b/guide/_sidebar.md
@@ -36,6 +36,7 @@
 	* **Database**
 	* [Storing data with Sequelize](/sequelize/)
 	* [Making a currency system](/sequelize/currency)
+	* [Building a per-guild prefix system](/keyv)
 
 	* **Frameworks**
 	* [Commando](/commando/)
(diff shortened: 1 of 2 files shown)


guide/_sidebar.md
```diff
@@ -36,6 +36,7 @@
 	* **Database**
 	* [Storing data with Sequelize](/sequelize/)
 	* [Making a currency system](/sequelize/currency)
+	* [Building a per-guild prefix system](/keyv/)
```
Danktuary: I think the sidebar link should be labeled differently, because if a user reads the first 2 pages about Sequelize and then clicks on this one, they'd assume it continues using Sequelize. And most of the page is more about using Keyv than just building a per-guild prefix system.

```suggestion
	* [Working with Keyv](/keyv/)
```

guide/keyv/README.md
```diff
@@ -0,0 +1,168 @@
...
+const prefixes = new Keyv('sqlite://path/to.sqlite');
+const globalPrefix = '.';
+```
+
+### Command handler
+
+This guide uses a very basic command handler with some added complexity to allow for multiple prefixes. For more a more robust command handler, look at the [command handling](/command-handling) guide.
+
+```js
+client.on('message', async (message) => {
```
Danktuary:
```suggestion
client.on('message', async message => {
```

guide/keyv/README.md
```diff
@@ -0,0 +1,168 @@
...
+		}
+		else {
+			// check the guild-level prefix
+			const guildPrefix = await prefixes.get(message.guild.id);
+			if (message.content.startsWith(guildPrefix)) prefix = guildPrefix;
+		}
+
+		// if we found a prefix, setup args; otherwise, this isn't a command
+		if (prefix) args = message.content.slice(prefix.length).split(/\s+/);
+		else return;
```
Danktuary: Not a fan of `else return` and long one-line if statements. I'd be more consistent if it were written as
```js
if (!prefix) return;
args = message.content.slice(prefix.length).split(/\s+/);

File: guide/keyv/README.md [hunk 4]
```diff
@@ -0,0 +1,168 @@
...
+```js
+client.on('message', async (message) => {
+	if (message.author.bot) return;
+
+	let args;
+	// handle messages in a guild
+	if (message.guild) {
+		let prefix;
+
+		// check if the message starts with the global prefix
```
Danktuary: Comments like this aren't necessary, since reading the line directly isn't confusing and reads just about the same as the sentence written above.

guide/keyv/README.md
```diff
@@ -0,0 +1,168 @@
...
+		// if we found a prefix, setup args; otherwise, this isn't a command
+		if (prefix) args = message.content.slice(prefix.length).split(/\s+/);
+		else return;
+	}
+	else {
+		// handle DMs
+		let slice = 0;
+
+		// check if they used the global prefix
+		if (message.content.startsWith(prefix)) slice = globalPrefix.length;
```
Danktuary: This could combined with the line above into a ternary.

```js
const slice = message.content.startsWith(prefix) ? globalPrefix.length : 0;
```

appellation:
Thanks for the review! I think I addressed everything you commented on.

PR #165, opened 2019-01-24
Adding the reaction to an uncached message

izexi:
Reacting to an uncached message would cache the message via:
https://github.com/discordjs/guide/blob/9fa60ac09b75d479e64cc1b89c5676fca558a7a6/code-samples/popular-topics/reactions/raw-event.js#L22

Then, when you unreact to that same message, it will not fire the `messageReactionRemove` event through the `raw` as it has been cached. It will also not be fired normally as [`reaction`](https://github.com/discordjs/discord.js/blob/stable/src/client/actions/MessageReactionRemove.js#L23) is [`null`](https://github.com/discordjs/discord.js/blob/stable/src/structures/Message.js#L596), because the reaction hasn't actually been added to the message.

A workaround to this issue would be to call [`Message#_addReaction(emoji, user)`](https://github.com/discordjs/discord.js/blob/stable/src/structures/Message.js#L567-L582) when an uncached message has been reacted to, which is done in this PR.

### DIFF
--- a/code-samples/popular-topics/reactions/raw-event.js
+++ b/code-samples/popular-topics/reactions/raw-event.js
@@ -28,6 +28,8 @@ client.on('raw', async event => {
 		reaction = new Discord.MessageReaction(message, emoji, 1, data.user_id === client.user.id);
 	}
 
+	if (event.t === 'MESSAGE_REACTION_ADD') message._addReaction(reaction.emoji, user);
+
 	client.emit(events[event.t], reaction, user);
 });
 


almostSouji:
First of all, thanks for your patience!
After https://github.com/discordjs/guide/pull/104#issuecomment-423777424 we decided to not further mess with the raw event but instead put the red warning banner at the beginning of <https://discordjs.guide/popular-topics/reactions.html#listening-for-reactions-on-old-messages>.

Due to your PR we discussed the handling of the "raw" event again now and came to the following conclusion:

At the end of the day the "raw" event is an unsupported and undocumented API and as such shouldn't be the primary way to address uncached reactions, especially since we keep piling workaround on top of workaround to satisfy specific needs. 

Instead we are going to provide a proper guide on handling reactions on uncached messages with partials once it lands on a stable version of the library.

PR #166, opened 2019-01-25

conver4y:
An added custom prefix in Miscellaneous example.


Drahcirius:
Hi, there already is a section in the guide that covers custom prefixes using [keyv](https://discordjs.guide/keyv/#command-handler) and a simple command handler. I would prefer to not confuse people by giving examples using different key-value packages, so if you would like to contribute something to the keyv section or change this to a link to it instead, that would be much preferred.

Issue #100, opened 2018-05-25
Add information on how to securely store API keys/pass


TallGuysFreeADMIN:
Multiple servers were taken over recently. Many of them followed discordjs' guide on how to connect a bot.  While looking it over, I see no information or warning to keep your credentials secure (.gitignore, system variable, pass thru commandline, etc). 
People are hardcoding keys into public git repos and being taken over. Please consider adding a warning and ways to securely store auth keys.

Danktuary:
In [one of the very first few pages of the guide](https://discordjs.guide/#/preparations/setting-up-a-bot-application?id=your-token), there's a section explaining everything about your token. What it is, why you should keep it safe, and an example of what can happen if someone gets a hold of your token. Under the "What is a token, anyway?" section, all of this is very clearly stated.

> A token is essentially your bot's password; it's what your bot uses to login to Discord. With that being said, **it is vital that you do not ever share this token with anybody, purposely or accidentally**. If someone does manage to get a hold of your token, they can use your bot as if it were theirs—this means they can perform malicious acts with it.

Committing your token to your bot's repo is essentially the same as committing your day-to-day password to any other repo you make.

If people have skipped over this section of guide, that's not our fault; the content is there and it's up to the reader to choose if they want to read it or skip over it. I'd also like to add that we don't mention Git anywhere in the guide, which is why there's no mention of using `.gitignore`. As soon as you start using Git, it's up to you to figure out how to handle these things.

With all this being said, if you wish, you're very welcome to open a PR to add some sort of information about keeping them safe. If you choose to do so, I'd recommend putting it on the [configuration files](https://discordjs.guide/#/creating-your-bot/configuration-files) page.

orderbs:
Some points to consider:

- Given the scale and nature of this attack, the status quo is perhaps insufficient.
- Many people using this guide are inexperienced (which is why they're using a tutorial), so despite the very high quality of the guide, some will have almost no real understanding of the tutorial code they might upload.
- At multiple points in the guide, one is linked here (i.e. GitHub), presenting the temptation to just fork/copy this repo and swap in one's token.  This seems less of a case where people are "skipping over" warnings but rather not fully comprehending what it means to upload unsanitized tutorial code to a public repo (like this).
- Obvious is in the eye of the beholder.
- It's not your fault (it's not your fault).
- Feedback is important, especially when invited.

Thanks for all your hard work, this is one of best-written guides out there.

Cheers!

Danktuary:
[...]

PR #167, opened 2019-01-26
Added hosting with heroku page.

jsmiith:
Added a tutorial on how to start Hosting with Heroku. Thinking on adding one for glitch.com too, amongst other things.

### DIFF
--- a/guide/.vuepress/sidebar.js
+++ b/guide/.vuepress/sidebar.js
@@ -121,6 +121,12 @@ module.exports = {
 			],
 		},
 		{
+			title: 'Hosting',
+			children: [
+				'/hosting/heroku'
+			]
+		},
+		{
 			title: 'Additional Information',
 			children: [
 				'/additional-info/notation',
--- a/guide/hosting/heroku.md
+++ b/guide/hosting/heroku.md
@@ -1,0 +1,112 @@
+# Hosting with Heroku
+
+## Intro
+
+Heroku is a container-based cloud **Platform as a Service** (PaaS). Developers use Heroku to deploy, manage, and scale modern apps. What this means, is that you can host your code on it so that your bot is running **all the time**. They offer a **free** plan which has a limit of **550 hours a month** across all of the applications on your account, which you shall be utilizing today (unless you want to throw in some money and get a paid tier).
+
+## Prerequisites
+
+Before you get started, you need to make sure that you have:
+
+- [installed](https://git-scm.com/book/en/v2/Getting-Started-Installing-Git) and [configured](https://git-scm.com/book/en/v2/Getting-Started-First-Time-Git-Setup) Git on your local machine
+- [installed](https://devcenter.heroku.com/articles/heroku-cli) the Heroku CLI
+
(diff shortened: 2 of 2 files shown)


Drahcirius:
This tutorial is nice and simple, however it doesn't really fit with the rest of the guide. There are a bunch of sections in the guide where we expect you to be able to ssh into the server and run commands. Heroku and other free services intentionally disallow you from doing that, meaning people following this page cannot follow the instructions in some of the other pages of the guide.

Aside from that, the biggest concern that I can see is that if they follow the page in https://discordjs.guide/creating-your-bot/configuration-files.html and they also follow your instruction to run `git add .` they will be committing their config file into their repo which is a security concern that was brought up in https://github.com/discordjs/guide/issues/100. If they had ssh or sftp, they could commit their code but upload their config file manually so that their token doesn't get added to their repository. 

I think in order for a tutorial like this to be added to the guide, it would need to at least have ssh access and be somewhat hosting-platform agnostic. Then we can say that the same instructions will work on digitalocean, vultr, linode, etc.

noneissome:
> This tutorial is nice and simple, however it doesn't really fit with the rest of the guide. There are a bunch of sections in the guide where we expect you to be able to ssh into the server and run commands. Heroku and other free services intentionally disallow you from doing that, meaning people following this page cannot follow the instructions in some of the other pages of the guide.
> 
> Aside from that, the biggest concern that I can see is that if they follow the page in https://discordjs.guide/creating-your-bot/configuration-files.html and they also follow your instruction to run `git add .` they will be committing their config file into their repo which is a security concern that was brought up in #100. If they had ssh or sftp, they could commit their code but upload their config file manually so that their token doesn't get added to their repository.
> 
> I think in order for a tutorial like this to be added to the guide, it would need to at least have ssh access and be somewhat hosting-platform agnostic. Then we can say that the same instructions will work on digitalocean, vultr, linode, etc.

True. It definitely needs some work.

almostSouji:
Going to close this for now due to lack of activity, feel free to reopen if you want to address the issues pointed out by Drahcirius though!

Monoverde888:
> 
> 
> This tutorial is nice and simple, however it doesn't really fit with the rest of the guide. There are a bunch of sections in the guide where we expect you to be able to ssh into the server and run commands. Heroku and other free services intentionally disallow you from doing that, meaning people following this page cannot follow the instructions in some of the other pages of the guide.
> 
> Aside from that, the biggest concern that I can see is that if they follow the page in https://discordjs.guide/creating-your-bot/configuration-files.html and they also follow your instruction to run `git add .` they will be committing their config file into their repo which is a security concern that was brought up in #100. If they had ssh or sftp, they could commit their code but upload their config file manually so that their token doesn't get added to their repository.
> 
> I think in order for a tutorial like this to be added to the guide, it would need to at least have ssh access and be somewhat hosting-platform agnostic. Then we can say that the same instructions will work on digitalocean, vultr, linode, etc.

You CAN do some thing that is like ssh with the heroku cli
`heroku run bash`

PR #188, opened 2019-03-16
Fix the Mention prefix example

didinele:
Hello, 

Since I am not the biggest fan of Regex I never got to make a "prefix mention" myself, so after a quick Google search I stumbled upon the Discord Guide example and started using it. After a while, I was testing something inside of my Command Handler and changed the prefix of the bot to `r!`, in which, the given example on the Guide `r` gets escaped, completely breaking usage and forcing you to mention the bot. The escapes are obviously there so characters like `?` don't break it. After discussing with Space in the other-js-questions channel he recommended me to use the [`escape-string-regexp`](https://www.npmjs.com/package/escape-string-regexp) package.

So that is pretty much what this PR fixes, though I am unsure how to word a tip specifying what the used package does & specifying installation, let me know if I can add anything else.

### DIFF
--- a/guide/popular-topics/miscellaneous-examples.md
+++ b/guide/popular-topics/miscellaneous-examples.md
@@ -84,8 +84,10 @@ const Discord = require('discord.js');
 const client = new Discord.Client();
 const prefix = '!';
 
+const escapeRegex = require('escape-string-regexp');
+
 client.on('message', message => {
-	const prefixRegex = new RegExp(`^(<@!?${client.user.id}>|\\${prefix})\\s*`);
+	const prefixRegex = new RegExp(`^(<@!?${client.user.id}>|${escapeRegex(prefix)})\\s*`);
 	if (!prefixRegex.test(message.content)) return;
 
 	const [, matchedPrefix] = message.content.match(prefixRegex);
--- a/guide/popular-topics/miscellaneous-examples.md
+++ b/guide/popular-topics/miscellaneous-examples.md
@@ -76,6 +76,10 @@ process.on('unhandledRejection', error => console.error('Uncaught Promise Reject
 
 ## Mention prefix
 
+::: tip
+The `escapeRegex` function is used to convert special characters into literal characters by escaping them, so that they don't terminate the pattern within the [Regular Expression](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Regular_Expressions)!
+:::
+
 When a user adds your bot to their server, they may not immediately know what the prefix is. This is why it's a good idea to allow your bot to be triggered by either a set prefix, or by pinging them. In addition, it's also a good idea to have a command that displays the available prefixes.
 
 ```js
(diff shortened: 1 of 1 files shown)


tipakA:
Instead of using external module for it, i suggest escaping special regex characters with String.replace(), as mentioned [here](https://stackoverflow.com/a/3561711).
However, as example above is mutating global, it can be used as a function instead.

didinele:
> Instead of using external module for it, i suggest escaping special regex characters with String.replace(), as mentioned [here](https://stackoverflow.com/a/3561711).
> However, as example above is mutating global, it can be used as a function instead.

Yeah, you're right, I believe @tipakA's approach is better, I think we could do something like this:
```js
const Discord = require('discord.js');

const client = new Discord.Client();
const prefix = '!';

const escapeRegex = (str) => str.replace(/[-\/\\^$*+?.()|[\]{}]/g, '\\$&');

	const prefixRegex = new RegExp(`^(<@!?${client.user.id}>|${escapeRegex(prefix)})\\s*`);
	if (!prefixRegex.test(message.content)) return;

	const [, matchedPrefix] = message.content.match(prefixRegex);
	const args = message.content.slice(matchedPrefix.length).trim().split(/ +/);
	const command = args.shift();

	if (command === 'ping') {
		message.channel.send('Pong!');
	} else if (command === 'prefix') {
		message.reply(`you can either ping me or use \`${prefix}\` as my prefix.`);
	}
});

client.login('your-token-goes-here');
```

Once again, I am not sure how to explain what's actually going on & if this is a better approach so I will wait for more comments & opinions.

izexi:
Both `-` and `/` don't need to be escaped. Hyphens only needs to be escaped when encapsulated with `[]`, which is being escaped, slashes don't need to explicitly be escaped. So, the RegExp can be reduced to:
```js
/[.*+?^${}()|[\]\\]/g
```
This is provided on the `escapeRegExp` function located on the [MDN for Regular Expressions](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Regular_Expressions).

almostSouji:
Are there any updates on this?   
I'd much prefer a solution that doesn't need an additional module just for escaping as others have provided above.

Danktuary:
To add, I think it'd also be in place to increment a `::: tip` after the codeblock regarding why a regex needs to be used there, with a link to the source or something.

didinele:
Glad there's some finally some activity to this, I'll take a look tomorrow and maybe use Havoc's regex & add a tip

didinele:
Alright, I am not sure if I worded it correctly, if there's any more improvements I can make please let me know.

guide/popular-topics/miscellaneous-examples.md
```diff
@@ -104,6 +104,9 @@ client.on('message', message => {
 client.login('your-token-goes-here');
 ```
 
+::: tip
+The `escapeRegex` arrow function is a simple function used to remove special characters, that would normally result in an error, from a [regular expression](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Regular_Expressions)!
```
izexi:
```suggestion
The `escapeRegex` function is used to convert special characters into literal characters by escaping them, so that they don't terminate the pattern within the [Regular Expression](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Regular_Expressions)!
```

guide/popular-topics/miscellaneous-examples.md
```diff
@@ -76,6 +76,10 @@ process.on('unhandledRejection', error => console.error('Uncaught Promise Reject
 
 ## Mention prefix
 
+::: tip
+The `escapeRegex` function is used to convert special characters into literal characters by escaping them, so that they don't terminate the pattern within the [Regular Expression](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Regular_Expressions)!
```
Danktuary: The tip was fine were it was, below the codeblock and above the other tip, because here you're mentioning something that hasn't been shown to the reader yet.

PR #202, opened 2019-04-16
Update OAuth2 Page

rei2hu:
There was a change in the fragment Discord provides when redirecting the user back to the redirect url so the regular expression did not work anymore.

Changed it to use URLSearchParams which was originally brought up in the past but didn't work because Discord wouldn't encode/decode certain things in the fragment leading to problems but now they do.

Also fixed a quote and added a tiny bit to the example for consistency i.e. getting user info from both client side and server side code.

### DIFF
--- a/code-samples/oauth/simple-oauth-webserver/index.html
+++ b/code-samples/oauth/simple-oauth-webserver/index.html
@@ -21,20 +21,21 @@
 		}
 
 		window.onload = () => {
-			const match = window.location.hash.match(/access_token=(.+?)&token_type=(.+?)&state=(.+?)(?:&|$)/);
+			const fragment = new URLSearchParams(window.location.hash.slice(1));
 
-			if (match) {
-				const [, access_token, token_type, urlState] = match;
+			if (fragment.has("access_token")) {
+				const urlState = fragment.get("state");
 				const stateParameter = localStorage.getItem('stateParameter');
-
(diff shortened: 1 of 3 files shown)


rei2hu:
> One small bit that confused me is that after applying the changes the greeting no longer happened and it sticks to the "login" screen (same behavior for the filled out code sample).

What changes do you mean?

If you mean the fact that setting the `response_type` to `code` doesn't give a greeting while having it as `token` does in the example, I agree that is confusing.

The problem is going any further requires covering SSR and/or creating and associating session IDs with information stored server side. I think it might be worth it to break that part into a new section, if someone is willing to write it.

If you mean none of the examples are working, they both work for me.

almostSouji:
> If you mean the fact that setting the `response_type` to `code` doesn't give a greeting while having it as `token` does in the example, I agree that is confusing.

exactly that, but makes sense!
Maybe it'd be less confusing to have the two sample codes separated in code samples?
One working with greeting, one with the token response?

Noxillio:
I've noticed that this page says `#access_token=ACCESS_TOKEN&token_type=Bearer&expires_in=EXPIRATION&scope=identify` gets appended to the redirect, but I don't get them in this order. I end up getting `#token_type=Bearer&access_token=ACCESS_TOKEN&expires_in=EXPIRATION&scope=identify` appended to the end of mine.

PR #206, opened 2019-04-22
Removed erroneous warning on not using hex-color notation.

holroy:
Tests indicate that we can currently write colours using the hex-color notation, so the warning should be removed.


holroy:
This was wrong. Please ignore.

PR #210, opened 2019-04-29
Fix recommending of old method

NightYoshi370:
`guild.channels.find('name', 'welcome-messages')` no longer works. Therefore, I updated it to use the newer function method

### DIFF
--- a/guide/additional-info/changes-in-v12.md
+++ b/guide/additional-info/changes-in-v12.md
@@ -775,7 +775,7 @@ Unfortunately, "default" channels don't exist in Discord anymore, and as such, t
 **A:** There are a few ways to tackle this. Using the example of a welcome message system, you can:
 
 1. Set up a database table to store the channel ID in a column when someone uses a `!welcome-channel #channel-name` command, for example. Then inside the `guildMemberAdd` event, use `client.channels.get('id')` and send a message to that channel. This is the most reliable method and gives server staff freedom to rename the channel as they please.
-2. Make a new command that creates a `welcome-messages` channel, use `guild.channels.find('name', 'welcome-messages')`, and send a message to that channel. This method will work fine in most cases, but will break if someone on that server decides to rename the channel. This may also give you unexpected results, due to Discord allowing multiple channels to have the same name.
+2. Make a new command that creates a `welcome-messages` channel, use `guild.channels.find(channel => channel.name == 'welcome-messages')`, and send a message to that channel. This method will work fine in most cases, but will break if someone on that server decides to rename the channel. This may also give you unexpected results, due to Discord allowing multiple channels to have the same name.
 
 <p class="tip">Not sure how to set up a database? Check out [this page](/sequelize/)!</p>
 
--- a/guide/additional-info/changes-in-v12.md
+++ b/guide/additional-info/changes-in-v12.md
@@ -775,7 +775,7 @@ Unfortunately, "default" channels don't exist in Discord anymore, and as such, t
 **A:** There are a few ways to tackle this. Using the example of a welcome message system, you can:
 
 1. Set up a database table to store the channel ID in a column when someone uses a `!welcome-channel #channel-name` command, for example. Then inside the `guildMemberAdd` event, use `client.channels.get('id')` and send a message to that channel. This is the most reliable method and gives server staff freedom to rename the channel as they please.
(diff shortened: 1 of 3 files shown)


KatieIsAKitty:
may also apply [here](https://github.com/discordjs/guide/blob/32bf8dc9da975fac2a54050e51b1041750fec6ea/guide/popular-topics/common-questions.md#how-do-i-add-a-role-to-a-guild-member)

anandre:
It's still correct for v11.x, which the guide is for.  It will be updated accordingly with the master version of the guide, it's on a list of things that need to be updated for v12.

Monbrey:
While "correct" for the stable version, it's still flagged as a deprecated method. I don't think the official guide should be recommending the less-correctway to use the function.

NightYoshi370:
@anandre Plus, the first commit was for the v12 upgrade page. It was removed in that version, so there's no purpose of recommending it

Monbrey:
```suggestion
-if (member.roles.exists('name', '<role name>'))
+if (member.roles.some(role => role.name === '<role name>'))
```

anandre:
Installing Node - probably nice to reference minimum required node version (it's not there at on for stable) - even though it says "install the latest version", not everyone follows that and it could be helpful to just say "get node 10 or higher"

Commands with input aka args -` displayAvatarURL` needs to be updated to a method (currently noted as a warning, just a clean-up there)

Embeds - same thing, has a warning about the `MessageEmbed` constructor change

Permissions - `guild.createRole` as an example needs to be updated, `guild.createChannel` ditto -- it says `permissionsFor` returns a bitfield in the warning but docs says it returns a Permissions object, needs to be fixed`

Reactions - `.find()` example needs to be updated to the function version, raw should be removed in favor of partials

Collectors - for the `awaitMessages` example, `maxMatches` -> `max`

Common questions - `addRole` example, update find and `addRole`; update following example from exists to some; prompt for more input `maxMatches` -> `max`

Misc examples - update the `const { voiceChannel }` declaration `voice.channel` now, no idea if any of the rest of the playing audio changed

That's from the list I made a little bit back about stuff that needed to be updated in the full d.js guide, going from v11 -> v12.  Can amend those too, if you'd like.  I didn't have a priority on it yet.

Monbrey:
At this point I think we need to be clear what we're aiming for. If its on the v12changes page, it needs to be correct for v12, which means deprecations must be updated since they've been removed.

If its on a regular guide page, it should show the non-deprecated version of the command for the stable version, where `addRole` and `createChannel` would still be correct.

anandre:
Well, given it's on the v12-changes branch, it should be for master.  I just didn't check before I commented the first time, oops.  The list in my last comment is all of the stuff I noted for myself that needs to be updated, in addition to the initial PR in all of this.

Danktuary:
Since the PR seems to be intended to update only old `.find()` syntax instances, and it's being PR'd to the v12-changes branch, I think the PR is fine the way it is. Other stuff can be done in other PRs, unless @NightYoshi370 wants to do multiple things in this one.
Just one more instance on guide/popular-topics/reactions.md, line 71.

guide/additional-info/changes-in-v12.md
```diff
@@ -775,7 +775,7 @@ Unfortunately, "default" channels don't exist in Discord anymore, and as such, t
...
 
 1. Set up a database table to store the channel ID in a column when someone uses a `!welcome-channel #channel-name` command, for example. Then inside the `guildMemberAdd` event, use `client.channels.get('id')` and send a message to that channel. This is the most reliable method and gives server staff freedom to rename the channel as they please.
-2. Make a new command that creates a `welcome-messages` channel, use `guild.channels.find('name', 'welcome-messages')`, and send a message to that channel. This method will work fine in most cases, but will break if someone on that server decides to rename the channel. This may also give you unexpected results, due to Discord allowing multiple channels to have the same name.
+2. Make a new command that creates a `welcome-messages` channel, use `guild.channels.find(channel => channel.name == 'welcome-messages')`, and send a message to that channel. This method will work fine in most cases, but will break if someone on that server decides to rename the channel. This may also give you unexpected results, due to Discord allowing multiple channels to have the same name.
```
Danktuary: This should be `===` instead of `==`

NightYoshi370:
@Danktuary I fixed the instance.

PR #218, opened 2019-05-20
Add branch version selector dropdown to user settings

### LINKED ISSUE #121: Multiple versions

Danktuary:
This PR adds a `<branch>` component to the guide so that we can use it to conditionally render content on pages depending on the version the user selects from the dropdown menu shown above. This will allow us to easily account for both v11 and v12 syntax on the same page the rest of the relevant content is.

Examples:

```vue
Here's some conditional content that'll display depending on your choice from the site settings, in the dropdown menu from the cog icon above.

<branch version="11.x">v11.x content</branch>
<branch version="11.4">v11.4 content</branch>
<branch version="11.5">v11.5 (stable) content</branch>
<branch version="12.x">v12.x (master) content</branch>
```

And of course, you can see a live demo of this here: https://yuu-version-selector.netlify.com/version-selector.html

Finally closes #121.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -240,3 +240,41 @@ The `author` and `avatar` attributes must be strings, and the `bot` attribute mu
 Do note the `<div is="discord-messages">` syntax instead of `<discord-messages>`. This is due to how VuePress renders markdown and HTML inside markdown files and doesn't recognize `<discord-messages>` as an HTML element, therefore rendering anything indented inside as a regular codeblock.
 
 You can read more about how to use these components by checking out [the package's GitHub repo](https://github.com/Danktuary/vue-discord-message).
+
+### Branch-specific content
+
+On some pages, you'll want to display content that applies only to the stable branch and other content that applies to a different branch. You can use the `<branch>` component inside any .md file like so:
+
+```md
+You can use
+<branch version="11.x" inline>`message.channel.fetchMessages()`</branch>
+<branch version="12.x" inline>`message.channel.messages.fetch()`</branch>
+to fetch all messages in a channel
+```
+
+If you're on the `11.x` branch, you'd see "You can use `message.channel.fetchMessages()` to fetch all messages in a channel. Use the `inline` attribute to make content display inline with the content around it. Otherwise, it'll be displayed on its own line.
+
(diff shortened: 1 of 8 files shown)


almostSouji:
- Unsure if the placement of the selection isn't too hidden in the gear menu, i doubt most people even notice it exists, might be better placed in the main nav
- I played around a bit and noticed that i can't get code blocks to work inside the component, however that might also be me doing something wrong there: https://paste.nomsy.net/ifosewemac.html

The latter being a major use case of version selection throughout the guide

Danktuary:
> * Unsure if the placement of the selection isn't too hidden in the gear menu, i doubt most people even notice it exists, might be better placed in the main nav
> * Is there a way to introduce a inline version of that tag as well? Maybe it's beneficial to default it to inline and make a block version instead, i see use cases where inline versioning could be more streamlined
> * I played around a bit and noticed that i can't get code blocks to work inside the component, however that might also be me doing something wrong there: https://paste.nomsy.net/ifosewemac.html
> 
> The latter being a major use case of version selection throughout the guide

1) I tried placing the dropdown menu to the left of the cog icon but it causes a problem on mobile:



I tried making it mobile friendly by duplicating the component inside the menu and hiding it on the navbar if you're on a mobile device, but it causes some reactivity issues and could cause your branch selection to be out of sync. For the time being, I think I'll leave it where it is and find an alternate solution later.

2) I added an `inline` prop to the Branch component so that you can use it like so now:

```md
You're on version
<branch version="11.x" inline>11.x</branch>
<branch version="12.x" inline>12.x</branch>
of discord.js
```

3) I've been messing around with this and noticed that markdown and Vue components are a bit tricky to handle.

For example, this doesn't work:

```md
<branch version="12.x">
\`\`\`js
message.channel.messages.fetch()
\`\`\`

alsjkfdnaskdfjn
</branch>
```

```md
<branch version="12.x">

\`\`\`js
message.channel.messages.fetch()
\`\`\`

alsjkfdnaskdfjn

</branch>
```

So I guess the workaround there would be to make sure to use extra spacing where you need it. It's not the most appealing when you're looking at the source code, but the page result is what matters the most.

---

I forgot to add this before but I've added a section in the contributing.md file on how to use this component.

PR #223, opened 2019-06-05
Move Promise.all approach to its own subheading

almostSouji:
We relatively regularly had and still have people on the support server interpreting this approach as if it should react in order, although it clearly states that it does not.

I can see however, where the confusion is coming from. The guide part is called "reacting in order" and deals with that topic, so the removed snippet really does not belong here.

If you are in favor of keeping it, maybe as misc. code sample please let me know, although i'm not too sure in which context we'd fit it in there.

**Edit:** Keeping it under its own subheading instead of removing, as it still is useful information on how to deal with parallel promises at once.


appellation:
Could it go as a separate section after "Reacting in order"?

PR #237, opened 2019-06-26
Update config.json

Designer-rohit:


Harsh-2909:
That config.json file is supposed to be that way. You are literally showing your token to others which can be very dangerous in security reason.

almostSouji:
No idea what kind of secret that's supposed to be, but it's not a discord token.
However unless there's any explanation as to why this should be integrated into the guide i'm going to close this as invalid.

PR #249, opened 2019-07-21
Update Windows install method for FFmpeg

gregariousalex:
The old installation method is now deprecated, so an update seems necessary.

### DIFF
--- a/guide/popular-topics/miscellaneous-examples.md
+++ b/guide/popular-topics/miscellaneous-examples.md
@@ -30,9 +30,14 @@ sudo apt-get install ffmpeg
 
 On Windows:
 
+::: tip
+You can get choco at [chocolatey.org](https://chocolatey.org/)
+:::
+
 ```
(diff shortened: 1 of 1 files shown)


amishshah:
It is also suitable to just run `npm install ffmpeg-static`

almostSouji:
I don't think we should propose choco as the prime solution here.
The package `ffmpeg-static` seems to do the job, as well as installing from the webpage, as crawl explains here: <https://www.youtube.com/watch?v=SW-iKrT_nJs>
If anything - as alternative - we should provide the side note that it can also be installed with choco, homebrew or the respective linux distro package manager.

Since amish has provided part of this information a long time ago but no activity happened i'll close this. Feel free to comment if you want to make the changes yourself and we'll re-open the PR.

At the current state, if someone wants to provide an alternative approach please feel free to do so, the reasoning above can serve as guideline.

PR #254, opened 2019-07-27
Update parsing-mention-arguments.md

TomSputz:
Add test for `&` in user mentions resolver, which indicates a role mention

The `regex` section directly after is comparing the simplicity of using regex as an alternative. If the algorithms arent actually equivalent, the point isnt conveyed as clearly

### DIFF
--- a/guide/miscellaneous/parsing-mention-arguments.md
+++ b/guide/miscellaneous/parsing-mention-arguments.md
@@ -84,7 +84,7 @@ Putting it into a function will make it easily reusable. We will use the name `g
 function getUserFromMention(mention) {
 	if (!mention) return;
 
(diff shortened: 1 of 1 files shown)


vzwGrey:
> Add test for & in user mentions resolver, which indicates a role mention

The function is specifically for parsing *user* mentions, as stated by its name. Adding checks for role mentions as well would convolute the function and is not necessary to explain the general functionality of the code.

> The regex section directly after is comparing the simplicity of using regex as an alternative. If the algorithms arent actually equivalent, the point isnt conveyed as clearly

I'm not sure what you mean? The regex section also only covers user mentions and does not work for role mentions.

TomSputz:
Ah, you're totally right. I'm tempted to say `fetch` should be being used, but it's not that important to the topic

PR #258, opened 2019-08-17
Fix a little typo that break a link

DreamXZE:
Corrected a little typo that break a redirection link. I'm not sure if that will fix it, but you can see where it is


almostSouji:
This issue happens due to the old callout structure component being used, the proposed fix does not actually fix the link, the referenced commit does so while also updating the other callouts

PR #264, opened 2019-10-04
Slight rework of the grammar/text in the sequelize guide

Darqam:
Mostly making the text consistent by removing instances of `we` and `us` in favor of a neutral or `you` speech.

Remove contractions for a slightly more formal speech; though that may be an arguable change.

I probably am missing some stuff so feel free to mention it.

There is also the sentence `SQlite will be used as the first storage engine, other database will be expanded upon later.`, which as far as I can tell is incorrect since the guide does not really expand on further DBs, so I'm not sure if this should stay or not.

### DIFF
--- a/guide/sequelize/README.md
+++ b/guide/sequelize/README.md
@@ -1,6 +1,6 @@
 # Storing data with Sequelize
 
-Sequelize is an object-relational-mapper, which means you can write a query using objects and have it run on almost any other database system that Sequelize supports.
+Sequelize is an object-relational-mapper, which means you can write a query using objects and have it runs on almost any other database system that Sequelize supports.
 
 ### Why use an ORM?
 
@@ -8,8 +8,8 @@ The main benefit of using an ORM like Sequelize is that it allows you to write c
 
 ## A simple tag system
 
-For this tutorial, we'll create a simple tag system. The tag system will allow you to add a tag, output a tag, edit the tag, show tag info, list tags, and delete a tag.   
-To begin, you should install Sequelize into your discord.js project. We'll use sqlite as our first storage engine, but we'll move into other databases later. Note that you'll need node 7.6 or above to utilize the `async/await` operators.
+For this tutorial, you will create a simple tag system. The tag system will allow you to add a tag, output a tag, edit the tag, show tag info, list tags, and delete a tag.   
+To begin, you should install Sequelize into your discord.js project. SQlite will be used as the first storage engine, other database will be expanded upon later. Note that you will need node 7.6 or above to utilize the `async/await` operators.
 
 ### Installing and using Sequelize
 
@@ -25,7 +25,7 @@ $ npm install --save sqlite3
 Make sure you use version 5 or later of Sequelize! Version 4 as used in this guide will pose a security threat. You can read more about this issue On the [Sequelize issue tracker](https://github.com/sequelize/sequelize/issues/7310).
 :::
 
-After you have installed discord.js and Sequelize, you can start with the following skeleton code. The comment labels will tell you where we'll insert our code.
+After having installed discord.js and Sequelize, you can start with the following skeleton code. The comment labels will tell you where to insert the later code blocks.
 
 <!-- eslint-disable require-await -->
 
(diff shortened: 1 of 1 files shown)


guide/sequelize/README.md
```diff
@@ -123,30 +123,30 @@ const Tags = sequelize.define('tags', {
...
 Tags.sync();
 ```
 
-The table doesn't actually get created until you `sync` it. The schema that we defined from before was simply creating the model that lets Sequelize know what our data should look like. For testing, you can use `Tags.sync({ force: true })` to recreate the table every time on startup. This way you can get blank slate each time.
+The table does not actually get created until you `sync` it. The schema that was defined from before was simply creating the model that lets Sequelize know what the data should look like. For testing, you can use `Tags.sync({ force: true })` to recreate the table every time on startup. This way you can get blank slate each time.
 
 ### [delta] Adding a tag
 
-We can finally get our first command. We'll start off with adding a tag.
+You can finally get your first command. Start off with adding a tag.
```
almostSouji: We is fine here, as we - the creators of the guide - decide to start with this command
Getting a command sounds rather confusing, giving the sentence a bit more flesh can help here in my opinion:
```suggestion
After all this preparation you can now write your first command. We will start off with the ability to add a tag.
```

guide/sequelize/README.md
```diff
@@ -172,17 +172,17 @@ catch (e) {
 }
 ```
 
-`Tags.create()` uses our models that we created previously. The `.create()` method inserts some data into the model. We're going to be inserting a tag name, description, and the author name into the database.  
-`catch (e)` This section is absolutely necessary for our insert. We offload checking for duplicates to the database, so that it will tell us if we create a tag that already exists. The alternative is to query the database before adding data, and checking if we get a result. If we don't, only then do we add the data. But this requires two queries instead of one, so this method is less work.   
-`if (e.name === "SequelizeUniqueConstraintError")` Although this was mostly for doing less work, it's always good to handle your errors, especially if you know what types of errors you will receive. This error comes up if your unique constraint is violated, i.e. someone inserted duplicate values.
+`Tags.create()` uses the models that was created previously. The `.create()` method inserts some data into the model. you are going to be inserting a tag name, description, and the author name into the database.
```
almostSouji: Again, should not be passively be created, the user did it
You capital after `.`
"going to insert" is much cleaner to read.
```suggestion
`Tags.create()` uses the models that you created previously. The `.create()` method inserts some data into the model. You are going to insert a tag name, description, and the author name into the database.  
```

guide/sequelize/README.md
```diff
@@ -172,17 +172,17 @@ catch (e) {
...
+`if (e.name === "SequelizeUniqueConstraintError")` Although this was mostly for doing less work, it is always good to handle your errors, especially if you know what types of errors you will receive. This error comes up if your unique constraint is violated, i.e. someone inserted duplicate values.
 
 ::: warning
 Do not use catch for inserting new data. Only use it for gracefully handling things that go wrong in your code, or logging errors.
 :::
 
 ### [epsilon] Fetching a tag
 
-Next we will fetch the tag we just inserted.
+Next you will fetch the tag that was just inserted.
```
almostSouji: same as above, we will explain how to do this
```suggestion
Next we will explain how to fetch the tag that was just inserted.
```

guide/sequelize/README.md
```diff
@@ -236,11 +236,11 @@ if (tag) {
...
+This section is very similar to our previous command, except you will be showing the tag metadata. `tag` contains our tag object. Notice two things here: firstly, it is possible to access our object properties without the `.get()` function. This is because the object is an instance of a Tag, which you can treat as an object, and not just a row of data. Second, you can access a property that was not defined explicitly, `createdAt`. This is because Sequelize automatically adds that column to all tables. This feature can be disabled by passing another object into the model with `{ createdAt: false }`, but in this case, it was useful to have.
 
 ### [lambda] Listing all tags
 
-We'll use the next command to fetch a list of all the tags we've created so far.
+You will use the next command to fetch a list of all the tags that were created so far.
```
almostSouji: Again, more flesh for easier reading
```suggestion
The next command will enable you to fetch a list of all the tags that were created so far.
```

Darqam:
I believe all the relevant changes have been made.

PR #265, opened 2019-10-05
Adding two examples

Darqam:
Adding an example for fetching (counting) all online members in a guild as well as a way to monitor role addition and removal in the guildMemberAdd event.

Prompted by the ToDo section(s) by Souji (so blame him if it's unwanted).

### DIFF
--- a/guide/popular-topics/common-questions.md
+++ b/guide/popular-topics/common-questions.md
@@ -202,3 +202,15 @@ Be sure to [limit this to your own ID](/popular-topics/common-questions.md#how-d
 ### What is the difference between a User and a GuildMember?
 
 A lot of users get confused as to what the difference between Users and GuildMembers is. The simple answer is that a User represents a global Discord user and a GuildMember represents a Discord user on a specific server. That means only GuildMembers can have permissions, roles, and nicknames, for example, because all of these things are server-bound information that could be different on each server that user is in.
+
+### How do I find all online members?
+
+Assuming the process is to be done for the guild the message is sent in.
+```js
+// First we fetchMembers to be 100% sure everyone is in the member cache
+<message>.guild.fetchMembers().then(fetchedGuild => {
+	const totalOnline = fetchedGuild.members.filter(member => member.presence.status === 'online');
+	// We now have an array with all online member objects in the totalOnline variable
+	<message>.channel.send(`There are currently ${totalOnline.size} members online in this guild!`);
+});
+```
--- a/guide/popular-topics/common-questions.md
+++ b/guide/popular-topics/common-questions.md
@@ -226,10 +226,10 @@ Assuming the process is to be done for the guild the message is sent in.
 // This code should be placed outside of any other listener callbacks to prevent listener nesting
 
 <client>.on('guildMemberUpdate', (oldMember, newMember) => {
-	const removedRoles = oldMember.roles.filter(role => !newMember.roles.some(roleTwo => roleTwo.name === role.name));
+	const removedRoles = oldMember.roles.filter(role => !newMember.roles.has(role.id));
 	if(removedRoles.size > 0) console.log(`The roles ${removedRoles.map(r => r.name)} were added to ${oldMember.displayName}.`);
 
-	const addedRoles = newMember.roles.filter(role => !oldMember.roles.some(roleTwo => roleTwo.name === role.name));
+	const addedRoles = newMember.roles.filter(role => !oldMember.roles.has(role.id));
 	if(addedRoles.size > 0) console.log(`The roles ${addedRoles.map(r => r.name)} were removed from ${oldMember.displayName}.`);
 });
 ```
--- a/guide/popular-topics/common-questions.md
+++ b/guide/popular-topics/common-questions.md
@@ -59,28 +59,6 @@ if (<message>.author.id === '<id>') {
 }
 ```
 
-### How do I check which role was added/removed, and for which member?
-
-<!-- eslint-skip -->
-
-```js
-// We start by declaring a guildMemberUpdate listener
-// This should ***not*** be inside the message listener (or any other listener)
-
-<client>.on('guildMemberUpdate', (oldMember, newMember) => {
-	if(oldMember.roles.size === newMember.roles.size) return;
-
-	if(oldMember.roles.size > newMember.roles.size) {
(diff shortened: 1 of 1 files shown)


GamesProSeif:
I agree with you in the former, code style needs some changes.
Though I feel these changes fall correctly under the common-questions category, since they're part of FAQs.
Souji did state the category (FAQ) in the project [Page ideas](https://github.com/discordjs/guide/projects/1).
- [FAQ: how to find online members](https://github.com/discordjs/guide/projects/1#card-21323889)
- [How to find a role that was added/removed on guildMemberUpdate (FAQ from discord)](https://github.com/discordjs/guide/projects/1#card-20773449)

guide/popular-topics/common-questions.md
```diff
@@ -59,6 +59,28 @@ if (<message>.author.id === '<id>') {
...
+### How do I check which role was added/removed, and for which member?
+
+<!-- eslint-skip -->
+
+```js
+// We start by declaring a guildMemberUpdate listener
+// This should ***not*** be inside the message listener (or any other listener)
+
+<client>.on('guildMemberUpdate', (oldMember, newMember) => {
+	if(oldMember.roles.size === newMember.roles.size) return;
```
almostSouji: This suggests that roles can only be changed if the size of the role collection changes, which is wrong.
Roles can not only be added or removed, and not only one by one either (relevant further down for the usage of `find`). 

The approach I tend to suggest to people with this request is the following:
is in old but not in new => removed
is in new but not in old =>added
(is in both => unchanged) which i'd not include here

guide/popular-topics/common-questions.md
```diff
@@ -202,3 +224,17 @@ Be sure to [limit this to your own ID](/popular-topics/common-questions.md#how-d
...
 
 A lot of users get confused as to what the difference between Users and GuildMembers is. The simple answer is that a User represents a global Discord user and a GuildMember represents a Discord user on a specific server. That means only GuildMembers can have permissions, roles, and nicknames, for example, because all of these things are server-bound information that could be different on each server that user is in.
+
+### How do I find all online members?
+
+Assuming the process is to be done for the guild the message is sent in.
+<!-- eslint-skip -->
+
+```js
+// First we fetchMembers to be 100% sure everyone is in the member cache
```
almostSouji: I don't think this reads very well, I'd suggest something along the lines of:
```suggestion
// First we use fetchMembers to make sure all members are cached
```

guide/popular-topics/common-questions.md
```diff
@@ -59,6 +59,28 @@ if (<message>.author.id === '<id>') {
 }
 ```
 
+### How do I check which role was added/removed, and for which member?
```
almostSouji: > Furthermore, the question is more of a miscellaneous question than an administrative one.

I agree, this should not be placed in the Administrative section but misc (subheading of common questions, not misc. examples)

guide/popular-topics/common-questions.md
```diff
@@ -59,6 +59,28 @@ if (<message>.author.id === '<id>') {
...
+
+<!-- eslint-skip -->
+
+```js
+// We start by declaring a guildMemberUpdate listener
+// This should ***not*** be inside the message listener (or any other listener)
```
almostSouji: I don't think this reads very well, I'd suggest something along the lines of:
```suggestion
// This code should be placed outside of any other listener callbacks to prevent listener nesting
```

Darqam:
I believe this addresses the concerns.

With respect to the comments about code style, I am not quite sure what the two of you mean. `some` and `filter` are both methods inherited/modified from arrays, so there's nothing that should be black magic to users.

The code could be far more explicit with `for` loops or something, but that would seem kind of pointless to me given the provided tools. I could very well be missing a way to make this simpler or cleaner, if so feel free to mention it, however as it stands I don't see it.

I'll also note that now the console.log returns assuming plurality of roles changed (which I forgot because setRoles is a thing), and will stay as such for single role changes. One could make a size check and adjust the sentence accordingly, but this seems like more noise than is necessary for such an example.

guide/popular-topics/common-questions.md
```diff
@@ -202,3 +202,34 @@ Be sure to [limit this to your own ID](/popular-topics/common-questions.md#how-d
...
+### How do I check which role was added/removed, and for which member?
+
+<!-- eslint-skip -->
+
+```js
+// We start by declaring a guildMemberUpdate listener
+// This code should be placed outside of any other listener callbacks to prevent listener nesting
+
+<client>.on('guildMemberUpdate', (oldMember, newMember) => {
+	const removedRoles = oldMember.roles.filter(role => !newMember.roles.some(roleTwo => roleTwo.name === role.name));
```
almostSouji: Not a fan of filtering by name here, id is way more robust against edge cases
```suggestion
	const removedRoles = oldMember.roles.filter(role => !newMember.roles.has(role.id));
```

guide/popular-topics/common-questions.md
```diff
@@ -202,3 +202,34 @@ Be sure to [limit this to your own ID](/popular-topics/common-questions.md#how-d
...
+	if(removedRoles.size > 0) console.log(`The roles ${removedRoles.map(r => r.name)} were added to ${oldMember.displayName}.`);
+
+	const addedRoles = newMember.roles.filter(role => !oldMember.roles.some(roleTwo => roleTwo.name === role.name));
```
almostSouji: same as above
```suggestion
	const addedRoles = newMember.roles.filter(role => !oldMember.roles.has(role.id));
```

Darqam:
Right, id makes far more sense and is also much simpler to read.

guide/popular-topics/common-questions.md
```diff
@@ -202,3 +202,36 @@ Be sure to [limit this to your own ID](/popular-topics/common-questions.md#how-d
...
+### How do I find all online members?
+
+Assuming the process is to be done for the guild the message is sent in.
+<!-- eslint-skip -->
+
+```js
+// First we use fetchMembers to make sure all members are cached
+<message>.guild.fetchMembers().then(fetchedGuild => {
+	const totalOnline = fetchedGuild.members.filter(member => member.presence.status === 'online');
+	// We now have an array with all online member objects in the totalOnline variable
```
SpaceEEC:
```suggestion
	// We now have a collection with all online member objects in the totalOnline variable
```
[`Collection#filter`](https://discord.js.org/#/docs/main/stable/class/Collection?scrollTo=filter) returns a Collection.

PR #267, opened 2019-10-06
Fix client.once (change to client.on)

UndecidedSource:


Drahcirius:
Why?

Danktuary:
`.once()` is being used on purpose here; `.on()` isn't necessary unless you need the code to run every single time the event fires. In this case, knowing the bot is ready just once is enough.

PR #271, opened 2019-10-08
Add guide page for audit logs

Darqam:
Ok so here's an addition for some examples of audit logs (3 main ones I could think of) usage.

I don't believe the branch is ready to merge as is and would like to have some extra comments on what should be added (or removed) and reworked/reworded.

There is a decent amount of repetition since I assumed that one example should be understood relatively independently from the others. Would perhaps a bit more digging into the functionality/options of fetchAuditLogs() be worthwhile? 

I realize it's a topic not many people care about; but the issues come up often enough in the discord server I think it's worth adding *some* info on the guide about it.

p.s did some github screwup with commits and previous PR stuff got mixed in, so early commit history is irrelevent


guide/popular-topics/audit-logs.md
```diff
@@ -146,7 +146,7 @@ client.login('your-token-goes-here');
 As was the case in the previous examples, we can see what happened, to whom it happened, but not who executed the action. Enter once again audit logs where we will limit ourselves to 1 entry, and looking at the `MEMBER_BAN_ADD` type. Our `guildBanAdd` listener then becomes.
 
 ```js
-client.on('guildBaneAdd', async message => {
+client.on('guildBanAdd', async message => {
```
Jiralite: Hey uhm... doesn't the `guildBanAdd` event take `Guild` as the first parameter?

tipakA: Looks like similiar thing also happens on line 97.

Darqam: resolved

guide/popular-topics/audit-logs.md
```diff
@@ -0,0 +1,167 @@
+# Working with Audit Logs
+
+## Some quick background
+Audit logs are a great moderation tool offered by discord to know what happened in a server, and usually by whom. There are quite a few cases where audit logs may be used, this guide will limit itself to the most common use cases. For more information on audit logs, feel free to consult the [relevant discord api page](https://discordapp.com/developers/docs/resources/audit-log).
```
almostSouji: I'd suggest rephrasing this section and put emphasis on the fact it is the only way to find out who the executor of moderative actions is and that events do not provide that information, hence making fetches necessary

guide/popular-topics/audit-logs.md
```diff
@@ -0,0 +1,167 @@
...
+
+In the following examples, a very simple case for some auditLog types will be explored. Some basic error handling is performed but these code segments are by no means fool proof and are meant to teach you how fetching audit logs work. For a rigorous system you will most likely need to expand on the examples based on your own goals.
+
+## Who Deleted a Message?
+Let us dive right into it with probably the most common use of audit logs; understanding who deleted any given message in a discord server.
+
+::: warning
+At the moment of writing this, discord does not emit an audit log if the person who deleted the message is a bot, or is the author of the message itself.
```
almostSouji: bot bulk deletes do now cause an audit log entry, iirc that's new since the latest audit log update; single message deletions by bot do not.

guide/popular-topics/audit-logs.md
```diff
@@ -0,0 +1,167 @@
...
+});
+
+client.on('messageDelete', message => {
+	console.log(`A message by ${message.author.tag} was deleted, but we don't know by who yet.`);
+});
+
+client.login('your-token-goes-here');
+```
+
+So far nothing should seem new or complicated, we get the message deleted event, and log that a message was removed from a channel. We could make use of more information from the message object, but that is left as an exercise for the student.
```
almostSouji:
```suggestion
So far nothing should seem new or complicated, we get the message deleted event, and log that a message was removed from a channel. We could make use of more information from the message object, but that is left as an exercise for the reader.
```

guide/popular-topics/audit-logs.md
```diff
@@ -0,0 +1,167 @@
...
+});
+
+client.on('guildMemberRemove', member => {
+	console.log(`${member.user.tag} left the guild... but was it of their own free will?`);
+});
+
+client.login('your-token-goes-here');
+```
+
+We will again fetchAuditLogs while limiting ourselves to 1 entry, and looking at the `MEMBER_KICK` type.
```
almostSouji: "fetch audit logs" or "use `fetchAuditLogs` to be consistent

guide/popular-topics/audit-logs.md
```diff
@@ -0,0 +1,167 @@
...
+		console.log(`${member.user.tag} left the guild; kicked by ${executor.tag}?`);
+	} else {
+		console.log(`${member.user.tag} left the guild, audit log fetch was inconclusive.`);
+	}
+});
+```
+
+## Who banned a user?
+
+The logic for this will be very similar to the above kick example, with the exception that this time we have an event specifically for guild bans, that is `guildBanAdd`. Starting with again a skeleton code, we have the following.
```
almostSouji: nit, "Again starting with" or "Starting with... again"

guide/popular-topics/audit-logs.md
Darqam: :(

Darqam:
WRT 'do you think a code sample would make any sense to include?'

I don't personally think this is warranted or a good idea. Each section already has it's own "resulting code", and each thing is rather independent from the other.

PR #276, opened 2019-10-18
Make most links to docs branch relative; small fixes around branch specific content

tipakA:
After adding branch selector, most links to docs were pointing to `stable`. While there is nothing wrong with that, there is not really much difference between most of linked classes / properties / etc. anyway, there still are small changes though, like for example `YELLOW` as `ColorResolvable` on master but not on stable, or deprecated permission flags that are removed on master. Additionally, at least for me, it's quite convienient to be pointed to functional documentation page, from which i can browse sidebar and jump to any other class, without need to change branch twice, first in guide, and then in docs.
~~However, not all links could be "branched". There is no documentation for `Collection` on master, so those were left on stable.~~
"Fixed" in 4b1dadb.

- Made all links to docs branch relative (`main/11.5.1` for `11.x` and `main/master` for `12.x`)
- Replaced all `stable` in links to `11.5.1`, in case of eventual v12 release, guide will not need instant update to account for different stable.
- Replaced all `v11.x` and `v12.x` to `11.x` and `12.x` for consistency with `CONTRIBUTING.md`, despite the fact that they worked.
- Fixed Embed example that was using `Discord.Attachment` on master
- Fixed Embed page linking to `Discord.MessageEmbed` instead of `Discord.RichEmbed` for embed utility class on stable
- Fixed one incorrect occurence of `main/12.0.0` link
- Fixed three typos in Partials, and a typo for hotlink


guide/command-handling/README.md
```diff
@@ -81,7 +81,7 @@ const client = new Discord.Client();
 :::
 
 ::: tip
-If you aren't exactly sure what Collections are, they're a class that extend JS's native Map class and include more extensive, useful functionality. You can read about Maps [here](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map), and see all the available Collection methods [here](https://discord.js.org/#/docs/main/stable/class/Collection).
+If you aren't exactly sure what Collections are, they're a class that extend JS's native Map class and include more extensive, useful functionality. You can read about Maps [here](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map), and see all the available Collection methods [here](https://discord.js.org/#/docs/main/11.5.1/class/Collection).
```
almostSouji: needs version specific docs link

guide/command-handling/adding-features.md
```diff
@@ -383,7 +383,7 @@ At the end of it all, you should be getting this as a result:
 No more manually editing your help command! If you aren't completely satisfied with how it looks, you can always adjust it to your liking later.
 
 ::: tip
-If you want to add categories or other information to your commands you can simply add properties reflecting it to your `module.exports`. If you only want to show a subset of commands remember that `commands` is a Collection you can [filter](https://discord.js.org/#/docs/main/stable/class/Collection?scrollTo=filter) to fit your specific needs!
+If you want to add categories or other information to your commands you can simply add properties reflecting it to your `module.exports`. If you only want to show a subset of commands remember that `commands` is a Collection you can [filter](https://discord.js.org/#/docs/main/11.5.1/class/Collection?scrollTo=filter) to fit your specific needs!
```
almostSouji: needs version specific docs link

guide/command-handling/README.md
tipakA: I mentioned that as no documentation for Collections on master, as they have separate category in docs now.
Thing is, the collection category is not really "working", as mentioned in https://github.com/discordjs/website/issues/41. I had to literally nuke whole local data for the website to be able to even see that category.

almostSouji: I don't think this is going to be addressed in a timely manner (if at all), but the Link should still be correct for v12. If you think it's an issue i could see linking to <https://discord.js.org/#/docs/collection/master/class/Collection> instead, as the separate package approach to all things Collection is here to stay.

PR #278, opened 2019-10-24
Enhance `configuration-files` example to utilize the configurized prefix.

Fedelaus:


code-samples/creating-your-bot/configuration-files/index.js
```diff
@@ -7,7 +7,11 @@ client.once('ready', () => {
 });
 
 client.on('message', message => {
-	if (message.content === '!ping') {
+	if (!message.content.startsWith(config.prefix) {
+		return;
```
tipakA: Personally, i wouldn't use whole block for simple `return;`, neither this format is used further in the guide.
For example: 
https://github.com/discordjs/guide/blob/adef8eae8421f149a4cff148f72d39f3abf3170b/code-samples/creating-your-bot/commands-with-user-input/11/index.js#L9-L10

Fedelaus: Many apologies, I actually came to the guide after helping someone out with this. I am happy to move to the single-line return, and will do so.

code-samples/creating-your-bot/configuration-files/index.js
```diff
@@ -7,7 +7,9 @@ client.once('ready', () => {
 });
 
 client.on('message', message => {
-	if (message.content === '!ping') {
+	if (!message.content.startsWith(config.prefix)) return;
+
```
almostSouji: Please remove the trailing space and for the future I'd suggest running the lint script to prevent things like these. `yarn run lint` | `npm run lint`
```suggestion

```

Fedelaus: Thanks @almostSouji, I had been doing this entirely from the web-ui.

Fedelaus:
Thanks for baring with me guys, appreciate the pointers.

code-samples/creating-your-bot/configuration-files/index.js
```diff
@@ -7,7 +7,8 @@ client.once('ready', () => {
 });
 
 client.on('message', message => {
-	if (message.content === '!ping') {
+	if (!message.content.startsWith(config.prefix)) return;
```
almostSouji: Looked at the guide flow again, these resulting code snippets should reflect what's taught in the guide and this is at the very beginning, where we just introduced directly comparing the message content, rather than dissecting it and returning if it doesn't start with the prefix. Line 10 is not mentioned in the respective guide https://discordjs.guide/creating-your-bot/#replying-to-messages yet (previous state of the guide before entering this section)

Fedelaus: @almostSouji that section of code is also referenced and linked here

https://discordjs.guide/creating-your-bot/configuration-files.html#implementing-your-config-file

and this is what I was working from. It feels weird for the guide to mention putting things within configuration and then not use said configuration.

almostSouji: I'm referring to line 2. not how 3 is changed to use the configured prefix, that's absolutely valid.

Fedelaus: Yeah - you're right. I think I came at that from the approach of "I would do this within my application" but it's out of scope for the tutorials that consume it.

To confirm, drop line 2?

almostSouji:
```suggestion
```

chroventer:
Alternatively, you can destruct only the prefix from the config file.
```js
const { prefix } = require('path/to/config');
```

almostSouji:
> 
> 
> Alternatively, you can destruct only the prefix from the config file.
> 
> ```js
> const { prefix } = require('path/to/config');
> ```

We explain destructuring later in the guide, that should not be introduced here

chroventer:
> We explain destructuring later in the guide, that should not be introduced here, neither for the prefix nor the Client constructor

We can perhaps include a link to the topic - "How to destruct in JavaScript"

almostSouji:
With the current state of the guide this would be inconsistent. If you want to propose the guide to change its approach to prefer destructuring from the beginning please do so in a separate PR or issue as it will require changes in a much bigger scope than the issue this PR attempts to address.

chroventer:
> @almostSouji how come this got unmerged?

Read more on [this PR](https://github.com/discordjs/guide/pull/279) as to why the commit/changes got reverted back.

PR #281, opened 2019-10-28
Add guide on reloading commands

tipakA:
Adds a guide on creating a command to reload commands.

### DIFF
--- a/guide/command-handling/adding-features.md
+++ b/guide/command-handling/adding-features.md
@@ -413,7 +413,7 @@ const command = message.client.commands.get(commandName)
 if (!command) return message.channel.send(`There is no command with name or alias \`${commandName}\`, ${message.author}!`);
 ```
 
-`message.client`? Yes! Every library specific object has `client` as a property. That means that you don't have to pass it as a parameter to commands to access for example `client.guilds` or `client.commands` that you will need now.
+`message.client`? Yes! Every library specific object has `client` as a property. That means that you don't have to pass it as a parameter to commands to access for example `client.guilds` or `client.commands` which you will need now.
 
 Now, in theory, all there is to do, is to delete the previous command from `client.commands` and require the file again. In practice though, you cannot do this that easy, as `require()` caches the file. If you would require it again, you would simply load previously cached file without any of your changes. In order to remove the file from the cache, you need to add the following line to your command's code:
 
@@ -433,9 +433,9 @@ try {
 }
 ```
 
-Code above uses `try/catch` to load the command file and add it to `client.commands`, and in case of error - logs whole error to console and notifies about the fact.
+Snippet above uses `try/catch` to load the command file and add it to `client.commands`, and in case of error - log whole error to console and notify about the fact sending only short `error.message`.
 
-Last thing you need to add, is to send message to confirm succesful reload:
+Last thing you might want to add is sending a message on succesful reload:
 
 ```js
 message.channel.send(`Command \`${commandName}\` was reloaded!`);
--- a/guide/command-handling/adding-features.md
+++ b/guide/command-handling/adding-features.md
@@ -434,7 +434,7 @@ try {
 }
 ```
 
-Snippet above uses `try/catch` to load the command file and add it to `client.commands`, and in case of error - log whole error to console and notify about the fact sending only short `error.message`. Note the fact that we never actually delete the command from commands collection, and instead just overwrite it. This prevents from deleting a command, and ending up with no such command in collection upon failed `require()`, as each use of the reload command checks that collection again.
+The snippet above uses a `try/catch` block to load the command file and add it to `client.commands`. In case of an error it will log the full error to console and notify the user about it with the error's message component `error.message`. Note the you never actually delete the command from the commands collection, and instead just overwrite it. This prevents you from deleting a command, and ending up with no command at all after a failed `require()` call, as each use of the reload command checks that collection again.
 
 The last thing you might want to add is sending a message if the reload was successful:
 
--- a/guide/command-handling/adding-features.md
+++ b/guide/command-handling/adding-features.md
@@ -414,7 +414,9 @@ const command = message.client.commands.get(commandName)
 if (!command) return message.channel.send(`There is no command with name or alias \`${commandName}\`, ${message.author}!`);
 ```
 
-`message.client`? Yes! Every library specific object has `client` as a property. That means that you don't have to pass it as a parameter to commands to access for example `client.guilds` or `client.commands` which you will need now.
+:::tip
+Every library specific object has `client` as a property. That means that you don't have to pass it as a parameter to commands to access for example `client.guilds` or `client.commands`, but can access it directly from `message` object, as in the snippet above.
+:::
 
 Now, in theory, all there is to do, is to delete the previous command from `client.commands` and require the file again. In practice though, you cannot do this that easily as `require()` caches the file. If you were to require it again, you would simply load the previously cached file without any of your changes. In order to remove the file from the cache, you need to add the following line to your code:
 
--- a/guide/command-handling/adding-features.md
+++ b/guide/command-handling/adding-features.md
@@ -433,7 +433,7 @@ try {
 }
 ```
 
-Snippet above uses `try/catch` to load the command file and add it to `client.commands`, and in case of error - log whole error to console and notify about the fact sending only short `error.message`.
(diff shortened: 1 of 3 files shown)


tipakA:
I think that's all i can add / improve by myself, so might as well mark as ready to review

guide/command-handling/adding-features.md
```diff
@@ -386,6 +386,62 @@ No more manually editing your help command! If you aren't completely satisfied w
...
+module.exports = {
+	name: 'reload',
+	description: 'Reloads a command',
+	execute(message, args) {
+		// ...
+	},
+};
+```
+
+In this command, you will be using command's name or alias as an only argument. First off, you need to check if command you want to reload exists. This can be done in similiar fashion as getting a command in your main file.
```
almostSouji:
```suggestion
In this command, you will be using a command name or alias as the only argument. First off, you need to check if the command you want to reload exists. This can be done in a similar fashion as getting a command in your main file.  
```

guide/command-handling/adding-features.md
```diff
@@ -386,6 +386,62 @@ No more manually editing your help command! If you aren't completely satisfied w
 If you want to add categories or other information to your commands you can simply add properties reflecting it to your `module.exports`. If you only want to show a subset of commands remember that `commands` is a Collection you can <branch version="11.x" inline>[filter](https://discord.js.org/#/docs/main/11.5.1/class/Collection?scrollTo=filter)</branch><branch version="12.x" inline>[filter](https://discord.js.org/#/docs/collection/master/class/Collection?scrollTo=filter)</branch> to fit your specific needs!
 :::
 
+## Reloading commands
+
+When writing your commands, you may find it tedious to restart bot every time you want to test even slightest change in your code. However, if you have a command handler, reloading commands can be done with a single command.
```
almostSouji:
```suggestion
When writing your commands, you may find it tedious to restart your bot every time you want to test even the slightest change in your code. However, if you have a command handler, reloading commands can be done with a single bot command.
```

guide/command-handling/adding-features.md
```diff
@@ -386,6 +386,62 @@ No more manually editing your help command! If you aren't completely satisfied w
...
+
+Create new command file and paste in the usual format:
```
almostSouji:
```suggestion
Create a new command file and paste in the usual format:
```

guide/command-handling/adding-features.md
```diff
@@ -386,6 +386,62 @@ No more manually editing your help command! If you aren't completely satisfied w
...
+	name: 'reload',
+	description: 'Reloads a command',
+	execute(message, args) {
+		// ...
+	},
+};
+```
+
+In this command, you will be using command's name or alias as an only argument. First off, you need to check if command you want to reload exists. This can be done in similiar fashion as getting a command in your main file.  
+Note that you can skip first line if you use [argument checker](/command-handling/adding-features.html#required-arguments) from above:
```
almostSouji:
```suggestion
Note that you can skip the first line if you use the [argument checker](/command-handling/adding-features.html#required-arguments) from above:
```

guide/command-handling/adding-features.md
```diff
@@ -386,6 +386,62 @@ No more manually editing your help command! If you aren't completely satisfied w
...
+```js
+if (!args.length) return message.channel.send(`You didn't pass any command to reload, ${message.author}!`);
+const commandName = args[0].toLowerCase();
+const command = message.client.commands.get(commandName)
+	|| message.client.commands.find(cmd => cmd.aliases && cmd.aliases.includes(commandName));
+
+if (!command) return message.channel.send(`There is no command with name or alias \`${commandName}\`, ${message.author}!`);
+```
+
+`message.client`? Yes! Every library specific object has `client` as a property. That means that you don't have to pass it as a parameter to commands to access for example `client.guilds` or `client.commands` which you will need now.
```
almostSouji: I suggest rephrasing this to a more factual approach and refactoring it into a tip element
"`message.client` here refers to..." or something along those lines

guide/command-handling/adding-features.md
```diff
@@ -386,6 +386,62 @@ No more manually editing your help command! If you aren't completely satisfied w
...
+
+Now, in theory, all there is to do, is to delete the previous command from `client.commands` and require the file again. In practice though, you cannot do this that easy, as `require()` caches the file. If you would require it again, you would simply load previously cached file without any of your changes. In order to remove the file from the cache, you need to add the following line to your command's code:
```
almostSouji:
```suggestion
Now, in theory, all there is to do, is to delete the previous command from `client.commands` and require the file again. In practice though, you cannot do this that easily as `require()` caches the file. If you were to require it again, you would simply load the previously cached file without any of your changes. In order to remove the file from the cache, you need to add the following line to your code:
```

guide/command-handling/adding-features.md
```diff
@@ -386,6 +386,62 @@ No more manually editing your help command! If you aren't completely satisfied w
...
+
+```js
+delete require.cache[require.resolve(`./${commandName}.js`)];
+```
+
+After removing file from cache, all you have to do is require the file and add freshly loaded command to `client.commands`:
```
almostSouji:
```suggestion
After removing the command from the cache, all you have to do is require the file again and add the freshly loaded command to `client.commands`:
```

guide/command-handling/adding-features.md
```diff
@@ -386,6 +386,62 @@ No more manually editing your help command! If you aren't completely satisfied w
...
+try {
+	const newCommand = require(`./${commandName}.js`);
+	message.client.commands.set(newCommand.name, newCommand);
+} catch (error) {
+	console.log(error);
+	message.channel.send(`There was an error while reloading a command \`${commandName}\`:\n\`${error.message}\``);
+}
+```
+
+Snippet above uses `try/catch` to load the command file and add it to `client.commands`, and in case of error - log whole error to console and notify about the fact sending only short `error.message`. Note the fact that we never actually delete the command from commands collection, and instead just overwrite it. This prevents from deleting a command, and ending up with no such command in collection upon failed `require()`, as each use of the reload command checks that collection again.
```
almostSouji:
```suggestion
The snippet above uses a `try/catch` block to load the command file and add it to `client.commands`. In case of an error it will log the full error to the console and notify the user about it with the errors message component `error.message`. Note that you never actually delete the command from the commands collection and instead just overwrite it. This prevents you from deleting a command and ending up with a failed `require()` call, as each use of the reload command checks that collection again.
```
> as each use of the reload command checks that collection again.

I'm not entirely sure what you want to say here, but to me that sentence doesn't make a lot of sense as is.

guide/command-handling/adding-features.md
```diff
@@ -386,6 +386,62 @@ No more manually editing your help command! If you aren't completely satisfied w
...
+
+Last thing you might want to add is sending a message on succesful reload:
```
almostSouji:
```suggestion
The last thing you might want to add is sending a message if the reload was successful:
```

guide/command-handling/adding-features.md
tipakA: Command handler in my bot breaks when i for example have a syntax error in the file i reload, so the command is not anymore in that collection, and every other use of command checks the collection, and not if you actually have such file.

almostSouji: Thanks for the clarification, yeah, that indeed makes sense!

guide/command-handling/adding-features.md
```diff
@@ -386,6 +386,64 @@ No more manually editing your help command! If you aren't completely satisfied w
...
+if (!args.length) return message.channel.send(`You didn't pass any command to reload, ${message.author}!`);
+const commandName = args[0].toLowerCase();
+const command = message.client.commands.get(commandName)
+	|| message.client.commands.find(cmd => cmd.aliases && cmd.aliases.includes(commandName));
+
+if (!command) return message.channel.send(`There is no command with name or alias \`${commandName}\`, ${message.author}!`);
+```
+
+:::tip
+Every library specific object has `client` as a property. That means that you don't have to pass it as a parameter to commands to access for example `client.guilds` or `client.commands`, but can access it directly from `message` object, as in the snippet above.
```
almostSouji:
```suggestion
A lot of library specific structures have `client` as a property. That means you don't have to pass the client reference as a parameter to commands to access for example `client.guilds` or `client.commands`, but can directly access the respective properties directly from the `message` object, as shown in the snippet above.
```

PR #282, opened 2019-11-04
Add issue templates

almostSouji:
This PR proposes some issue templates for use on the guide, filing as PR to allow feedback/requests for additions/changes/wording improvements etc.

This removes the need for manual tagging for the most common issue types.
Where applicable uses the standard GitHub suggestion, possibly adapted to fit the context of the guide better.

* Added the `branching` tag to mark requests and PRs that deal with branching content (e.g. needs v12 implementation) which will automatically be assigned if the corresponding issue template is selected.

### DIFF
--- a/.github/ISSUE_TEMPLATE/branching-content.md
+++ b/.github/ISSUE_TEMPLATE/branching-content.md
@@ -1,0 +1,13 @@
+---
+name: Branching Content
+about: Content is missing for a certain version and requires branching sections
+  or code samples
+labels: 'code refactoring'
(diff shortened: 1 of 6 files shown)


almostSouji:
> What do you think about keeping the text under the labels as comments instead of keeping them there? 

Definitely! That one was of my own issues with it, however i mostly went with what GH seems to propose/consider as "standard" and adapted it to fit the guide context.

Thanks for the feedback, will apply it (hopefully soon)

PR #295, opened 2019-12-01
Add topic to Reactions about reacting in a loop

duckthecuck:
This pull requests adds a topic to the Reactions page which talks about how to react using a loop and how to use an array of emojis to react. It also warns about using a `.forEach()` and why it will not work as `for... of` would.

### DIFF
--- a/guide/popular-topics/reactions.md
+++ b/guide/popular-topics/reactions.md
@@ -145,18 +145,18 @@ If you aren't familiar with Promises or `async`/`await`, you can read more about
 Lets say you have the reactions in an array called `reactions` such as:
 
 ```js
-const reactions = ['🍎','🍊','🍇'];
+const reactions = ['🍎', '🍊', '🍇'];
 ```
 
 And you would like to react with the emojis in the array. You can use a simple `for... of` loop to achieve this.
 
 ```js
-const reactions = ['🍎','🍊','🍇'];
+const reactions = ['🍎', '🍊', '🍇'];
 
 client.on('message', async message => {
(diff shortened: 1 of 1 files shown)


Danktuary:
I personally don't feel the need for a section like this. The ["Reacting in order"](https://discordjs.guide/popular-topics/reactions.html#reacting-in-order) and ["Handling multiple reactions if the order doesn't matter"](https://discordjs.guide/popular-topics/reactions.html#handling-multiple-reactions-if-the-order-doesn-t-matter) sections both show examples of reacting with multiple emojis (without a loop), and a loop is a basic concept readers should already know; the reader should be able to take from those examples and construct a loop themselves if that's what they want/need.

duckthecuck:
Ive personally encountered a lot of occurences where people either don't know how to do that because its async and it does not do it in order, or occurences where people are very confused on why its not working because they either do not await their loop or use a forEach and are confused on why its not in order even if they declare it async and await it. i believe adding something like this would clear up some of that confusion.

almostSouji:
The job of the guide is to introduce concepts used in the library and how to achieve basic functionality by utilizing it. It is and should be up to the reader to enhance and expand the given examples to fit their specific use case.  If we start to mention and explain every little variant of a problem we run into the issue of scope creep quite quickly. 

Providing the literal solution to anything a user might encounter is simply not the purpose of this guide, and as Danktuary already said: the relevant concepts are already explained on this exact page, applying the concepts and approaches to one another is and should be left as exercise to the reader.

PR #296, opened 2019-12-03
Fixing a small bug

FizzesShark:
The current code returns an error, as message.client does not return the list of commands, as the rest of the code seems to assume.

### DIFF
--- a/code-samples/command-handling/adding-features/11/commands/help.js
+++ b/code-samples/command-handling/adding-features/11/commands/help.js
@@ -8,7 +8,7 @@ module.exports = {
 	cooldown: 5,
 	execute(message, args) {
 		const data = [];
-		const { commands } = message.client;
+		const { commands } = message.client.commands;
 
 		if (!args.length) {
 			data.push('Here\'s a list of all my commands:');


tipakA:
Uh, that's destructuring, it should work normally

didinele:
uh... what..?

You should research stuff and test them before openning issues y'know.

As far as I know any Node version over 8 supports destructuring.

Here's how it works:
```js
const foo = {
  bar: 'test'
});

const { bar } = foo;

console.log(bar); // test```

[Here's the MDN page on destructuring](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Destructuring_assignment#Object_destructuring)

FizzesShark:
The reason I changed it was because I tested this on my computer, and it was returning errors as I couldn't call commands.map().

tipakA:
Errors like...

Also, this isn't really place for that, you can join [Discord server](https://discord.gg/bRCvFy9) for easier communication

didinele:
Your... "fix" breaks it even more? Our code right now is the same as doing `message.client.commands` and yours is doing `message.client.commands.commands`

PR #299, opened 2019-12-20
In the help command, added args: true

mikreo:
This is to keep in line with the Required Arguments section


tipakA:
But that doesn't make sense. If you will require arguments for help command, you will not be able to use command to list all commands.

mikreo:
Oh sorry, I forgot about that possibility.

PR #300, opened 2019-12-23
Create config.yml

NightYoshi370:
Adds a direct link to Discord Server in the issues page


almostSouji:
This PR should also remove the current "ask questions" template then

NightYoshi370:
I disagree since you could be asking about the guide itself but I've still done so.

almostSouji:
> I disagree since you could be asking about the guide itself but I've still done so.

You are right! I absolutely missed that use case, although the structure of the template clearly suggests so (and was the one making the PR adding it)

It seems like I mixed it up with the library, where the template indeed only serves the purpose of redirecting questions about the library.

NightYoshi370:
I've added it back and added a link to the library's github repository. This should clarify what goes where

almostSouji:
Thanks for the PR and sorry for the confusion!

PR #304, opened 2020-01-02
Add topic to reactions page about removing reactions

RDambrosio016:
This PR adds a topic to the reactions page focusing on how to remove reactions. it talks about:
- removing all reactions
- removing reactions by emoji
- removing reactions by user
in accordance with https://github.com/discordjs/guide/projects/1#card-16181089

### DIFF
--- a/guide/popular-topics/reactions.md
+++ b/guide/popular-topics/reactions.md
@@ -157,7 +157,7 @@ The benefit of this small optimization is that you can use `.then()` to handle w
 
 ## Removing reactions
 
-Now that you know how to add reactions, you might be asking, how do we remove them? In this section you will learn how to remove all reactions, remove reactions by user, and remove reactions by emoji.
+Now that you know how to add reactions, you might be asking, how do you remove them? In this section you will learn how to remove all reactions, remove reactions by user, and remove reactions by emoji.
 
 ::: warning
 All of these methods require `MANAGE_MESSAGES` permissions. Make sure your bot has permissions before attempting to utilize any of these methods, as it will error if it doesnt.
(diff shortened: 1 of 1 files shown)


guide/popular-topics/reactions.md
```diff
@@ -155,6 +155,127 @@ if (message.content === '!fruits') {
...
+
+### Removing all reactions
+
+Removing all reactions from a message is the easiest, the API allows us to do this through a single call. It can be done through the <branch version="11.x" inline> `message.clearReactions()` </branch> <branch version="12.x" inline> `message.reactions.removeAll()` </branch> method. 
+
+<branch version="11.x">
+
+```js
+message.clearReactions()
+	.catch(err => console.error('Failed to clear reactions: ', err));
```
Danktuary: This can be a one-liner, and `err` should be `error`.

guide/popular-topics/reactions.md
```diff
@@ -155,6 +155,127 @@ if (message.content === '!fruits') {
...
+Removing reactions by emoji is not as straightforward as clearing all reactions. The API does not provide a method for selectively removing reactions by emoji, it only allows us to remove a user from a specific reaction. This means we will have to get the users who reacted with that emoji, and loop through and remove each one of them. Firstly, we need to get the reaction, for unicode emojis, this is simple as escaping the emoji in discord and copying the result as you saw before.
+
+Reaction collections are keyed by <branch version="11.x" inline>`name:id`</branch> <branch version="12.x" inline>`id`</branch> for custom emojis and by `name` for unicode emojis. This means we can simply run a `.get()` on `message.reactions` to get the reaction representing the emoji we want. After we have the reaction, we can loop through `reaction.users` and call <branch version="11.x" inline>`reaction.remove(user)`</branch> <branch version="12.x" inline>`reaction.users.remove()`</branch> on each of them.
+
+For this example we will fetch a specific message for which we want to remove a specific emoji's reaction on.
+
+<branch version="11.x">
+
+```js
+<!-- eslint-skip -->
```
Danktuary: These `<!-- eslint-skip -->`s should be outside the codeblock, or else they'll show up on the live site.

guide/popular-topics/reactions.md
```diff
@@ -155,6 +155,127 @@ if (message.content === '!fruits') {
...
+//	Get the reaction representing the emoji
```
Danktuary: This comment and the next aren't necessary; it's fairly easy to understand what the code does by reading it.

guide/popular-topics/reactions.md
```diff
@@ -155,6 +155,127 @@ if (message.content === '!fruits') {
...
+	.catch(err => console.error('Failed to clear reactions: ', err));
+```
+
+</branch>
+
+### Removing reactions by emoji
+
+Removing reactions by emoji is not as straightforward as clearing all reactions. The API does not provide a method for selectively removing reactions by emoji, it only allows us to remove a user from a specific reaction. This means we will have to get the users who reacted with that emoji, and loop through and remove each one of them. Firstly, we need to get the reaction, for unicode emojis, this is simple as escaping the emoji in discord and copying the result as you saw before.
+
+Reaction collections are keyed by <branch version="11.x" inline>`name:id`</branch> <branch version="12.x" inline>`id`</branch> for custom emojis and by `name` for unicode emojis. This means we can simply run a `.get()` on `message.reactions` to get the reaction representing the emoji we want. After we have the reaction, we can loop through `reaction.users` and call <branch version="11.x" inline>`reaction.remove(user)`</branch> <branch version="12.x" inline>`reaction.users.remove()`</branch> on each of them.
```
Danktuary: The final sentence in the paragraph before could be moved into this one, or removed entirely. It could be changed to something along the lines of:

> Reaction collections are keyed by <branch version="11.x" inline>`name:id`</branch> <branch version="12.x" inline>`id`</branch> for custom emojis and by `name` for unicode emojis (represented by their unicode character, see [here](/reactions.html#unicode-emojis)). Once you have the key, you can simply run a `.get()` on `message.reactions` to get the reaction representing the emoji you want.```

guide/popular-topics/reactions.md
RDambrosio016: Woops, interpreted it as on top of the code block and inside of it

RDambrosio016: That was for the eslint skip comment, god im a mess today

guide/popular-topics/reactions.md
```diff
@@ -155,6 +155,112 @@ if (message.content === '!fruits') {
...
+```js
+const userReactions = message.reactions.filter(reaction => reaction.users.has(userId));
+try {
+	for (const reaction of userReactions) {
+		await reaction.users.remove(userId);
+	}
+} catch (error) {
+	console.error('Failed to remove reactions.');
+}
+});
```
Danktuary: Additional unneeded `});` here

guide/popular-topics/reactions.md
```diff
@@ -155,6 +155,112 @@ if (message.content === '!fruits') {
...
+
+Now that you know how to add reactions, you might be asking, how do you remove them? In this section you will learn how to remove all reactions, remove reactions by user, and remove reactions by emoji.
+
+::: warning
+All of these methods require `MANAGE_MESSAGES` permissions. Make sure your bot has permissions before attempting to utilize any of these methods, as it will error if it doesnt.
+:::
+
+### Removing all reactions
+
+Removing all reactions from a message is the easiest, the API allows you to do this through a single call. It can be done through the <branch version="11.x" inline> `message.clearReactions()` </branch><branch version="12.x" inline> `message.reactions.removeAll()` </branch> method.
```
Danktuary: No need for the spaces inside the `<branch>` components.

guide/popular-topics/reactions.md
```diff
@@ -155,6 +155,112 @@ if (message.content === '!fruits') {
...
+}
+```
+
+</branch>
+
+The reason we use a `for... of` loop over something like `forEach()` is due to `forEach()`'s behavior for async operations. `forEach()` will send out all calls almost at once even if we await inside of the function. However, if we `await` inside of a `for... of` loop, it will wait for the previous reaction to go through, and we avoid spamming the API with a lot of calls at once.
+
+### Removing reactions by user
+
+Removing reactions by user is similar to what you did before. However, instead of iterating through users of a reaction, you will iterate through reactions which include a user. To do this you will get all reactions and filter based on whether the user has reacted. If you are not familiar with `Collection.filter()` and `Collection.has()` take the time to understand what they do and then come back.
```
Danktuary: I think the last sentence could be a `::: tip` block, since that's usually what we do with these sort of notes. Links to the methods would be nice too.

guide/popular-topics/reactions.md
```diff
@@ -155,6 +155,112 @@ if (message.content === '!fruits') {
...
+		await reaction.users.remove(user);
+	}
+} catch (error) {
+	console.error('Failed to remove reactions.');
+}
+```
+
+</branch>
+
+The reason we use a `for... of` loop over something like `forEach()` is due to `forEach()`'s behavior for async operations. `forEach()` will send out all calls almost at once even if we await inside of the function. However, if we `await` inside of a `for... of` loop, it will wait for the previous reaction to go through, and we avoid spamming the API with a lot of calls at once.
```
Danktuary: `for... of` -> `for...of`, `forEach()` -> `.forEach()`

PR #305, opened 2020-01-05
Add webhook section

RDambrosio016:
This PR adds a section to the `popular-topics` tab about working with webhooks. it currently includes:
- What webhooks are
- Detecting webhook messages
- Fetching webhooks
- Creating webhooks
- Editing webhooks
- Using webhooks

It talks about the above topics with respect to `Webhook` and `WebhookClient`. 
I wrote this when i was tired, so revisions will be made, and any feedback is greatly appreciated 👍


guide/popular-topics/webhooks.md
```diff
@@ -0,0 +1,209 @@
...
+	avatar: 'https://i.imgur.com/IgscS45.jpg',
+	channel: '222197033908436994',
+}).then(webhook => console.log(`Edited webhook ${webhook}`)).catch(console.error);
+```
+
+</branch>
+
+## Using webhooks
+
+Webhooks, unlike bots, can send more than one embed per message, up to 10. They can also send attachments and normal content. The <branch version="11.x" inline> [`Webhook.send()`](https://discord.js.org/#/docs/main/stable/class/Webhook?scrollTo=send)</branch><branch version="12.x" inline>[`Webhook.send()`](https://discord.js.org/#/docs/main/master/class/Webhook?scrollTo=send)</branch> method to send to a webhook is very similar to the method for sending to a text channel. Webhooks can also change their name and avatar when sending a message.
```
almostSouji: > Webhooks can also change their name and avatar when sending a message.

Not quite accurate, they can send an avatar and name to be used for this specific request, yes, it does not really "change" them in any way, webhooks just accept these parameters in the payload instead of inferring them from the logged in client, like the case for "standard" messages

PR #310, opened 2020-01-07
Update node version for d.js master

PyroTechniac:
updates required node version for the master branch

I'm not sure if this is the only spot it's mentioned, so let me know if there is anything I missed


PyroTechniac:
Should be fixed in the latest commit

almostSouji:
> 
> 
> Should be fixed in the latest commit

found one more

PR #328, opened 2020-02-15
feat/fix: more branch tags + some typos

NotSugden:
This PR adds some more branch tags (some that i may have missed in my previous PR), specifically for the new managers commit, adding .cache in various places missed in @RDambrosio016's PR

also fixes some typos along the way

MessageReaction#fetch fetches the message the reaction belongs to, and updates the reaction using that data, so there is no need to fetch both the reaction, and the message seperately, ~~(it currently doesn't update the existing message if its a partial, i made a PR https://github.com/discordjs/discord.js/pull/3789 to fix this however)~~ pr was merged

(reference for only displaying the warning about the color not being resolved on v11 https://github.com/discordjs/discord.js/commit/52bc5b0170e290291d9760c13a1969874f936e6d)

### DIFF
--- a/guide/sharding/extended.md
+++ b/guide/sharding/extended.md
@@ -276,7 +276,7 @@ Now, run this code, and you will surely get a result that looks like the followi
     _roles: [] } ]
 ```
 
-While this result isnt *necessarily* bad or incorrect, it's simply a raw object that got `JSON.parse()`'d and `JSON.stringify()`'d over, so all of the circular references are gone. More importantly, The object is no longer a true <branch version="11.x">`Emoji`</branch><branch version="11.x">`GuildEmoji`</branch> object as provided by discord.js. This means none of the convenience methods usually provided to you are available. If this is not a concern to you, then you can effectively skip the rest of this section. However, this is a tutorial, so it should be covered regardless! Let's remedy this issue, shall we?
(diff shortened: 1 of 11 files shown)


guide/sharding/extended.md
```diff
@@ -153,7 +276,9 @@ Now, run this code, and you will surely get a result that looks like the followi
     _roles: [] } ]
 ```
 
-While this result isnt *necessarily* bad or incorrect, it's simply a raw object that got `JSON.parse()`'d and `JSON.stringify()`'d over, so all of the circular references are gone. More importantly, The object is no longer a true `Emoji` object as provided by discord.js. This means none of the convenience methods usually provided to you are available. If this is not a concern to you, then you can effectively skip the rest of this section. However, this is a tutorial, so it should be covered regardless! Let's remedy this issue, shall we?
+While this result isnt *necessarily* bad or incorrect, it's simply a raw object that got `JSON.parse()`'d and `JSON.stringify()`'d over, so all of the circular references are gone. More importantly, The object is no longer a true <branch version="11.x">`Emoji`</branch><branch version="11.x">`GuildEmoji`</branch> object as provided by discord.js. This means none of the convenience methods usually provided to you are available. If this is not a concern to you, then you can effectively skip the rest of this section. However, this is a tutorial, so it should be covered regardless! Let's remedy this issue, shall we?
```
almostSouji:
```suggestion
While this result isnt *necessarily* bad or incorrect, it's simply a raw object that got `JSON.parse()`'d and `JSON.stringify()`'d over, so all of the circular references are gone. More importantly, The object is no longer a true <branch version="11.x">`Emoji`</branch><branch version="12.x">`GuildEmoji`</branch> object as provided by discord.js. This means none of the convenience methods usually provided to you are available. If this is not a concern to you, then you can effectively skip the rest of this section. However, this is a tutorial, so it should be covered regardless! Let's remedy this issue, shall we?
```
the latter branch should be v12

guide/sharding/extended.md
```diff
@@ -173,6 +298,29 @@ function findEmoji(id) {
 }
 ```
 
+</branch>
+<branch version="11.x">
```
almostSouji:
```suggestion
<branch version="12.x">
```
11 twice again

PR #337, opened 2020-02-29
Fix inline branch tags

tipakA:
Recent #328 left new inline branch tags without inline property, causing them to not display as inline. This fixes those.


almostSouji:
thank you very much, did not notice when reviewing that PR, will give proper tags more attention going forward!

PR #340, opened 2020-03-02
Change sentence structure to eliminate confusion

yuta0801:
The explanation of abort Breaking Changes and Deletions section in the previous section is a bit strange, and I think it will confuse readers as the horizontal line is quite inconspicuous due to the style of the site.
And comment out the message that seems to be guide editors not needed by readers


almostSouji:
resolved with f435ade4039add6b659067329a93326dba4a9ccf

PR #349, opened 2020-03-12
Added Module.exports tip

diligamer:
Sometimes, You need to add more info to help command, this informs users of that.

### DIFF
--- a/guide/command-handling/adding-features.md
+++ b/guide/command-handling/adding-features.md
@@ -283,6 +283,9 @@ module.exports = {
 	},
 };
 ```
+::: tip
+You Can Export Anything In Module.exports and call it in help
+:::
 
 You're going to need your prefix variable a couple times inside this command, so make sure to require that at the very top of the file (outside of the `module.exports` bit).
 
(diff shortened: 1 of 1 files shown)


guide/command-handling/adding-features.md
```diff
@@ -283,6 +283,9 @@ module.exports = {
 	},
 };
 ```
+::: tip
+You Can Export Anything In Module.exports and call it in help
```
tipakA: If anything, this would have to not be overcapitalized like that.
Additionally, exported object is available not only in help command, so i think this sounds slightly more fitting.
```suggestion
You can export anything in `module.exports` object and access it later.
```
On the other hand, "anything" may not exactly be true either, but that would depend on reader.

diligamer: Sounds Good!

diligamer:
@tipakA  When Will This Be Merged

tipakA:
Uh, merging pull requests does not belong to me, but i can try to pass this to someone that actually can merge.

almostSouji:
This information is redundant to the already provided tip further down the same file
> If you want to add categories or other information to your commands you can simply add properties reflecting it to your module.exports. If you only want to show a subset of commands remember that commands is a Collection you can filter to fit your specific needs

PR #354, opened 2020-03-24
feat: Gateway Intents article

Fyko:
**Please describe the changes this PR makes and why it should be merged:**
This PR creates another article for the Discord.js guide covering [Gateway Intents](https://github.com/discord/discord-api-docs/issues/1363), what they do and how to use them.

PR #356, opened 2020-03-26
Add information about intents

BrightnessTM:
This PR adds information to the v11 to v12 update guide about `disabledEvents` being removed in favor of specifying intents in `clientOptions`.


guide/additional-info/changes-in-v12.md
```diff
@@ -548,6 +549,16 @@ There have been several changes made to the `ClientOptions` object located in `c
 
 `clientOptions.sync` has been removed entirely, along with all other user account-only properties and methods.
 
+#### ClientOptions#disabledEvents
+
+`clientOptions.disabledEvents` has been removed in favor of using intents
+
+```diff
+- const client = new Discord.Client({ disabledEvents: [] })
++ const intents = new Discord.Intents([]);
```
almostSouji: This would compute empty intents, causing the bot to not emit any events, the guide should not suggest that as replacement here, instead this should have a note leading to the section about to be added with #354 explaining the process of computing intents and what it entails in much more detail than possible in this short update part.

This section should also (shortly) mention that specifying intents may break the internal capabilities of the library and that users should only opt out of events if they know what they're getting into

PR #357, opened 2020-03-27
fix: for..of iteration in reactions guide

PASTUSEDBY:

Whenever you iterate over a Collection with ``for...of`` loop, it has two parameters, a key and a value. This pr fixes those parts where ``for (const reaction of userReactions)`` is used, because the first parameter of the collection represents the `Snowflake`, and the second represents the `Object`. This has been fixed by the use of destructured array here.

### DIFF
--- a/guide/popular-topics/reactions.md
+++ b/guide/popular-topics/reactions.md
@@ -256,7 +256,7 @@ Removing reactions by user is similar to what you did before. However, instead o
 ```js
 const userReactions = message.reactions.filter(reaction => reaction.users.has(userId));
 try {
-	for (const reaction of userReactions) {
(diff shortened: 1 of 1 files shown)


almostSouji:
not a typo, but a bug fix
since the id is not used however this should really use the [ `<Map>.values()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map/values) iterator rather than the destructuring approach proposed here.

PASTUSEDBY:
Fixed with [Map#values](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map/values) instead of the destructured method as the id is not really required, mainly, the corresponding object is required.

PR #362, opened 2020-04-02
fix(Permissions): overwritePermissions example for v12

Monbrey:
The example was wrong.

Now it isn't.


RDambrosio016:
nice duplicate of #361

Monbrey:
nvm I should look at open PRs first.

Duplicate of #361

PR #366, opened 2020-04-07
fix: expect filename to be commandname

thrien:
The reloading command expects the file to be called `./${command.name}.js`. Even though this is reasonable it might be better not to rely on it. Also the command handler does not make such a requirement.

The suggested solution is to search the old module in the cache and read it's id.

However this solution might be more confusing for the reader.

### DIFF
--- a/guide/command-handling/adding-features.md
+++ b/guide/command-handling/adding-features.md
@@ -421,7 +421,7 @@ A lot of library specific structures have `client` as a property. That means you
 Now, in theory, all there is to do, is to delete the previous command from `client.commands` and require the file again. In practice though, you cannot do this that easily as `require()` caches the file. If you were to require it again, you would simply load the previously cached file without any of your changes. In order to remove the file from the cache, you need to add the following lines to your code:
 
 ```js
-const id = Object.values(require.cache).find(module => module.exports === command).id;
+const { id } = Object.values(require.cache).find(module => module.exports === command);
 delete require.cache[id];
 ```
 
--- a/guide/command-handling/adding-features.md
+++ b/guide/command-handling/adding-features.md
@@ -418,17 +418,18 @@ if (!command) return message.channel.send(`There is no command with name or alia
 A lot of library specific structures have `client` as a property. That means you don't have to pass the client reference as a parameter to commands to access for example `client.guilds` or `client.commands`, but can directly access the respective properties directly from the `message` object, as shown in the snippet above.
 :::
 
-Now, in theory, all there is to do, is to delete the previous command from `client.commands` and require the file again. In practice though, you cannot do this that easily as `require()` caches the file. If you were to require it again, you would simply load the previously cached file without any of your changes. In order to remove the file from the cache, you need to add the following line to your code:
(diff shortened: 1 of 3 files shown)


guide/command-handling/adding-features.md
```diff
@@ -418,17 +418,18 @@ if (!command) return message.channel.send(`There is no command with name or alia
...
 :::
 
-Now, in theory, all there is to do, is to delete the previous command from `client.commands` and require the file again. In practice though, you cannot do this that easily as `require()` caches the file. If you were to require it again, you would simply load the previously cached file without any of your changes. In order to remove the file from the cache, you need to add the following line to your code:
+Now, in theory, all there is to do, is to delete the previous command from `client.commands` and require the file again. In practice though, you cannot do this that easily as `require()` caches the file. If you were to require it again, you would simply load the previously cached file without any of your changes. In order to remove the file from the cache, you need to add the following lines to your code:
 
 ```js
-delete require.cache[require.resolve(`./${command.name}.js`)];
+const { id } = Object.values(require.cache).find(module => module.exports === command);
```
tipakA, line 424: Will that find there ever return anything? As far as i understand that's comparing raw objects between require's cache and command object taken out from `client.commands` Collection, which would never actually return `true`.

thrien: I tested it out and it worked

thrien: Also the exported object of the module is compared

thrien: From the [Node.js Documentation](https://nodejs.org/api/modules.html#modules_caching "Modules Caching"):
> Modules are cached after the first time they are loaded. This means (among other things) that every call to `require('foo')` will get exactly the same object returned, if it would resolve to the same file.

I guess the one in the cache is also exactly the same.

tipakA: That's not exactly what i asked about. Value in the `client.commands` Collection is not a call to require the command file, but already an object on its own.
What i meant wasn't that objects are different from each other, but that in JavaScript, compring non-primitives (like objects) using `===` doesn't check structure, but if they are distinct from each other, as somewhat explained [here](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Equality_comparisons_and_sameness).

thrien: In the main file we essentially do `client.commands.set(..., require(filename))` and now we do `client.commands.get(...)`. So `command` is exactly the same object as the one `require()` returned, which is `module.exports`.
Is that right? Or am i missing something and `.set()` and `.get()` don't undo themselves!? Or something else?

ItsPapaia:
Why not just have the path on the command object?
You will just have to delete the cache for `command.filepath` and reload the same file instead of checking each file.

thrien:
That's actually how i ended up doing it in my own project, but it would probably require a bit more explanation in the guide.
It's probably better to just leave it like it is, in order to keep the guide more concise. This is after all a minor issue.
If people want to name their files more freely they can choose to follow one of these approaches or find their own solution.

Danktuary:
For simplicity's sake, I think it's okay to keep it as it is. What's there now will work fine in basic use-cases, and should readers want to expand upon the given code, that's up to them.

PR #381, opened 2020-04-25
Add missing 's' in method call

iAmZiNc:
Typo fix on the MessageEmbed#addFields reference for blank fields.

### DIFF
--- a/guide/additional-info/changes-in-v12.md
+++ b/guide/additional-info/changes-in-v12.md
@@ -1369,7 +1369,7 @@ The `max` and `maxMatches` properties of the `MessageCollector` class have been 
 
 #### MessageEmbed#addBlankField
 
-`messageEmbed.addBlankField()` has been removed entirely. To add a blank field, use `messageEmbed.addField('\u200b', '\u200b')`.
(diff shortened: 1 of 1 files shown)


iAmZiNc:
Unfamiliar with how to add labels, but this falls under the spelling/grammar label.

tipakA:
There is no big reason to change that, `addField` calls `addFields` internally anyway.
https://github.com/discordjs/discord.js/blob/1726651c71ed68aa4eb1bbe2fe30a014d8ef4f4d/src/structures/MessageEmbed.js#L249-L251

almostSouji:
> Unfamiliar with how to add labels, but this falls under the spelling/grammar label.

Labels are either automatically added by picking the right issue template (not applicable to pull requests) or by bots or maintainers, nothing you can do yourself. A change like this couldn't be interpreted as typo either, due to the reason detailed below:

Not only is the switch to `addFields` obsolete as `addField` is still valid, the proposed code is invalid, as each value will be interpreted as a field through spreading (each being a string) without the respective object values that are expected when using `addFields` resulting in two fields with "undefined" as both name and value.

If anything it needs to be `.addFields({name: "\u200B", value: "\u200B"})`, which however is not needed.

PR #383, opened 2020-04-26
Changed the definition of args slightly

Tri-tip:
Used the trim method on args so that it eliminates white-space at the end of the command. This may make things slightly more effective.


tipakA:
This should also be changed in every other place that defines `args`, including respective code samples.

PR #390, opened 2020-04-29
Update "which role was added/removed" example v12

anandre:
Use the Collection#difference method for v12 on the "which roles changed" example, because that is a good use case for it.


anandre:
Closing 'cause this is more work than the filter method.

PR #393, opened 2020-05-03
Tweaks to Sharding guide

kav2k:

This PR has multiple improvements to the Sharding guide:

1. The guide stated that for 3600 guilds and 4 shards, they will be distributed as `[1000, 1000, 1000, 600]`. My best understanding of sharding is that every shard will have ~900 guilds. The guide was updated to reflect it.
2. A notice was added that ShardingManager can only run shards on the same machine (as processes / threads).
3. Some rewording (e.g. "I" -> "we") and clarifications
4. Small formatting fix: newline after `</branch>`

**Why is the PR submitted as a draft**

I never worked with sharding, so I would like confirmation that my understanding is correct. Also, perhaps there's a better place for the remark about scaling beyond processes on the same machine.

### DIFF
--- a/guide/sharding/README.md
+++ b/guide/sharding/README.md
@@ -2,7 +2,7 @@
 
 ## When to shard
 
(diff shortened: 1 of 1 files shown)


kav2k:
Rebased off master.

PR #397, opened 2020-05-09
fix(embeds): addFields should be passed an Array

Monbrey:


almostSouji:
The implementation of `addFields` uses argument spreading to both allow an array as well as a variable amount of arguments to be passed. The choice to demonstrate the latter here is deliberate. I do see how this might be confusing though and would be okay with adding a short explanation or enhance the example to demonstrate both behaviors as you see fit.

ref(code): <https://github.com/discordjs/discord.js/blob/stable/src/structures/MessageEmbed.js#L259>
ref(docs): <https://discord.js.org/#/docs/main/stable/class/MessageEmbed?scrollTo=addFields> (reflects this in the argument notation `...fields`)

almostSouji:
Since there hasn't been any activity: if you want to apply a note about this please feel free to do so in a separate PR

PR #401, opened 2020-05-12
Make the emoji characters correct

tbnritzdoge:


tbnritzdoge:
I made this so people like me who refer to this to get the characters can actually use them

almostSouji:
> 
> 
> I made this so people like me who refer to this to get the characters can actually use them

Please elaborate on this. What exactly is the issue with the current characters and where do they cause which problems?

tbnritzdoge:
I put them inside my bot for usage and they weren't sendable, but these are

tbnritzdoge:
image
Those numbers specifically didn't work

NotSugden:
i believe there was a change a while back with the tweemoji discord uses (can't remember exactly, but something along those lines) the emojis shown in https://github.com/discordjs/guide/pull/401#issuecomment-629701974 are no longer used, and instead the new ones that are in this PR are instead used

tbnritzdoge:
are there any updates on this?

almostSouji:
I'm still confused about why this change should be necessary.
I tried the current emoji characters out with sending, reacting and comparing from incoming reactions and they seem to work just fine.

What are you trying to do with them that doesn't work but does so after this change is applied?

tbnritzdoge:
Well the ones I provided worked for me when these ones did not, maybe it was a bug or something but as Sugden said there was a change in the discord emojis and the ones provided by the guide currently don't match that. I guess it's also a better look to have them look like the actual emoji instead of 
image that.

NotSugden:
> I'm still confused about why this change should be necessary.
> I tried the current emoji characters out with sending, reacting and comparing from incoming reactions and they seem to work just fine.
> 
> What are you trying to do with them that doesn't work but does so after this change is applied?

the old emojis do still work yes, however, as i mentioned before, the tweemoji discord uses was updated, and now the new emojis (in this PR) are used (eg: when the client reacts, it uses these emojis in this PR)

Monbrey:
tl;dr the old workaround of needing the actual unicode sequence rather than the unicode emoji is no longer required.

tbnritzdoge:
It's also much more visually appealing as they actually render as the emoji instead of showing whatever it does now.

Monbrey:
> It's also much more visually appealing as they actually render as the emoji instead of showing whatever it does now.

I don't disagree, but that's really not relevant. The whole reason this snippet existed was because the visual emojis did not work previously. If that's no longer the case, great

Monbrey:
I'd also like to clarify something about why there's been so much confusion on this PR - the emoji characters still dont render properly in many browsers, which makes this *appear* like nothing is changing.

Here is Firefox and Chrome side by side. To Firefox users, the GitHub diff looks like nothing has changed.

image

almostSouji:
> tl;dr the old workaround of needing the actual unicode sequence rather than the unicode emoji is no longer required.

aha, so what this really is about is that discord now allows the visual representation you get from <https://emojipedia.org/keycap-digit-one/> (amongst other places) so really this snippet is no longer needed or can at the very least be majorly simplified

> Well the ones I provided worked for me when these ones did not

Answering "didn't work" to "what didn't" work is really not much more useful. What did you do when they didn't work, what was the unexpected outcome, please provide a minimal code sample so we can try and reproduce it.

> this could also bring up an issue if someone where to wait for a specific number reaction on a message, 

 Is this actually even the case? I tried everything i could think of with the old snippet yesterday and could not get any unexpected behavior from it

almostSouji:
Anyways, after chatting with @Monbrey for a bit this seems alright.

This snippet was created since the visual representation you could find across the internet in various other resources was not accepted by discord when sent as reaction. This is no longer the case. These work as intended now.

However in discord the number emojis still can not be escaped with the usual `\` escape sequence, which is the chosen way of many (also in our support channels) to retrieve unicode emojis. The snippet still solves this "issue".

After some more testing i still can not reproduce any issues where the old snippet produces errors or unwanted behavior but that seems to be less relevant with the above situation being the case.

PR #407, opened 2020-05-26
Note to not use the Client Secret

CommanderTso:
Add a note to not use the Client Secret from the General Information panel.

In the event the user ignores the warning to read carefully (mea culpa) and is used to apps only having one secret token, it's possible to confuse the Token with the Client Secret.  If you end up in that spot, you'll get an invalid token response from Discord and be very confused.


almostSouji:
The guide shows screen by screen exactly where to go, from the Applications view to the Bot tab, creating a bot and then copying the token. If you can't even follow this pictured step by step guide you're likely going to fail to follow the rest of the guide. An additional yellow box won't change that.

CommanderTso:
@almostSouji  That's completely your right.  But for what it's worth, I took time from my day to try to improve a product I've just barely used.  The rudeness in your reply is unwarranted and unappreciated.

PR #436, opened 2020-07-20
Fixed Small Syntax Error With ping Example

shadowtime2000:
At line 52 there was an unnecessary comma

### DIFF
--- a/guide/command-handling/README.md
+++ b/guide/command-handling/README.md
@@ -49,7 +49,7 @@ module.exports = {
 	description: 'Ping!',
 	execute(message, args) {
 		message.channel.send('Pong.');
-	},
+	}
 };
 ```
 


PyroTechniac:
That comma is perfectly valid

shadowtime2000:
How? There are no elements appearing in it afterward.

iCrawl:
It's a stylistic choice.

shadowtime2000:
There is really no point for that comma to exist.

Fyko:
> 
> 
> There is really no point for that comma to exist.

It's required by the [eslint config](https://github.com/Danktuary/eslint-config-sora/blob/67a9352a3d780897854c7f6284d891f2d4c26e67/index.js#L29). So yes, there is a point.

shadowtime2000:
Oh ok.

PR #441, opened 2020-07-26
Corrected misleading information

GeniusTimo:
I read your article about [gateway intents](https://discordjs.guide/popular-topics/intents.html) in detail and I noticed that the [discord support article](https://support.discord.com/hc/articles/360040720412) says something different conecerning privileged intent whitelisting:
> Please note that this is currently in a deprecation period and you will require a whitelisted bot in order to use privileged intents as of October 7, 2020. ~ https://discordjs.guide/popular-topics/intents.html

> After our deprecation period ends, if your bot is in over 100 servers, you cannot turn on Privileged Intents until you get whitelisted first. ~ https://support.discord.com/hc/articles/360040720412#what-if-im-already-in-100-guilds-1

So whitelisting is only required if a bot is in more than 100 guilds 😅 I beg to have someone look at the language, because I'm actually not very good at English grammar (but I've tried my best 🙈)

### DIFF
--- a/guide/popular-topics/intents.md
+++ b/guide/popular-topics/intents.md
@@ -67,9 +67,11 @@ If you want to view the built flags you can utilize the `.toArray()`, `.serializ
 Discord defines some intents as "privileged" due to the sensitive nature of the data sent through the affected events.
 At the time of writing this article privileged intents are `GUILD_PRESENCES` and `GUILD_MEMBERS`
 
-For now you can simply enable these intents in the [Discord Developer Portal](https://discordapp.com/developers/applications) through a toggle. Please note that this is currently in a deprecation period and if your bot is in more then 100 guild it must be whitelisted to use privileged intents as of October 7, 2020. You can find more information on whitelisting in [this discord support article](https://support.discordapp.com/hc/en-us/articles/360040720412-Bot-Verification-and-Data-Whitelisting)
+You can enable privileged gateway intents in the [Discord Developer Portal](https://discordapp.com/developers/applications) under "Privileged Gateway Intents" in the "Bot" section.
 
-Should you receive the error `[DISALLOWED_INTENTS]: Privileged intent provided is not enabled or whitelisted` please review your settings for all privileged intents you use. The official documentation for privileged intents can be found on the [discord API documentation](https://discordapp.com/developers/docs/topics/gateway#privileged-intents).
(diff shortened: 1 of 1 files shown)


guide/popular-topics/intents.md
```diff
@@ -67,7 +67,7 @@ If you want to view the built flags you can utilize the `.toArray()`, `.serializ
...
 At the time of writing this article privileged intents are `GUILD_PRESENCES` and `GUILD_MEMBERS`
 
-For now you can simply enable these intents in the [Discord Developer Portal](https://discordapp.com/developers/applications) through a toggle. Please note that this is currently in a deprecation period and you will require a whitelisted bot in order to use privileged intents as of October 7, 2020. You can find more information on whitelisting in [this discord support article](https://support.discordapp.com/hc/en-us/articles/360040720412-Bot-Verification-and-Data-Whitelisting)
+For now you can simply enable these intents in the [Discord Developer Portal](https://discordapp.com/developers/applications) through a toggle. Please note that this is currently in a deprecation period and if your bot is in more then 100 guild it must be whitelisted to use privileged intents as of October 7, 2020. You can find more information on whitelisting in [this discord support article](https://support.discordapp.com/hc/en-us/articles/360040720412-Bot-Verification-and-Data-Whitelisting)
```
almostSouji, line 70: Flaws in the current iteration:
- [then/than](https://www.merriam-webster.com/words-at-play/when-to-use-then-and-than) here a comparison, not related to time, "than"
- 100 guild**s**, plural

I am however not a huge fan of this massive sentence it has grown to be and that this adaptation still suggest behavior that's not accurate in reality.

In order to join additional guilds after max(100, <current server count>) after the deadline it needs to go through the verification process anyways. This is required. During the verification the developer needs to request access to privileged intents in the application form of the bot needs them. If it doesn't verification is still required.

I propose splitting this in two sentences/sections instead: enabling and caveats at 100+

- You can enable privileged intents...
- Note that bots in 100 or more guilds need to request access [...] during discords required [verification process]"

This takes care of the following inaccuracies:
- "more than 100" implies 101+ -> it's 100+
- verification is only needed for privileged intents -> verification is needed for 100+ guilds anyways, access to priv. intents has to be requested during that process.
- "For now..." -> from all we know this is how it works and likely won't change even with explicit grant through verification.

Additionally the date and deprecation period are mentioned in the article and can be omitted here to make the sentence structure much cleaner.

GeniusTimo:
Thank you 👏

PR #445, opened 2020-07-30
Add an extra way of restarting the bot.

haroongames-git:
Added client.destroy() as a way to shut down the bot, and restart it.


HarmoGlace:
It is not the same thing. The code you provided only restart the WS connection, it doesn't fully restart the process

PR #449, opened 2020-08-06
Updated querystring size parameter for avatar urls

R-Andrei:
Avatar URLs are currently broken because of incorrect size params. If an invalid integer is given to the size param, the resource request returns a "400 Bad Request" status. 

URLs should also work without a size param, but since one was used in the example I decided to correct rather than remove it. 

Updated the size from 2084 to 2048, as it needs to be 2^[3, →) in order to work. (That's 2 to the power of 3 or more)

The image requests should now return an OK status.


almostSouji (changes requested):
~~This needs branching. The fix is right for v11, but not for v12, where the returned URL does not include a size parameter at all with `.displayAvatarURL({ dynamic: true, format: 'png' })`.~~

Additionally the same needs to be applied in [adding-features](https://github.com/discordjs/guide/blob/master/guide/command-handling/adding-features.md) as well.

PR #459, opened 2020-08-17
Updated legend to use v12 syntax

flamelessdev:
With v11 support being dropped soon, this PR updates the legend so it uses up to date v12 methods.


tipakA:
That should be done with the inline version tag, so it can be changed with version selector.

flamelessdev:
Added branching - let me know if there are any other necessary changes.

PR #463, opened 2020-08-27

DevNoahh2:
Made the Emoji characters a little better to understand and bit clearer to new users.

### DIFF
--- a/guide/popular-topics/miscellaneous-examples.md
+++ b/guide/popular-topics/miscellaneous-examples.md
@@ -180,5 +180,5 @@ console.log(emojiCharacters['!']); // ❗
 ```
 
 ::: tip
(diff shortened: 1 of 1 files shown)


almostSouji:
I don't really think this is an improvement of the guide:
- [windows themselves](https://blogs.windows.com/windowsexperience/2018/02/05/windows-10-tip-get-started-emoji-keyboard-shortcut/) only recommend/show the `WIN` + `.` key combo, `WIN` + `;` seems non-standard and confusing for non-querty users where the `;` key might be hidden behind a modification and does not open the emoji picker
- the current "to all the unicode emojis available to you" is semantically correct. Depending on your update status this does not necessarily include all emojis that are agreed on by the consortium.
- The naming of websites (if the sole purpose is to provide a native alternative to them) is not necessary and seems biased, there are a ton out there and i can not really recommend any of them

In summary I don't think the purpose "making the section clearer to new users" is fulfilled. The second shortcut makes it more confusing for a subset of people and the PR introduces a semantic downgrade.

PR #469, opened 2020-08-29
added a quick tip and a new section on permissions

storate:

added a new section on permissions, specifically how to display permissions/overwrites as an array and some extra tips.


Extroonie:
I don't really think so it was needed, but you've got a typo at line 333. Also, there is no necessary use of `let` over there, since the value is never reassigned, make use of a constant variable. The library expects you to keep consistency as the contribution guidelines.

storate:
yes, i'll fix that

PR #470, opened 2020-08-29
add section about converting Permissions into flags

storate:

added new section on permissions.md about displaying permissions flags

### DIFF
--- a/guide/popular-topics/permissions.md
+++ b/guide/popular-topics/permissions.md
@@ -324,6 +324,30 @@ If you want to know how to work with the returned Permissions objects keep readi
 The <branch version="11.x" inline>[Permissions](https://discord.js.org/#/docs/main/v11/class/Permissions)</branch><branch version="12.x" inline>[Permissions](https://discord.js.org/#/docs/main/stable/class/Permissions)</branch> object is a discord.js class containing a permissions bit field and a bunch of utility methods to manipulate it easily.
 Remember that using these methods will not manipulate permissions, but create a new instance representing the changed bit field.
 
+### Displaying permissions as an array
+
+Discord.js provides a `toArray()` method, which can be used to convert a `Permissions` object of a member into an array containing permission flags. This is useful if you want to display/list them and it enables you to use other array methods. For example:
+
+```js
+const memberPermissions = member.permissions.toArray()
+const rolePermissions = role.permissions.toArray();
+//Will return something along the lines of ['SEND_MESSAGES', 'ADD_REACTIONS', 'CHANGE_NICKNAME'] ...etc...
+```
+
+The return value of `toArray()` depends entirely on where the `Permissions` object is coming from. In the examples above, we grabbed a `GuildMember`'s and a `Role`'s [base](https://discordjs.guide/popular-topics/permissions.html#base-permissions) permissions, which does not encompass any overwrites.
+
+Another possible way to get the `Permissions` object would be to access channel overwrites. For exmaple, 
+
+```js
+const overwrites = channel.permissionOverwrites
+/*Permission overwites are presented in the form of a collection, mapped by the user/role id, and the corresponding permissions object.
+Therefore, we must get/find the specific overwrite we want to access, and then specify if we want to view allowed/denied before we can use an methods on the Permissions object. */
+const deniedPermissions = overwrites.get('ROLE_OR_USER_ID').deny
+const permissionArray = deniedPermissions.toArray()
+```
+
+
+
 ### Converting permission numbers
 
 Some methods and properties in Discord.js return permission decimals rather than a Permissions object, making it hard to manipulate or read them if you don't want to use bitwise operations.
--- a/guide/popular-topics/permissions.md
+++ b/guide/popular-topics/permissions.md
@@ -326,7 +326,7 @@ Remember that using these methods will not manipulate permissions, but create a 
 
 ### Displaying permissions as an array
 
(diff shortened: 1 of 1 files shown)


tipakA:
> to convert a `Permissions` object of a member

That is not exactly correct. Not only members have permissions. Roles and channels also have them, as you even point that out later.

Also, not exactly sure what line 337 is supposed to tell. It's not exactly the resulting array that changes depending on what permissions you convert. It's the permissions that are different, and converting to array just changes how Permissions values are displayed.

storate:
how do you add commits to this pull requests

> > to convert a `Permissions` object of a member
> 
> That is not exactly correct. Not only members have permissions. Roles and channels also have them, as you even point that out later.
> 
> Also, not exactly sure what line 337 is supposed to tell. It's not exactly the resulting array that changes depending on what permissions you convert. It's the permissions that are different, and converting to array just changes how Permissions values are displayed.

oh, i see, i'll fix it up

guide/popular-topics/permissions.md
```diff
@@ -324,6 +324,34 @@ If you want to know how to work with the returned Permissions objects keep readi
 The <branch version="11.x" inline>[Permissions](https://discord.js.org/#/docs/main/v11/class/Permissions)</branch><branch version="12.x" inline>[Permissions](https://discord.js.org/#/docs/main/stable/class/Permissions)</branch> object is a discord.js class containing a permissions bit field and a bunch of utility methods to manipulate it easily.
 Remember that using these methods will not manipulate permissions, but create a new instance representing the changed bit field.
 
+### Displaying permissions as an array
+
+Discord.js provides a `toArray()` method, which can be used to convert a `Permissions` object into an array containing permission flags. This is useful if you want to display/list them and it enables you to use other array methods. For example:
```
almostSouji, line 329: - Seeing that #toArray is mostly used for debugging and pretty-printing permissions it should not be at the end of this section, rather than the beginning before even explaining the constructor
- can explain both `.toArray()` as well as `.serialize()` under the heading "Displaying permission flags" or similar

guide/popular-topics/permissions.md
```diff
@@ -324,6 +324,34 @@ If you want to know how to work with the returned Permissions objects keep readi
...
+
+```js
+const memberPermissions = member.permissions.toArray();
+const rolePermissions = role.permissions.toArray();
+//	Will return something along the lines of ['SEND_MESSAGES', 'ADD_REACTIONS', 'CHANGE_NICKNAME'] ...etc...
```
almostSouji, line 334:
```suggestion
//	output: ['SEND_MESSAGES', 'ADD_REACTIONS', 'CHANGE_NICKNAME', ...]
```

to be consistent with other pragraphs of this article

guide/popular-topics/permissions.md
```diff
@@ -324,6 +324,34 @@ If you want to know how to work with the returned Permissions objects keep readi
...
+```
+
+The meaning of the return value of `toArray()` can vary, depending on where the `Permissions` object is coming from. In the examples above, we grabbed a `GuildMember`'s and a `Role`'s [base](https://discordjs.guide/popular-topics/permissions.html#base-permissions) permissions, which is basically what a member/role can do, by default. However, you can also view things such as denied/allowed channel overwrites, and the meaning of the resulting array is different.
```
almostSouji, line 337: - The meaning does not vary, #toArray always represent the permissions present in the permission flags present in the underlying bit field of the Permissions instance it's called on. Maybe you mean the right thing, but it's worded very confusingly.
- wording: "grabbed" does not read well
- wording: "The base permissions of a GuildMember and Role"
- you can omit the last part, it's worded rather confusingly and you link to the explanation in the terminology primer

guide/popular-topics/permissions.md
```diff
@@ -324,6 +324,34 @@ If you want to know how to work with the returned Permissions objects keep readi
...
+
+
+For example,
```
almostSouji, line 340: Unnecessary  (consistency with the rest of the guide) + obsolete whitespace

guide/popular-topics/permissions.md
```diff
@@ -324,6 +324,34 @@ If you want to know how to work with the returned Permissions objects keep readi
...
+
+```js
```
almostSouji, line 342: This part of the guide should not also describe permission overwrites, focus on the permissions object and how to display its flags. If the section is short that's okay.

If you think the overwrites section should have a paragraph about allow/deny that's not in scope of this PR and should happen in the part where overwrites are actually the topic.

almostSouji (changes requested):

guide/popular-topics/permissions.md
storate: Thanks, I'll remove it

guide/popular-topics/permissions.md
storate: yup, added that

guide/popular-topics/permissions.md
```diff
@@ -324,6 +324,35 @@ If you want to know how to work with the returned Permissions objects keep readi
...
+/* output: {
+SEND_MESSAGES: true,
+ADD_REACTIONS: true,
+BAN_MEMBERS: false,
+...
+}
+*/
+```
+
+
```
almostSouji, line 355: formatting nitpick, one empty line is enough here

almostSouji (changes requested):

guide/popular-topics/permissions.md
storate: Thanks for the heads up. I've removed any unnecessary empty lines as of the most recent commit.

almostSouji (approved):

PR #478, opened 2020-09-15
fix(embeds): remove outdated embed note

vBread:
3 inline fields display regardless if a thumbnail is set

Screen Shot 2020-09-15 at 12 59 56 PM


almostSouji:
For future reference: please make a separate PR if you want to cover multiple topics (embed fields and typos in section headings of another guide section)

PR #483, opened 2020-09-24
ping documentation page

gautam-paranjape:
In this change, I created a new webpage (file), in the popular_topics folder (section). That is where my markdown is. I feel like this page should be added since I've seen many people wanting to get the ping of their bot, but have no idea despite how simple it is.

### DIFF
--- a/guide/popular-topics/bot-ping.md
+++ b/guide/popular-topics/bot-ping.md
@@ -30,4 +30,15 @@ It's pretty simple. First, you're fetching your bot (client), and then its **Web
 
 Additionally, **MS** stands for **milliseconds**, as that's what ping is measured in.
 
+And here's the result!
+<div is="discord-messages">
+	<discord-message author="User" avatar="djs">
+		!ping
+	</discord-message>
+	<discord-message author="Tutorial Bot" avatar="blue" :bot="true">
+		My ping is 100 MS!
+	</discord-message>
+</div>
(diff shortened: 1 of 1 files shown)


almostSouji (changes requested):
Good idea! This is a commonly asked for feature, indeed. As such it would fit better into the [common questions](https://discordjs.guide/popular-topics/common-questions.html) section. There it doesn't need any major explanation or mocked discord message for results.

It then can also cover both websocket heartbeat as well as API latency (full round trip). We have a tag on the server which you can use as reference.

Please also follow our [contributing guidelines](https://github.com/discordjs/guide/blob/master/CONTRIBUTING.md#linting) and lint your code to prevent the CI red you're experiencing at the moment. We wish for the codebase to be somewhat of the same style and enforce the lint rules for this.

---

Ping (heartbeat) of gateway connection(s): `<Client>.ws.ping`
*Note: A certain shards ping can be found on the WebSocketShard instance from `<Client>.ws.shards` -> `.ping`*

Using a full roundtrip to determine latency:

```js
message.channel.send('Pinging...').then(sent => {
    sent.edit(`Pong! Took ${sent.createdTimestamp - message.createdTimestamp}ms`);
});
```

gautam-paranjape:
Ah, I see. those changes make sense. I was thinking of having the API latency there too, but I thought it would be too long or something. Thanks, I'll change it!

PR #487, opened 2020-09-30
Add nsfw warning for urban dictionary command

conqr3:
Urban dictionary's API returns NSFW results. I added a warning about limiting it to NSFW channels.


advaith1:
Discord has said that they're fine with Urban Dictionary commands working in non-NSFW channels.

jxpsert:
This is pretty unneccesary, if you ask me. I think we can safely assume that people wanting to incorporate the Urban Dictionary in their bot are aware of what the website is, and thus, know what can happen.

Besides, I don't think UB has that graphic entries.

PR #493, opened 2020-10-05
Update useful-packages.md

Purukitto:
Since in Winston's case we are showing sort of the complete code example, I think it's important to the module import statement as well.

### DIFF
--- a/guide/miscellaneous/useful-packages.md
+++ b/guide/miscellaneous/useful-packages.md
@@ -160,6 +160,8 @@ But let us get a quick overview of what it can do:
 
 ```js
 const client = new Discord.Client();
+const winston = require('winston');
+
 const logger = winston.createLogger({
 	transports: [
 		new winston.transports.Console(),


Purukitto:
Also I do feel it will be helpful to add the `npm install` commands to all these packages, should I those too? Is there any specific reason why they weren't included?

monbrey:
Given that each package has a link to it's documentation I don't think its necessary for us to cover how to install them all.

People who are looking to incorporate additional third party libraries into their code should know how to install packages already.

Purukitto:
@monbrey as I said it is def not necessary but it's just adding a extra step for the user. In that light we might as well as not add the example since the documentations mentioned do already have examples as well. I just feel adding the install call would make it easier for the user to get started immediately.

almostSouji:
> I just feel adding the install call would make it easier for the user to get started immediately.

Preventing this is precisely the purpose here.

The idea of this guide section is to give the user a very very brief impulse as to which packages might be a good addition to their project, if deemed useful. Readers definitely should check out the official documentation on how to install and use the package after seeing the entry in the guide, not just blindly and mindlessly copy and paste the example use case given here.

As part of this nudge towards reading the documentation a complete "resulting code" snippet as well as the package imports and setups are deliberately not included.

PR #519, opened 2020-11-01
Fix: unknown command response

Irian3x3:
adding `unknownCommandResponse: false` on your client variable doesn't work (or at least it doesn't work for me) so I changed it to what did work.


dragonfire535 (approved):

almostSouji (changes requested):
Please [fix the lint errors](https://github.com/discordjs/guide/blob/master/CONTRIBUTING.md#linting) then this is ready to be merged.

PR #520, opened 2020-11-01
Breaking changes in awaitMessages

MicrowaveJak:
In V11, awaitMessages allowed the use of the maxMatches property. Now that the property has been removed along with `textChannel.createCollector()`, continuing to use maxMatches breaks awaitMessages entirely.

Hypothetically this information could also live in #### TextChannel#createCollector but the user would need to be aware of the relationship between createCollector() and awaitMessages()

### DIFF
--- a/guide/additional-info/changes-in-v12.md
+++ b/guide/additional-info/changes-in-v12.md
@@ -1699,6 +1699,14 @@ The `streamDispatcher.time` property has been renamed to `streamDispatcher.strea
 
 Has been removed entirely, along with all other user account-only methods and properties.
 
+#### TextChannel#awaitMessages
+
+The legacy property `maxMatches` is no longer supported in v12 in the AwaitMessagesOptions object. Please use `max` as described in the documentation.
+```diff
+- channel.awaitMessages(filterFunction, { maxMatches: 2, time: 15000 });
++ channel.awaitMessages(filterFunction, { max: 2, time: 15000 });
+```
+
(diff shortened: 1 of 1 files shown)


almostSouji:
This is covered in the respective options object section <https://discordjs.guide/additional-info/changes-in-v12.html#messagecollectoroptions>, though a link to that section from #awaitMessages as well as <https://discordjs.guide/additional-info/changes-in-v12.html#textchannel-createcollector> could be useful.

If you do link keep in mind that the anchor elements are a bit unusual due to the use of the `#` character in the heading, which usually denotes the element id to anchor to. [`#messagecollectoroptions-max-matches`](https://discordjs.guide/additional-info/changes-in-v12.html#messagecollectoroptions-max-matches) would be the correct anchor to apply in this case.

almostSouji (changes requested):
This is covered in the respective options object section <https://discordjs.guide/additional-info/changes-in-v12.html#messagecollectoroptions>, though a link to that section from #awaitMessages as well as <https://discordjs.guide/additional-info/changes-in-v12.html#textchannel-createcollector> could be useful.

If you do link keep in mind that the anchor elements are a bit unusual due to the use of the `#` character in the heading, which usually denotes the element id to anchor to. [`#messagecollectoroptions-max-matches`](https://discordjs.guide/additional-info/changes-in-v12.html#messagecollectoroptions-max-matches) would be the correct anchor to apply in this case.

PR #521, opened 2020-11-02
add what doesn't support pings in embeds

Irian3x3:
just an example of what doesn't support pings in embeds


vBread:
The [notes](https://discordjs.guide/popular-topics/embeds.html#notes) section already states where mentions will render.

Irian3x3:
> 
> 
> The [notes](https://discordjs.guide/popular-topics/embeds.html#notes) section under Embeds already states where mentions will render.

oh

PR #522, opened 2020-11-02
Update es6-syntax.md

conqr2:
Use shorthand arrow functions in the ES6 "arrow functions" section


vzwGrey:
No, this is not good.
The `_` is generally used to name values which are ignored/unused. The ready event listener does not have any parameters and as such there is also nothing to ignore. A simple no-parameter arrow function `() => {}` is perfectly fine and the correct choice in this case.

PR #524, opened 2020-11-03
Add FAQ for user blocking

### LINKED ISSUE #460: Block command
I'd love to see a command(s) to block a user from a discord bot. Or block from a specified server.

tipakA:
Adds quick answer for how to block user(s) from using a bot, including a quick tip on how to utilize database for that.

Closes #460

### DIFF
--- a/guide/popular-topics/faq.md
+++ b/guide/popular-topics/faq.md
@@ -380,7 +380,7 @@ You do not need to have a constant local variable like `blockedUsers` above. If 
 <client>.on('message', async message => {
 	const blockedUsers = await database.query('SELECT user_id FROM blocked_users;');
 	if (blockedUsers.includes(message.author.id)) return;
-})
+});
 ```
 
 Note that this is just a showcase of how such a check could be done.
--- a/guide/popular-topics/faq.md
+++ b/guide/popular-topics/faq.md
@@ -360,6 +360,32 @@ Mentions in embeds may resolve correctly in embed description and field values b
 If you want to learn more about this syntax or reaction collectors, check out [this dedicated guide page for collectors](/popular-topics/collectors.md)!
 :::
 
+### How do I block a user from using my bot?
+
+<!-- eslint-skip -->
+
+```js
+const blockedUsers = [ 'id1', 'id2' ];
+<client>.on('message', message => {
+	if (blockedUsers.includes(message.author.id)) return;
+});
+```
+
+:::tip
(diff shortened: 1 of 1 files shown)


guide/popular-topics/faq.md
```diff
@@ -360,6 +360,32 @@ Mentions in embeds may resolve correctly in embed description and field values b
...
+You do not need to have a constant local variable like `blockedUsers` above. If you have a database system which you use to store IDs of blocked users, you can query the database instead:
+
+<!-- eslint-skip -->
+
+```js
+<client>.on('message', async message => {
+	const blockedUsers = await database.query('SELECT user_id FROM blocked_users;');
+	if (blockedUsers.includes(message.author.id)) return;
+});
+```
```
pizzafox, line 384: This should use a [Set](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Set) instead of an Array. If that array gets big performance costs can start making a serious impact.

```js
const blockedUsers = new Set([ 'id1', 'id2' ]);
<client>.on('message', message => {
	if (blockedUsers.has(message.author.id)) return;
});
```

pizzafox (changes requested):
I am optimizing the Discord bot :smiley:.

guide/popular-topics/faq.md
tipakA: How big wouldn't that constant array have to be again for it to become slow

kyranet: The array needs to have at least 1 million elements for the difference to be noticeable, at which point, not only the file would be impossible to maintain due to its extreme size, but it'd also use more than twice/thrice the memory as it needs to store the hashes for each one of its value.

almostSouji: SQL injection: agree with kyra
Set instead of Array: Not a fan, literally doesn't matter in this case (hard coded list). Using different structures or approaches to determine blocked users is left as exercise to the reader. A dynamic block command inserting user ids into a persistent storage is much preferable in orders of magnitude where this becomes relevant.

tipakA: The idea of the query was to not take any parameters at all, and instead return all ids into the array. This way nothing is passed into the query at all.
True, that is relatively inefficient and can come with some weird behavior, but while i wouldn't be very surprised if someone blocked thousands if not hundreds of thousands users one by one, i also don't think it would be a very common/big issue. It's a quick FAQ afterall.

almostSouji: I somehow missed this reply, fair point! Also keeps it consistent with the example used in the static version.

PR #525, opened 2020-11-04
Some REST API guide improvements.

nceon:

I have deleted .then() in async function, because it is not needed at all and replaced with other stuff.


almostSouji:
The "then stuff" is used here to produce the json response in one logical step, the resulting list is destructured from the result `{list} = ...` which is the parsed json `await fetch()`, which is transformed to json in the then callback `.then(r => r.json())`.

Introducing another variable complicates this and makes it less readable.

PR #527, opened 2020-11-06
Update README.md

danielkickface:
If you use the code provided, the console gives an error stating message is not defined.
Message has to be defined for the code to work.


monbrey:
The text above this snippet clearly states which line from the previous example it replaces - the `console.log(message.content)` line which was already inside a message event.

The guide does not need to provide every single piece of code in a purely copy-pasteable manner.

PR #528, opened 2020-11-12
fix the old-messages example

redbrain:
This fixes the example code for using partials to detect reactions on old messages.
The USER partial is needed now that Discord's enforcing privileged intents.


NotSugden (changes requested):
you need to also note in the paragraph directly above that the `USER` partial is required

almostSouji:
Temporary measure, will be fixed with https://github.com/discordjs/discord.js/pull/4969

PR #534, opened 2020-11-21
removed section of code to help beginners

TravisC671:
there is a part right here that requires the user that sent it, needs to be the one reacting to it. as many of the projects require reacting to messages the bot sent, I think the part after the && statement is useless and confusing for beginners.
const filter = (reaction, user) => {
	return reaction.emoji.name === '👍' && user.id === message.author.id;
};

### DIFF
--- a/guide/popular-topics/collectors.md
+++ b/guide/popular-topics/collectors.md
@@ -128,7 +128,7 @@ You can read the docs for the `.createReactionCollector()` method <branch versio
 
 ```js
 const filter = (reaction, user) => {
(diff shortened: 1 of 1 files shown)


almostSouji:
> The filter will check for the 👍 emoji - in the default skin tone specifically, so be wary of that. It will also check that the person who reacted shares the same id as the author of the original message that the collector was assigned to.

The section explains in detail what it does and deliberately uses a more complex 2-part condition to also remind people what the logical AND operation looks like and how it is used to prevent various misuses of a filter and "but how do i also restrict it to..." we have seen in support.

PR #535, opened 2020-11-26
Update reactions.md

LukeGT:
I found that discord reactions are dropped if the 'USER' partial is also not enabled. The user needs to have not sent a message in a long time for the bug to trigger. 

I changed the advice in the documentation in the hopes this will help others. Happy to adjust the example code if you don't like it, I opted to parallelise the fetching of the reaction and user if required.

As an aside, this took me a long time to debug and it would be wonderful if the library emitted a debug message whenever an event is dropped.


advaith1:
> I found that discord reactions are dropped if the 'USER' partial is also not enabled. 

Sounds like you're on an old version? The `'USER'` partial is no longer needed starting with v12.5.0.

LukeGT:
You're right, this fixed my issue. Thanks for the pointer!

PR #537, opened 2020-11-27
Code correction on line 238

ToledoSDL:

I was analyzing the page and the codes, and I realized that there were two quotes that caused errors, and so I think it should be corrected.


tipakA:
That fix also has to be applied in the block for v12 branch below.

almostSouji (changes requested):
https://github.com/discordjs/guide/pull/537#issuecomment-734639197 mandatory

PR #543, opened 2020-12-08
Additional Features update

sweetchikoo2403:
Command Permissions Added

This PR makes the changes of The Additional Features by Adding Command permissions. This PR is made so that it can be added to the guide

### DIFF
--- a/guide/command-handling/adding-features.md
+++ b/guide/command-handling/adding-features.md
@@ -398,7 +398,7 @@ In the main file, You can do so by adding the following code between the Guild C
 
 ```js
 if (command.permissions && !message.member.hasPermission(command.permissions)) {
-	return message.reply(`You must have the \`${command.permissions}\` in your role or channel to use that command`);
+	return message.reply(`you must have the \`${command.permissions}\` permission to use that command`);
 }
 ```
 
--- a/guide/command-handling/adding-features.md
+++ b/guide/command-handling/adding-features.md
@@ -398,8 +398,8 @@ In the main file, You can do so by adding the following code between the Guild C
 
 ```js
 if (command.permissions && !message.member.hasPermission(command.permissions)) {
- return message.reply(`You must have the \`${command.permissions}\` in your role or channel to use that command`)
-}
+	return message.reply(`You must have the \`${command.permissions}\` in your role or channel to use that command`)
+};
 ```
 
 Now, in the `kick.js` file, add the following code:
--- a/guide/command-handling/adding-features.md
+++ b/guide/command-handling/adding-features.md
@@ -397,9 +397,9 @@ If you want members having a specific permission in the guild to use a command s
 In the main file, You can do so by adding the following code between the Guild Commands checker and the Args checker:
 
 ```js
-  if (command.permissions && !message.member.hasPermission(command.permissions)) {
-  	return message.reply(`You must have the \`${command.permissions}\` in your role or channel to use that command`)
-  }
+if (command.permissions && !message.member.hasPermission(command.permissions)) {
+ return message.reply(`You must have the \`${command.permissions}\` in your role or channel to use that command`)
+}
 ```
 
 Now, in the `kick.js` file, add the following code:
@@ -409,7 +409,7 @@ module.exports = {
 	name: 'kick',
 	description: 'Kick a user from the server.',
 	guildOnly: true,
-+	permissions: 'KICK_MEMBERS',
++	permissions: 'KICK_MEMBERS';
 ```
 
 ::: tip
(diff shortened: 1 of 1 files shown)


guide/command-handling/adding-features.md
```diff
@@ -386,6 +386,38 @@ No more manually editing your help command! If you aren't completely satisfied w
 If you want to add categories or other information to your commands you can simply add properties reflecting it to your `module.exports`. If you only want to show a subset of commands remember that `commands` is a Collection you can <branch version="11.x" inline>[filter](https://discord.js.org/#/docs/main/v11/class/Collection?scrollTo=filter)</branch><branch version="12.x" inline>[filter](https://discord.js.org/#/docs/collection/master/class/Collection?scrollTo=filter)</branch> to fit your specific needs!
 :::
 
+## Command permissions
+
+::: tip
+If you need more help regarding permissions, You can refer to the permissions article which can be found [here](https://discordjs.guide/popular-topics/permissions.html) and the extended permissions article [here](https://discordjs.guide/popular-topics/permissions-extended.html).
+:::
+
+If you want members having a specific permission in the guild to use a command such as the `KICK_MEMBERS` permission for using the kick command? You can do so by adding the following code.
```
tipakA, line 395: This line doesn't really make much sense, sounds like you forgot what you meant in the middle of a sentence.  
Additionally, line just below it has repeated `You can do so by adding the following code`.

guide/command-handling/adding-features.md
```diff
@@ -386,6 +386,38 @@ No more manually editing your help command! If you aren't completely satisfied w
...
+
+In the main file, You can do so by adding the following code between the Guild Commands checker and the Args checker:
```
tipakA, line 397: IMO this doesn't really explain much regarding where to put that. Remember that this is a list of additional but optional features.

guide/command-handling/adding-features.md
```diff
@@ -386,6 +386,38 @@ No more manually editing your help command! If you aren't completely satisfied w
...
+```diff
+module.exports = {
+	name: 'kick',
+	description: 'Kick a user from the server.',
+	guildOnly: true,
++	permissions: 'KICK_MEMBERS';
+```
+
+::: tip
+If you really want to kick the member, you can refer to the [documentation](https://discord.js.org/#/docs/main/stable/examples/moderation) for an example on kicking a member.
```
tipakA, line 416: That tip sounds quite unnecessary here. Judging from what you said above, it appears that you explain how to add it to an already existing kick command.

guide/command-handling/adding-features.md
```diff
@@ -386,6 +386,38 @@ No more manually editing your help command! If you aren't completely satisfied w
...
+:::
+
+Now that you have made the code, any member who is not having the permissions to kick the member in the guild will be prompted to have the permissions to kick the member in the guild
```
tipakA, line 419:
```suggestion
Now that you have added the check, any member who doesn't have permissions to kick members from the guild will be prompted to have the permissions to kick.
```
Still not sure about second part (`will be prompted...`) here though, could use some rewording.

guide/command-handling/adding-features.md
```diff
@@ -386,6 +386,38 @@ No more manually editing your help command! If you aren't completely satisfied w
...
+If you need more help regarding permissions, You can refer to the permissions article which can be found [here](https://discordjs.guide/popular-topics/permissions.html) and the extended permissions article [here](https://discordjs.guide/popular-topics/permissions-extended.html).
+:::
+
+If you want members having a specific permission in the guild to use a command such as the `KICK_MEMBERS` permission for using the kick command? You can do so by adding the following code.
+
+In the main file, You can do so by adding the following code between the Guild Commands checker and the Args checker:
+
+```js
+if (command.permissions && !message.member.hasPermission(command.permissions)) {
+	return message.reply(`You must have the \`${command.permissions}\` in your role or channel to use that command`);
```
tipakA, line 401: Not sure about saying `you need 'KICK_MEMBERS' in your role or channel`, not every permission can be in channel overwrites, especially the one used as an example.
```suggestion
	return message.reply(`you must have the \`${command.permissions}\` permission to use that command`);
```

tipakA (changes requested):
This should also have respective code samples added.

guide/command-handling/adding-features.md
sweetchikoo2403: it already existed here- https://discordjs.guide/command-handling/adding-features.html#guild-only-commands

guide/command-handling/adding-features.md
sweetchikoo2403: oh yes let me remove that

guide/command-handling/adding-features.md
sweetchikoo2403: ok then

almostSouji (changes requested):
This should also have respective resulting code samples added.

PR #553, opened 2020-12-22
Add error handling for missing args

ObertoIsOBS:


tipakA:
This is not necessary.

That check was not added to the resulting code, because resulting code is a full structure, utilizing args checking integrated into the command handler itself, that's why every command in the repository has `args: true` or `args: false` in the exported object.

almostSouji:
the guide previously introduces [required arguments](https://discordjs.guide/command-handling/adding-features.html#required-arguments), making this unnecessary since it's taken into consideration in the command handler to prevent code duplication

PR #560, opened 2021-01-06
refactor: move eslint stuff to yaml

vaporox:

This updates the linter guide and internal `.eslintrc` files to use `.yml` over `.js(on)` because it can be read/written more easily, which is why it's recommended for configuration files.

I know that it's purely personal preference so I'd understand if this was rejected.

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -1,4 +1,0 @@
-module.exports = {
-	extends: 'sora/vue',
-	plugins: ['markdown'],
-};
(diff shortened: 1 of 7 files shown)


Danktuary:
Although I personally don't mind the change, we'll keep the config files as is. There isn't a large benefit in changing it and JSON format is quite common for config files and many newer users tend to lean towards that over other formats. We also create a `config.json` file just a few pages later, so I'll decide keep it consistent. Thank you though!

PR #561, opened 2021-01-06
Update README.md

ThePalmerStudioGithub:


almostSouji:
Identifier names are arbitrarily (but verbosely) chosen. This is a base concept of JavaScript (and any modern programming language, really) and does not need to be explicitly stated in the guide. We do after all require a solid understanding of JavaScript.

PR #571, opened 2021-01-18
feat(Intents): replace Intents placeholder

### LINKED ISSUE #570: Intents guide
Replace the intents guide placeholder with a more verbose explanation of how intents work and what they are trying to achieve.

almostSouji:

⚠️ closes #570

When Discord stopped sending events for privileged intents and required action from developers in order to restore the full functionality of their bots the number of requests in various support channels has made it necessary to provide a straightforward manual on how to handle the situation.

In preparation for v13 (which will require websocket intents to be specified due to Discord.js using v8 of the Discord REST API and gateway) we should explain how users provide intents to the Client constructor.

This PR aims to do so. As always I'm grateful for reviews and improvement suggestions.

### DIFF
--- a/guide/popular-topics/intents.md
+++ b/guide/popular-topics/intents.md
@@ -55,14 +55,11 @@ const client = new Client({ ws: { intents: ['GUILDS', 'GUILD_MESSAGES'] } });
 
 ## The Intents Bitfield
 
-Discord.js provides a utility structure [`Intents`](https://discord.js.org/#/docs/main/stable/class/Intents) which can be utilized to easily adapt the underlying bit field. This class also has static fields for all (`Intents.ALL`), privileged (`Intents.PRIVILEGED`) and non-privileged (`Intents.NON_PRIVILEGED`) intents. You can provide these as-is or pass them to the Intents constructor to further modify to your needs.
+Discord.js provides a utility structure [`Intents`](https://discord.js.org/#/docs/main/stable/class/Intents) which you can use to easily modify bitfields. The class also features static attributes for all (`Intents.ALL`), privileged (`Intents.PRIVILEGED`), and non-privileged (`Intents.NON_PRIVILEGED`) intents.
 
-```js
-const { Client, Intents } = require('discord.js');
-const client = new Client({ ws: { intents: Intents.NON_PRIVILEGED } });
-```
+These are mostly meant to serve as templates. While using them directly is possible we strongly discourage you from using them that way. You should instead think about which events your bot strictly needs access to based on the functionality you want it to provide.
 
-You can use the `.add()` and `.remove()` methods to add or remove flags to modify the bit field. You can provide single flags as well as an array or bit field. To use a set of intents as a template you can pass it to the constructor. A few approaches are demonstrated below:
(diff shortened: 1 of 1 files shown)


guide/popular-topics/intents.md
```diff
@@ -1,97 +1,69 @@
...
 
-Intents are not available in version 11, please update to version 12 of the library if you want to use gateway intents in your bot.
+Before storming off and doing so you should stop and carefully think about if you need these events. Discord made them opt-in so users across the platform can enjoy a higher level of [privacy](https://en.wikipedia.org/wiki/Privacy_by_design). Presences can expose quite a bit of personal information through games and online times, for example. You might find it sufficient for your bot to have a little less information about all guild members at all times, considering you still get the command author as GuildMember from the command execution message and can fetch targets separately.
 
-</branch>
+### Symptoms in version 12
 
-<branch version="12.x">
+- Login times out if you try to fetch all members when starting your bot
```
SpaceEEC, line 28: Logins don't timeout on v12 anymore, you should only see a debug log that fetching members timed out (as mentioned a few lines below), and then the client should ready itself as usual:

https://github.com/discordjs/discord.js/blob/8a7abc9f06d44bf693e35a615bb6ba2c3eb1d6e7/src/client/websocket/WebSocketManager.js#L402-L415

guide/popular-topics/intents.md
```diff
@@ -100,7 +72,7 @@ myIntents.add('GUILD_PRESENCES', 'GUILD_MEMBERS');
 
 const client = new Client({ ws: { intents: myIntents } });
 
-// more examples of manipulating the bit field
+// more examples about manipulating the bitfield
 
 const otherIntents = new Intents(Intents.NON_PRIVILEGED);
 otherIntents.remove(['GUILDS', 'GUILD_MESSAGES']);
```
SpaceEEC, line 78: That's not a line you changed, but it may be good to _not_ remove GUILDS, even in an example, since that'll break everything guild-related. (read: everything but DMs)

```suggestion
otherIntents.remove(['DIRECT_MESSAGES', 'GUILD_MESSAGES']);
```
(Or pretty much any other intent)

guide/popular-topics/intents.md
```diff
@@ -1,97 +1,68 @@
...
-Discord.js provides a utility structure [`Intents`](https://discord.js.org/#/docs/main/stable/class/Intents) which can be utilized to easily adapt the underlying bit field.
+## The Intents Bitfield
 
-We also provide static fields for all, privileged and non-privileged intents. You can provide these as-is or pass them to the Intents constructor to further modify to your needs.
+Discord.js provides a utility structure [`Intents`](https://discord.js.org/#/docs/main/stable/class/Intents) which can be utilized to easily adapt the underlying bit field. This class also has static fields for all (`Intents.ALL`), privileged (`Intents.PRIVILEGED`) and non-privileged (`Intents.NON_PRIVILEGED`) intents. You can provide these as-is or pass them to the Intents constructor to further modify to your needs.
 
 ```js
 const { Client, Intents } = require('discord.js');
-const client = new Client({ ws: { intents: Intents.ALL } });
+const client = new Client({ ws: { intents: Intents.NON_PRIVILEGED } });
```
advaith1, line 62: I wouldn't include this because it goes against the purpose of intents and should not be used, but if this is here users will probably just copy it

almostSouji: Removed the entire section in aad8a10 and clarified that the static attributes are meant to be used as templates rather than as-is.

PR #574, opened 2021-01-21
Didn't thought you would forget adding __dirname

sayantan300:
Added `__dirname` for fs.readdirSync().... I just wasted 1 full hour on finding the problem.... Anyways, add this same thing to v11 folder and other command handler folders... Hope this fixes bugs for everyone. 😏

### DIFF
--- a/code-samples/command-handling/adding-features/12/index.js
+++ b/code-samples/command-handling/adding-features/12/index.js
@@ -5,7 +5,7 @@ const { prefix, token } = require('./config.json');
 const client = new Discord.Client();
 client.commands = new Discord.Collection();
 
-const commandFiles = fs.readdirSync('./commands').filter(file => file.endsWith('.js'));
+const commandFiles = fs.readdirSync(__dirname+'./commands').filter(file => file.endsWith('.js'));
 
 for (const file of commandFiles) {
 	const command = require(`./commands/${file}`);


iShibi:
The guide assumes that your `index.js` and `commands` folder are on the same level, so that shouldn't be an issue.

Fyko:
> Anyways, add this same thing to v11  

v11 is no longer supported

iShibi:
Also, `__dirname` will resolve the path with `\` in it which will clash with the `/` in `./commands`.

sayantan300:
> The guide assumes that your `index.js` and `commands` folder are on the same level, so that shouldn't be an issue.

I know they should be at the same level, but adding `__dirname` will give the directory name the file is in.
If you don't add __dirname, fs will check for commands folder in the root directory, probably what's called $HOME directory, instead of current directory. More information here: [link](https://www.geeksforgeeks.org/node-js-fs-readdirsync-method/amp/)
This site shows using __dirname to goto the current folder where commands folder is there.

sayantan300:
You can just clone this repo for example, and come to this v12 folder, fill in config.json, and run the index.js file, you can see there's a error from fs, saying can't find "./commands/" folder... If you want to fix it, you just need to add __dirname+"./commands/" in your index.js code.

tipakA:
> You can just clone this repo for example, and come to this v12 folder, fill in config.json, and run the index.js file, you can see there's a error from fs, saying can't find "./commands/" folder... 

And that's where you're ~~wrong~~ not entirely correct.
if you `cd` to `guide\code-samples\command-handling\adding-features\12`, make a `package.json` and install `discord.js`, and try to run the `index.js` file that is there, since it passes `./commands` it resolves correctly, because commands dir is literally next to the index file.

Issue starts when you try to run that file not with `node .` or `node index.js`, but for example `node ./12/index.js`, or from any other directory. Then indeed, the `./` in `./commands` points to current directory which is not the one with `index.js`, making the `__dirname` a valid addition.
<hr>

However, the way you have added the `__dirname` is incorrect. Your `__dirname+'./commands'` results in<br>`adding-features\12./commands`<br>instead of<br>`adding-features\12/commands` (There seem to be no difference between `/` and `\` for path for `fs.readdirSync`, at least for node 14.15).

Apart from that, it uses string concatenation without any linting instead of template literals.
Oh, and this PR only edits the resulting code part, and does not update the actual gudie content at all. Also worth noting that the `__dirname` itself is not used anywhere in the guide, so it might be useful to explain it as well.

sayantan300:
> > You can just clone this repo for example, and come to this v12 folder, fill in config.json, and run the index.js file, you can see there's a error from fs, saying can't find "./commands/" folder...
> 
> And that's where you're ~wrong~ not entirely correct.
> if you `cd` to `guide\code-samples\command-handling\adding-features\12`, make a `package.json` and install `discord.js`, and try to run the `index.js` file that is there, since it passes `./commands` it resolves correctly, because commands dir is literally next to the index file.
> 
> Issue starts when you try to run that file not with `node .` or `node index.js`, but for example `node ./12/index.js`, or from any other directory. Then indeed, the `./` in `./commands` points to current directory which is not the one with `index.js`, making the `__dirname` a valid addition.
> 
> However, the way you have added the `__dirname` is incorrect. Your `__dirname+'./commands'` results in
> `adding-features\12./commands`
> instead of
> `adding-features\12/commands` (There seem to be no difference between `/` and `\` for path for `fs.readdirSync`, at least for node 14.15).
> 
> Apart from that, it uses string concatenation without any linting instead of template literals.
> image
> 
> 
> Oh, and this PR only edits the resulting code part, and does not update the actual gudie content at all. Also worth noting that the `__dirname` itself is not used anywhere in the guide, so it might be useful to explain it as well.

Oh yes, adding package.json will treat fs as if 12 is the main folder... Anyways, using __dirname is good if we have our index.js in inner directory. Also, yeah, sorry for not removing "./" Stuff... I hope you all can do it. Since I was watching in GitHub web, I thought of just making this pr to inform you about it. Currently I'm working on a different project so I'm not wasting my time for updating the full guide about it. When my work is over, I shall contribute towards to docs 😁

iShibi:
Yup just checked, the `__dirname` inside `fs.readdirSync` also works with a path that is made up of both `/` and `\`. I got the error due to the `.` in the `./commands`. So that's my fault, should have caught your syntax mistake.

Anyway, the guide is written for new readers and this will be a superfluous change, that will make new readers hard to follow the guide. As the guide already assumes `index.js` and `commands` are on the same level, keeping it as `./commands` gets the work done and is easier for the readers to understand.

sayantan300:
You are saying wrong... Even if index.js is in the same folder where commands folder is, it will not work unless you add package.json (or npm init -y) to make that current folder as the main folder.

iShibi:
Creating a config file like `package.json` is a necessary step and is already covered in the guide before this point. So, I don't understand how the reader will get the error that you're describing.

sayantan300:
Also, you all are not teaching some async stuff in a correct way.. like the example for fetching urban dictionary and making it as an embed.. that's fully async work, and you can only run it if you have 
```js
(async () => {
//Put code inside here
});
```
Moreover, __dirname should be learnt by new readers because that's essential part of learning fs.

sayantan300:
> Creating a config file like `package.json` is a necessary step and is already covered in the guide before this point. So, I don't understand how the reader will get the error that you're describing.

Oh.  But adding __dirname won't do anything wrong...

iShibi:
Yes, It won't. But to introduce that, we will have to make changes in a lot of places in the guide. The guide's goal (at least what I believe personally) is to teach how to make a discord bot using this library as easily as possible. Introducing `__dirname` when `./` is already there (and working without error, if right steps are followed) will just be an unnecessary addition to the guide and readers education of the library.

sayantan300:
Ok then, I'm closing this pr 😶

tipakA:
The `package.json` has absolutely nothing to do with what is current directory. *Only* reason i created it was because i needed to install discord.js. If the `index.js` wouldn't need any external packages, `package.json` wouldn't be necessary.
The current directory will be current directory no matter what is inside, so the "issue" of executing main file from other directory than the file is in interfering with `./` "remains", but is avoided in guide (and code samples) by placing `commands` directory next to `index.js`.

PR #588, opened 2021-01-23
feat: add section for event-handler

iShibi:

Feature requested in: #586
This PR adds a new section that teaches readers how to make an event handler.


iShibi:
As discussed in the server, the PR doesn't show how to get hold of `client` in `ready.js` yet. Unlike other events `ready` doesn't return an object, so it's not possible to get it from there. We could fix this problem by passing the `client` along with `args` in the event handler like this:
```js
client.on(eventName, (...args) => event(client, ...args));
```
Now, if this is what we decide to go with, I would propose that we don't edit this into the section where we introduce the event handler, as doing so would make the already complicated section more complicated. I think we can add this as an update to the event handler in a section after the `Message Event` section.

tipakA:
Note that prepending event args with `client` isn't "hot-swappable", so if someone would follow "base" and have events being called without passing `client` separately, they won't be able to just go ahead and apply that change in handler just for `ready` without being forced to update every other event

iShibi:
Yes, that's a valid point. I was aware of it and was planning to teach readers how to adapt their event files to this change in this new section. But on second thought, the above style does force the reader to always take `client` as a parameter in every event file they create, even when they don't need it. So, I think this would be a better version of the above snippet:
```js
client.on(eventName, (...args) => event(...args, client));
```
This way they can take the arguments returned by the particular event first and only take the `client` in the even file when they need it.

tipakA:
Sounds good. Maybe could throw in a tip / warning about the fact that `on('messageUpdate', (newMessage, client) =>{})` is not going to be valid 👍

guide/event-handling/README.md
```diff
@@ -0,0 +1,237 @@
...
+
+So far you have been listening to these events in your main `index.js` file. This is fine for small bots that do only a few things, but if you want your bot to do a lot of things, you'll have to listen for more than just the above two events. If you went to the documentation site, you would have seen that there are a lot of events that the bot can listen for. Writing code for these events in the `index.js` file will make this file bigger and messy. This is bad for project maintenance. Therefore, you'll have to take your code for each event from `index.js` and keep them in their own separate file inside an `events` folder. This is somewhat similar to what you did with commands while making the Command Handler.
```
Danktuary, line 5:
```suggestion
Node.js uses an event-driven architecture. This makes it possible to execute code when a certain event occurs. The discord.js library takes full advantage of this. You can visit the [discord.js documentation site](https://discord.js.org/#/docs/main/stable/class/Client) to see the full list of `Client` events.

Here's the base code we'll be using:

\```js
const Discord = require('discord.js');
const { prefix, token } = require('./config.json');

const client = new Discord.Client();

client.once('ready', () => {
	console.log(`Ready! Logged in as ${client.user.tag}`);
});

client.on('message', message => {
	console.log(`${message.author.tag} in #${message.channel.name} sent: ${message.content}`);
});

client.login(token);
\```

Currently, the event listeners are placed in the `index.js` file. The `ready` event emits once when the `Client` becomes ready for use and the `message` event emits whenever a message is received. Moving the event listener code into individual files is simple and we'll be taking a similar approach to the [command handler](/command-handling/). 
```
Remove the `\` before the 3 ticks from my suggestion

guide/event-handling/README.md
```diff
@@ -0,0 +1,237 @@
...
+
+## Individual event files
+
+If you have followed along so far, your entire project structure should look something like this:
```
Danktuary, line 9:
```suggestion
Your project structure should look something like this:
```

guide/event-handling/README.md
```diff
@@ -0,0 +1,237 @@
...
+
+![Project structure before events folder](~@/images/project-structure-before-events-folder.png)
+
+Create a new `events` folder in the directory where your `index.js` file is kept. This is how your project structure should look like after that:
```
Danktuary, line 13:
```suggestion
Create an `events` folder in the directory where your `index.js` file is kept. This is how your project structure should look after that:
```

Danktuary (changes requested):
- Since this is a separate section, the commands from the command handling page shouldn't be included in code samples or guide page. This page can be kept similar, with a brief introduction and some base code.
  - The code samples version folder should be renamed from `v12` to `12`

- The individual event files should export an object similar to the ones in the command handler, with the properties: `name`, `type`, and `execute`. We use `client.once('ready', ...)` on the other pages but as is, it would make all events use `.on()`.

- `path.basename()` should be used instead of `.slice()`. And since `path` would be getting included, I suppose using `path.join()` for `fs.readdirSync()` arguments would be more ideal?

- The images should get updated and moved to `/guide/event-handling/images/*.png`

iShibi:
> The individual event files should export an object similar to the ones in the command handler, with the properties: `name`, `type`, and `execute`. We use `client.once('ready', ...)` on the other pages but as is, it would make all events use `.on()`.

I understand that you want to keep it like the command-handler but do we really need these properties? The current approach seems a lot easier, where we are just exporting an arrow function and calling it in the event-handler. Also, `ready` event only emits once, so it won't make any difference even if we use `.on` for it too. Yes, it's not in-line with what we were doing previously in the guide, to address that we can put a little tip in there.

> `path.basename()` should be used instead of `.slice()`. And since `path` would be getting included, I suppose using `path.join()` for `fs.readdirSync()` arguments would be more ideal?

Again, I understand, but I was hoping that we keep things a little simple. I've been testing and using this event-handler for quite some time now and it has been working without any issue. Using `path` will just make things a little complicated in my opinion. Why use `path` when we can do the same thing without it?

iShibi:
As a result of the chat we just had on the server: I'm unaware of the fact that how `ready` event can trigger twice in the same session but just for keeping it safe and in-line with the guide, we can make a check for the `ready` eventName in the event handler and register its arrow function using `.once`. Would that be okay?

Danktuary:
I'd very much prefer to keep it like the command handler. While this approach is simpler, it is not consistent, and consistency should be preferred. It's not much harder to understand, and those who have read the command handling page will be able to adapt this easily.

The `name` property should be used instead of the file name, and the `type` property can default to `on` (i.e., `client[event.type || 'on'](...)`) so that the user doesn't always have to provide it. Using the `name` property would also mean that using `.slice()` and `path.basename()` aren't necessary.

Personally, I try to use `path.join()` instead of relative path strings to avoid confusion with folders being relevant to the current file directory or current working directory. Because, for example, having `readdirSync('./commands/)` inside `/commands/reload.js` doesn't look right at a glance imo. In retrospect, I should've used `path` for the command handler as well.
For the time being, this is okay, but I may refactor this at a later date.

iShibi:
Got it. I'll have to rewrite a few stuff to accommodate this change. Just wanna conform a few things before doing that:

- Event-handler:
```js
const eventFiles = fs.readdirSync('./events').filter(file => file.endsWith('.js'));

for (const file of eventFiles) {
  const event = require(`./events/${file}`);
  if (event.type === 'once') {
    client.once(event.name, () => event.execute(client));
  } else {
    client.on(event.name, (...args) => event.execute(...args, client));
  }
}
```
- Module for `ready` event:
```js
module.exports = {
  name: 'ready',
  type: 'once',
  execute(client) {
    console.log(`Logged in as ${client.user.tag}`);
  },
};
```
Anything, that should be changed in it? Also, now we are passing the `client` to the `ready` module separately, do we still need to pass it as an extra argument to other modules too?

Danktuary:
Actually I realized that using a `once: true` property would make a bit more sense. 😅 Sorry for all the confusion.
The `event.execute(...args, client)` line should be the same for both.

guide/event-handling/README.md
```diff
@@ -0,0 +1,284 @@
...
+if (event.once) {
+	client.once(event.name, (...args) => event.execute(...args));
+} else {
+	client.on(event.name, (...args) => event.execute(...args));
+}
+```
+
+You can listen for an event either only once or multiple times by registering it using `once` or `on` respectively. In both of the above conditions, you are registering a callback function that will be called when its respective event emits, but for events that have their `once` property set to `true`, you will register them using the `once` method, so that it gets unregistered after a single event.
+
+Before you go any further you should check the event handler is working or not. Open your terminal, type `node index.js` and press enter. If you see `Ready!` gets printed, then congratulations, you've made an event handler. After this, listening for other events is as easy as creating a new event file in the `events` folder. The event handler will automatically retrieve this new event file and register the respective event whenever you restart the bot.
```
Danktuary, line 139:
```suggestion
In order to listen for events, you have to register them first. This is done by using the `on` or `once` methods of an `EventEmitter` instance. The `on` method for events that can emit multiple times, while `once` will run once and unregister the listener after a single emit.

::: tip
You can learn more about `EventEmitter` [here](https://nodejs.org/api/events.html#events_class_eventemitter).
:::

The `Client` class in discord.js extends the `EventEmitter` class. Therefore, the `client` object also has these `on` and `once` methods that you can use to register events. These methods take two arguments: name of the event and a callback function.

The callback function passed takes argument(s) returned by its respective event, collects them in an `args` array using the `...` [rest parameter syntax](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/rest_parameters), then calls `event.execute` function while passing in the `args` array using the `...` [spread syntax](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Spread_syntax). They are used here because different events in discord.js have different number of arguments. The rest parameter collects these variable number of arguments into a single array, and the spread syntax then takes these elements and passes it to the `execute` function.

After this, listening for other events is as easy as creating a new file in the `events` folder. The event handler will automatically retrieve and register it whenever you restart your bot.
```

guide/event-handling/README.md
```diff
@@ -0,0 +1,284 @@
...
+Here's the base code we'll be using:
+
+```js
+const Discord = require('discord.js');
+const { prefix, token } = require('./config.json');
+
+const client = new Discord.Client();
+
+client.once('ready', () => {
+	console.log(`Ready! Logged in as ${client.user.tag}`);
```
Danktuary, line 14:
```suggestion
	console.log('Ready!');
```

guide/event-handling/README.md
```diff
@@ -0,0 +1,284 @@
...
+Your next step would be to define the `client` object. In the `index.js` file, you defined it by calling its constructor. What you have to do here is to get hold of that `client` so that you can use it in this file too. Most of the structures in discord.js have the `client` as their property. This makes it possible to get hold of the client through these structures. The `message` object returned by the `message` event is one of those structures. Therefore you can define the `client` by adding the following line after checking for the `prefix`:
+
+```diff
+if (!message.content.startsWith(prefix) || message.author.bot) return;
++ const client = message.client;
+```
+
+Now run your bot and check whether it's working or not. If you encounter any error, go back and check you haven't made any mistake.
+
+You can add back all the additional features in the `message.js` event file. Just make sure you take care of the things that we talked about above.
```
Danktuary, line 191: The `message.js` code should be kept as a single `console.log()` and not resemble the command handling code. As per one of my other suggestions, the message.js code was moved up into the previous section, essentially making this section unneeded.

guide/event-handling/README.md
```diff
@@ -0,0 +1,284 @@
...
+
+## Passing client to event files
```
Danktuary, line 193:
```suggestion
## Passing `Client` to event files
```

guide/event-handling/README.md
```diff
@@ -0,0 +1,284 @@
...
+
+You must have noticed how important `Client` class is in discord.js. You created a `client` instance of this class in the `index.js` file. Most of the time you can use this `client` in other files by either obtaining it from one of the `discord.js` structures or from function parameters. You did the former in the above section by using `<Message>.client`. When you don't have access to any of the structures that have `client` as one of their properties, you'll have to use the latter method. The perfect example of this is the `ready.js` event file.
+
+The `ready` event does not return anything, this means that `args` will be an empty array, thus nothing will be passed to the `execute` function in the `ready.js` event file. Now to obtain the `client` you'll have to pass it as an argument along with the `args` array in the event-handler. So, head back to the `index.js` file and make the following changes in it:
```
Danktuary, line 197:
```suggestion
You may have noticed how important the `Client` class is. You created a `client` instance of this class in the `index.js` file. Most of the time you can use this `client` instance in other files by either obtaining it from one of the other discord.js structures or from function parameters. The above section does the former by using `message.client`. When you don't have access to any of the structures that have the `client` property, you'll have to use the latter method. The prime example of this is the `ready` event.

The `ready` event does not have arguments, meaning that `args` will be an empty array, thus nothing will be passed to the `execute` function in `ready.js`. To obtain the `client` instance, you'll have to pass it as an argument along with the `args` array in the event handler. Back in `index.js`, make the following changes:
```

guide/event-handling/README.md
```diff
@@ -0,0 +1,284 @@
...
+
+Now, `client` will be available as the last argument to the `execute` function in every event file you'll create. Since, `ready` event doesn't return anything, `client` will be the only argument to the `execute` function of `ready.js` event file. You can make use of `client` in `ready.js` by logging your bot's name in the console when it becomes ready. Head back to `ready.js` and make the following changes to it:
+
+```diff
+module.exports = {
+	name: 'ready',
+    once: true,
++	execute(client) {
+-		console.log('Ready!');
++		console.log(`Logged in as ${client.user.tag}`);
```
Danktuary, line 220:
```suggestion
+		console.log(`Ready! Logged in as ${client.user.tag}`);
```

guide/event-handling/README.md
```diff
@@ -0,0 +1,284 @@
...
+-		client.once(event.name, (...args) => event.execute(...args));
++		client.once(event.name, (...args) => event.execute(...args, client));
+	} else {
+-		client.on(event.name, (...args) => event.execute(...args));
++		client.on(event.name, (...args) => event.execute(...args, client));
+	}
+}
+```
+
+Now, `client` will be available as the last argument to the `execute` function in every event file you'll create. Since, `ready` event doesn't return anything, `client` will be the only argument to the `execute` function of `ready.js` event file. You can make use of `client` in `ready.js` by logging your bot's name in the console when it becomes ready. Head back to `ready.js` and make the following changes to it:
```
Danktuary, line 212:
```suggestion
This allows `client` to available as the **last** argument to the `execute` function in every event file you'll create. You can make use of `client` in `ready.js` by logging your bot's tag in the console when it becomes ready:
```

guide/event-handling/README.md
```diff
@@ -0,0 +1,284 @@
...
+module.exports = {
+	name: 'message',
+	once: false,
++	execute(message, client) {
+		if (!message.content.startsWith(prefix) || message.author.bot) return;
+-		const client = message.client;
+		// ...
+	},
+};
+```
```
Danktuary, line 237:
```suggestion
You can omit the `client` argument from the `execute` function in files where you don't need it. For example, it isn't needed in the `message.js` file because its first argument is a `Message` instance, meaning you can use `message.client`.
```

guide/event-handling/README.md
```diff
@@ -0,0 +1,284 @@
...
+```js
+// This will create problems :(
+module.exports = {
+	name: 'messageUpdate',
+	once: false,
+	execute(newMessage, client) {
+		// ...
+	},
+};
+```
```
Danktuary, line 280:
```suggestion
It is worth noting that the position of `client` argument matters. For example, the `messageUpdate` event has two arguments: `oldMessage` and `newMessage`. Your event file for this event would look something like this:

```js
module.exports = {
	name: 'messageUpdate',
	execute(oldMessage, newMessage, client) {
		// ...
	},
};
```

If you were to try `execute(newMessage, client)`, this would mean that `newMessage` is an old `Message` object and `client` is a new `Message` object.
```

**Review by Danktuary — CHANGES_REQUESTED** [review]
- The message event file doesn't need `once: false`
- Instances of "event-handler" should be "event handler"
- Since the code samples for this should work on both v11 and v12, `event-handling/file-setup/12/*` can be moved up to `event-handling/file-setup/*`.

File: guide/event-handling/README.md (hunk 6 above)
Comment (iShibi) [reply guide/event-handling/README.md]: Removed the message event section :) 👍

File: guide/event-handling/README.md [hunk 13]
```diff
@@ -0,0 +1,144 @@
...
+
+Now, you'll write the code for dynamically retrieving all the event files in the `events` folder. Add this below the `const client` line in `index.js`:
+
+```diff
+const client = new Discord.Client();
+
++ const eventFiles = fs.readdirSync('./events').filter(file => file.endsWith('.js'));
+```
+
+You have already learned about this in the [command handler](/command-handling/) section. The `fs.readdirSync()` method returns an array of all the file names in a given directory. Here, the directory is `events`, therefore if you log `eventFiles` it will print: `[ 'ready.js' ]`. Since you only need JavaScript files, you filtered out all the files that do not end with `.js` extension. Currently, you have a single file in the `events` folder but you'll create more once you have made the necessary changes in `index.js`.
```
Danktuary, line 67:
```suggestion
This same method is used in our [command handler](/command-handling/) section. The `fs.readdirSync().filter()` calls return an array of all the file names in the given directory and filter for only `.js` files, i.e. `['ready.js', 'message.js']`.
```

Danktuary (changes requested):

guide/event-handling/README.md
```diff
@@ -0,0 +1,144 @@
...
+	name: 'ready',
+    once: true,
++	execute(client) {
+-		console.log('Ready!');
++		console.log(`Ready! Logged in as ${client.user.tag}`);
+	},
+};
+```
+
+You can omit the `client` argument from the `execute` function in files where you don't need it. For example, it isn't needed in the `message.js` file because its first argument is a `Message` instance, meaning you can use `message.client`.
```
Danktuary, line 127:
```suggestion
::: tip
You can omit the `client` argument from the `execute` function in files where you don't need it. For example, it isn't needed in the `message.js` file because its first argument is a `Message` instance, meaning you can use `message.client`.
:::
```

guide/event-handling/README.md
```diff
@@ -0,0 +1,144 @@
...
+
+Your project structure should look something like this:
+
+![Project structure before events folder](./images/project-structure-before-events-folder.png)
+
+Create an `events` folder in the directory where your `index.js` file is kept. This is how your project structure should look after that:
+
+![Project structure after events folder](./images/project-structure-after-events-folder.png)
+
+Now, you'll take your existing events code in `index.js` and move them to individual files inside the `events` folders. Create a `ready.js` and a `message.js` file in the `events` folder and place in the code for the respective files:
```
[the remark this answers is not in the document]
iShibi: That image is missing the `commands` folder (not important here, but it will be consistent with the previous section) and it also looks a little different from these two.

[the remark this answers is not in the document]
Danktuary: That's the idea; there is no commands folder in this guide section and the image used on this page should look the same as the existing one.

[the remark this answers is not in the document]
iShibi: Yeah, I realized that after posting that comment. Was on my way to commit the changes. 😓

Danktuary (approved):

PR #594, opened 2021-01-31
Rework of missing privileged intents symptoms as table

GeniusTimo:
I read the revised version of you on Gateway Intents and thought it was very good. I still have the suggestion to separate the effects of the individual privileged intents and format them as a list or table. Here are both options, currently the table is my choosen format in the PR.

> Table
> 
> Lightmode
> 
> Darkmode

> List
> 
> Lightmode
> 
> Darkmode

### DIFF
--- a/guide/popular-topics/intents.md
+++ b/guide/popular-topics/intents.md
@@ -25,11 +25,15 @@ Before storming off and doing so you should stop and carefully think about if yo
 
 ### Symptoms in version 12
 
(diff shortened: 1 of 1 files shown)


Danktuary:
The table display is nice but as each gets longer or as more items get added, the table contents will get tighter and more awkward to read, especially on mobile devices. I'd go with the list format.

almostSouji (changes requested):
After consideration, I fully agree with sanctuary here. List it is.

almostSouji (approved):

PR #600, opened 2021-02-18
update Guild Only Commands

FoXZillaPT17:
The description to add main file its confused, im add:

```js
    if (command.args && !args.length) {  let reply = `You didn't provide any arguments, ${message.author}!`;  if (command.usage) { reply = `\nThe proper usage would be: \`${prefix}${command.name} ${command.usage}\``;  } return message.channel.send(reply); 	  } 
```


tipakA:
What is this supposed to be? This code is already mentioned above, definitely in way more readable way.

PR #610, opened 2021-02-26
Update setting-up-a-linter.md

thepwrtank18:
Added WebStorm to the ide list


Danktuary:
I think the current options are more than enough, especially since VS Code has everything you'd need for a Discord bot project, in either JS or TS. Thanks though. 😄

PR #612, opened 2021-03-06
Update webhooks.md | console.log() on webhook create/edit, logs an object

xAtsuUC:
Replaced ${webhook} with ${webhook.name}

The change implemented in this request was, fixing that, when you create or edit a webhook following the guide, it logs `Created webhook [object Object]`, and with this PR, will log `Created webhook ${webhook.name}`


guidojw:
A nicer way to resolve this problem (imo) could be overriding the toString method in the Webhook structure on discord.js. 
Kind of how it's done in the [Guild structure](https://github.com/discordjs/discord.js/blob/d744e51c1bdb4c7a26c0faeea1f2f45baaf5fd3c/src/structures/Guild.js#L1433) as well.

Issue #615, opened 2021-03-07
Event handling guide should make an edit that puts cooldowns into client.cooldowns


JavascriptLearner815:
**Is your feature request related to a problem? Please describe.**
Without this part, developers will have to come up with this solution on their own.

**Describe the solution you'd like**
In the Event handling guide, make a diff that edits `const cooldowns` to be `client.cooldowns`.

**Describe alternatives you've considered**
- When cooldowns are first introduced, use `client.cooldowns` immediately instead of `const cooldowns` instead of a change in the Event handling guide.
- Instead of defining it in the client, check if the event name is "message", then also pass `cooldowns`.

**Additional notes**
- The main solution is proved to work.
- The "message" event handler requires `cooldowns` if you follow the Additional features guide. Thus, `cooldowns` should be defined in the client for use via `message.client` and/or by the already-aforementioned `Client` passed into the event handlers.
- I've had to think up this solution myself for one of my own bots.

PR #616, opened 2021-03-08
feat(command-handling): Make cooldowns a property of client

iShibi:

fixes: #615
This PR makes the `cooldowns` collection a property of `client`. Thus, making it possible to access the `cooldowns` in the `message.js` file later in the `event-handler` section. This change will somewhat reduce the hassle of adding back additional command-handling features after a reader has implemented the event-handler.

### DIFF
--- a/code-samples/command-handling/adding-features/11/index.js
+++ b/code-samples/command-handling/adding-features/11/index.js
@@ -4,6 +4,7 @@ const { prefix, token } = require('./config.json');
 
 const client = new Discord.Client();
 client.commands = new Discord.Collection();
+client.cooldowns = new Discord.Collection();
 
 const commandFolders = fs.readdirSync('./commands');
 
@@ -14,8 +15,6 @@ for (const folder of commandFolders) {
 		client.commands.set(command.name, command);
 	}
 }
-
-const cooldowns = new Discord.Collection();
 
 client.once('ready', () => {
 	console.log('Ready!');
@@ -52,6 +51,8 @@ client.on('message', message => {
 
 		return message.channel.send(reply);
 	}
+
+	const cooldowns = client.cooldowns;
 
 	if (!cooldowns.has(command.name)) {
 		cooldowns.set(command.name, new Discord.Collection());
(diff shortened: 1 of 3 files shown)


code-samples/command-handling/adding-features/11/index.js
```diff
@@ -53,6 +52,8 @@ client.on('message', message => {
 		return message.channel.send(reply);
 	}
 
+	const { cooldowns } = client;
```
Danktuary, line 55:
```suggestion
const { cooldowns } = client;
```

code-samples/command-handling/adding-features/12/index.js
```diff
@@ -53,6 +52,8 @@ client.on('message', message => {
 		return message.channel.send(reply);
 	}
 
+	const { cooldowns } = client;
```
Danktuary, line 55:
```suggestion
const { cooldowns } = client;
```

guide/command-handling/adding-features.md
```diff
@@ -203,14 +203,16 @@ module.exports = {
...
 ```
 
-This initializes an empty Collection (remember, Collection is a utility data structure, which works based on key/value pairs) which you can then fill later when commands are used. The key will be the command name and the value will be another Collection associating the user id (key) to the last time (value) this specific user used this specific command. Overall the logical path to get a specific user's last usage of a specific command will be `cooldowns > command > user > timestamp`.
+This initializes an empty Collection (remember, Collection is a utility data structure, which works based on key/value pairs) as a property of the `client` object, which you can then fill later when commands are used. The key will be the command name and the value will be another Collection associating the user id (key) to the last time (value) this specific user used this specific command. Overall the logical path to get a specific user's last usage of a specific command will be `cooldowns > command > user > timestamp`.
 
 In your main file, directly above the `try/catch` block causing command execution, add in the following:
 
 ```js
+	const { cooldowns } = client;
```
Danktuary, line 214:
```suggestion
const { cooldowns } = client;
```

PR #622, opened 2021-03-20
corrected a small typo

mushahidq:


tipakA:
This is invalid. `false` is `false`, but it's only one of few falsy values node has. `0`, `NaN`, empty string, `false` are all examples of falsy values which are *not* `false`.

mushahidq:
Oh sorry. I didn't know that

Issue #596, opened 2021-02-08
Commando install instructions outdated


redbrain:
It seems like `discord.js-commando` no longer needs to be installed from Git for discord.js v12.
Should the docs be changed?

PR #625, opened 2021-03-22
docs(Commando): Update installation instructions

iShibi:

Closes: #596

### DIFF
--- a/guide/commando/README.md
+++ b/guide/commando/README.md
@@ -4,13 +4,10 @@ pageTheme: blue
 
 # Getting started with Commando
 
-When you got your first bot up and running with Discord.js, you should've installed Discord.js using npm, Node.js' Package Manager. The same applies to Commando, which must be separately installed. You can do this in one of two ways:
-
-If using Discord.js v11: `npm install discord.js-commando`  
-If using Discord.js v12: `npm install discordjs/Commando`
+When you got your first bot up and running with Discord.js, you should've installed Discord.js using npm, Node.js' Package Manager. The same applies to Commando, which must be separately installed. To install Commando, run `npm install discord.js-commando` in the terminal.
 
 ::: warning
-You need at least Node.js version <branch version="11.x" inline>8.0.0</branch><branch version="12.x" inline>12.0</branch> to use Commando. The main branch will also require you to install [Git](https://git-scm.com/downloads).
+You need at least Node.js version <branch version="11.x" inline>8.0.0</branch><branch version="12.x" inline>12.0</branch> to use Commando.
 :::
 
 ## Creating your index.js file


guide/commando/README.md
```diff
@@ -4,13 +4,10 @@ pageTheme: blue
...
 
-When you got your first bot up and running with Discord.js, you should've installed Discord.js using npm, Node.js' Package Manager. The same applies to Commando, which must be separately installed. You can do this in one of two ways:
-
-If using Discord.js v11: `npm install discord.js-commando`  
-If using Discord.js v12: `npm install discordjs/Commando`
+When you got your first bot up and running with Discord.js, you should've installed Discord.js using npm, Node.js' Package Manager. The same applies to Commando, which must be separately installed. To install Commando, run `npm install discord.js-commando` in the terminal.
 
 ::: warning
-You need at least Node.js version <branch version="11.x" inline>8.0.0</branch><branch version="12.x" inline>12.0</branch> to use Commando. The main branch will also require you to install [Git](https://git-scm.com/downloads).
+You need at least Node.js version <branch version="11.x" inline>8.0.0</branch><branch version="12.x" inline>12.0</branch> to use Commando.
```
kyranet, line 10: `11.x` is not longer supported in Commando, so there's no point on keeping the branch.
```suggestion
You need at least Node.js version 12.0 to use Commando.
```

kyranet: I also thought of doing this instead:
```suggestion
You need <branch version="11.x" inline>to upgrade discord.js to 12.x and </branch>at least Node.js version 12.0 to use Commando.
```

So when it's in 11.x, it shows:

```markdown
:::warning
You need to upgrade discord.js to 12.x and at least Node.js version 12.0 to use Commando.
:::
```

And when in 12.x:

```markdown
:::warning
You need at least Node.js version 12.0 to use Commando.
:::
```

kyranet (changes requested):

guide/commando/README.md
iShibi: There are a lot of places in the guide where this [condition](https://discordjs.guide/preparations/#installing-on-linux) arises. Wouldn't it be inconsistent to add this change only here? I feel the popup is enough to tell the user that v11 isn't supported anymore.

kyranet: I honestly don't know. cc: @discordjs/guides

almostSouji: There's still a djs-v11 branch for commando, or am I missing something?
We generally recommend updating to v12, however, keeping the v11 guide intact as far as possible is preferable.

iShibi: I assume this is resolved.

almostSouji: Well, __if__ that branch actually works with v11, the install instructions still differ and I suppose v11 requires git then. Should that not be the case and commando not function with v11 whatsoever the commando guide itself would need the branching tags and display that fact for 11

dragonfire535 (approved):

kyranet (approved):

PR #551, opened 2020-12-17
feat: v13 guide

monbrey:
This is by no means a complete guide on upgrading from v12 to v13, but the page needs to start somewhere.

Others are welcome to contribute directly to this PR, or build upon this page if/when it's merged.

---

✏ currently includes changes up until https://github.com/discordjs/discord.js/commit/2345c6d

PR #627, opened 2021-03-28
Client Intents specification

dolphinh:

In ClientOptions, websocket options and intents are now separate and will cause an invalid error

### DIFF
--- a/guide/popular-topics/intents.md
+++ b/guide/popular-topics/intents.md
@@ -53,7 +53,7 @@ Please make sure to provide the full list of gateway intents and partials you us
 
 ```js
 const { Client } = require('discord.js');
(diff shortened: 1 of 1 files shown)


tipakA:
This is already handled in #551, and is incorrect for v12 or v11.

dolphinh:
ah I guess if a separate v13 guide would be made that would be logical, this was for master :S

almostSouji:
The master branch for the guide still targets versions 11 and 12 (with branches) of discord.js.
We will archive this version of the guide under a subdomain and apply v13 in the base guide going forward, once we're closer to releasing v13 and have a solid idea of what the API looks like.

PR #634, opened 2021-04-06
refactor(oauth): Use express instead of native http module

### LINKED ISSUE #472: use express instead of http in oauth2 guide
Moved into own issue:

> I think the guide should use express instead of HTTP native library
cause the native library is very complicated for newbie/example, and express is easier than native library.
Just suggestion, don't mind me.
Server-side: https://pastebin.com/RvtZBTZ3 (edit to not expired bin)
HTML: https://pastebin.com/mZwYQwi3 (edit to not expired bin)

_Originally posted by @KurokuTetsuya in https://github.com/discordjs/guide/issues/466#issuecomment-683608408_

Danktuary:
This PR rewrites the OAuth page to use express instead of the native http module. A few touch-ups have been made as well, in terms of readability, comprehension, etc.

Closes #472.

### DIFF
--- a/code-samples/oauth/simple-oauth-webserver/index.html
+++ b/code-samples/oauth/simple-oauth-webserver/index.html
@@ -10,10 +10,10 @@
 	<a id="login" style="display: none;" href="your-oauth2-url-here">Identify Yourself</a>
 	<script>
 		function generateRandomString() {
-			const rand = Math.floor(Math.random() * 10);
 			let randomString = '';
+			const randomNumber = Math.floor(Math.random() * 10);
 
-			for (let i = 0; i < 20 + rand; i++) {
+			for (let i = 0; i < 20 + randomNumber; i++) {
 				randomString += String.fromCharCode(33 + Math.floor(Math.random() * 94));
 			}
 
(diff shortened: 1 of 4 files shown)


catcosmicice (commented):
This seems good to go!

code-samples/oauth/simple-oauth-webserver/index.html
```diff
@@ -1,59 +1,52 @@
...
-						const { username, discriminator } = response;
-						document.getElementById('info').innerText += ` ${username}#${discriminator}`;
-					})
-					.catch(console.error);
+			if (!accessToken) {
+				const randomString = generateRandomString();
+				localStorage.setItem('oauth-state', randomString);
 
+				document.getElementById('login').href += `&state=${encodeURIComponent(btoa(randomString))}`;
+				return document.getElementById('login').style.display = 'block';
```
kyranet, line 32: You can declare `document.getElementById('login')` as a variable, so you don't get it twice.

Danktuary: Normally I would but I don't think it's entirely necessary here. It's only used for these two lines; the other `document.getElementById()` call is for the info element.

PR #635, opened 2021-04-09
Update eslint install instructions

thatrobotdev:

- Marks eslint as a dev dependency, as it's recommended on https://eslint.org/docs/user-guide/getting-started
- Mark caveats in global install mentioned on the same eslint getting started page.


Danktuary:
I think we could remove the mention of the global installation command, there's no benefit to it. Installing ESLint locally is ideal and makes sense for the guide's context.

thatrobotdev:
I've removed mention of the global install.

PR #653, opened 2021-04-19
The `getUserFromMention()` returns undefined if only id was sent as a…

sayantan300:
…rgument...

My implementation will allow sending id only as argument and return back the user, so we don't need to ping the user, just not to disturb him 🤷🏻‍♂️


Danktuary:
Sorry, I don't understand the changes you made. It seems to be whitespace and image changes that aren't needed.

This page is aimed specifically at using mentions and not including IDs because of the described situations at the beginning. Building upon this implementation can be left up to the reader. Thanks. 😄

PR #654, opened 2021-04-21
Add Internal Sharding

radiantly:
Add a tip that educates the user on the existence and usage of Internal Sharding (see #guide-discussion).

### DIFF
--- a/guide/sharding/README.md
+++ b/guide/sharding/README.md
@@ -14,7 +14,7 @@ This guide only explains the basics of sharding using the built-in ShardingManag
 :::
 
 ::: tip
-Apart from ShardingManager, discord.js also supports a sharding mode known as Internal sharding. Internal sharding creates multiple websocket connections from the same process, and does not require major code changes. To enable, simply pass `shardCount: 'auto'` to the Client constructor. However, internal sharding is not ideal for bigger bots due to high memory usage of the single main process and is not further discussed in this guide.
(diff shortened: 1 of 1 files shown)


Danktuary (approved):

guide/sharding/README.md
```diff
@@ -13,6 +13,10 @@ As an application grows large, a developer may find it necessary to split their
 This guide only explains the basics of sharding using the built-in ShardingManager, which can run shards as separate processes or threads on a single machine. If you need to scale beyond that (e.g., running shards on multiple machines/containers), you can still do it with discord.js by passing appropriate options to the Client constructor. Nevertheless, you will be on your own regarding managing shards and sharing information between them.
 :::
 
+::: tip
+Apart from ShardingManager, discord.js also supports a sharding mode known as Internal sharding. Internal sharding creates multiple websocket connections from the same process, and does not require major code changes. To enable, simply pass `shardCount: 'auto'` to the Client constructor. However, internal sharding is not ideal for bigger bots due to high memory usage of the single main process and is not further discussed in this guide.
```
ChiefChippy2, line 17:
```suggestion
Apart from ShardingManager, discord.js also supports a sharding mode known as Internal sharding. Internal sharding creates multiple websocket connections from the same process, and does not require major code changes. To enable it, simply pass `shards: 'auto'` as ClientOptions to the Client constructor. However, internal sharding is not ideal for bigger bots due to high memory usage of the single main process and will not be further discussed in this guide.
```

ChiefChippy2 (changes requested):
Changed some wording because it sounds a little weird otherwise.

`shardCount: 'auto'` **does not work**, should be `shards: 'auto'`.

PR #656, opened 2021-04-27
Fixed a sentence that might cause confusion

AYUSHBlaze:
Start by creating a new command file inside `commands` folder => Start by creating a new command file inside one of the subfolders in the `commands` folder


WiseDevHelper:
Disagree, file creation would depend on the reader's folder structure and command handler, so it's best to have instructions for the structure the rest of the guide uses.

PR #658, opened 2021-05-01
Update permissions.md

no803:
Fixed a bug...


tipakA:
There was no bug here in the first place. Guide doesn't always show full path to objects, so `guild` is correct. All that this "fix" does is actually introduce a bug, since `client.guild` is not a thing.

PR #660, opened 2021-05-02
refactor(command-handling): Merge command executing page into command handling page

Danktuary:
This PR removes pieces of code and explanations  from the "Dynamically executing commands" page that may confuse the reader and overall aren't necessary.

Since that page was only two sections long, I've decided to merge it into the previous page ("Command handling"), which ends up with a less fragmented guide explanation section. The sidebar links, code-samples files, and relative page links have been deleted/updated to reflect these changes.


WiseDevHelper (approved):
LGTM! 🎉

guide/command-handling/README.md
```diff
@@ -99,7 +99,38 @@ for (const file of commandFiles) {
...
+client.on('message', message => {
+	if (!message.content.startsWith(prefix) || message.author.bot) return;
+
+	const args = message.content.slice(prefix.length).trim().split(/ +/);
+	const command = args.shift().toLowerCase();
+
+	if (!client.commands.has(command)) return;
+
+	try {
+		client.commands.get(command).execute(message, args);
```
monbrey, line 116: Not sure what concepts we want to cover here. Should the function be async and await the `.execute()`? A rejection in an async execute function won't be caught by this handler.

guide/command-handling/README.md
```diff
@@ -99,7 +99,38 @@ for (const file of commandFiles) {
...
+	} catch (error) {
+		console.error(error);
+		message.reply('there was an error trying to execute that command!');
```
monbrey, line 119:
```suggestion
		message.reply('There was an error trying to execute that command!');
```

guide/command-handling/README.md
```diff
@@ -99,7 +99,38 @@ for (const file of commandFiles) {
...
+	}
+});
+```
+
+If there isn't a command with that name, exit early. If there is, `.get()` the command, call its `.execute()` method, and pass in your `message` and `args` variables as its arguments. In case something goes wrong, log the error and report back to the member to let them know.
```
monbrey, line 124:
```suggestion
If there isn't a command with that name, you don't need to do anything further, so exit early with `return`. If there is, `.get()` the command, call its `.execute()` method, and pass in your `message` and `args` variables as its arguments. In case something goes wrong, log the error and report back to the member to let them know.
```

guide/command-handling/README.md
Danktuary: I noticed this before the PR but didn't include it in this one because it'd require a few more edits in other files. Another PR can be made to make the handler use async/await

guide/command-handling/README.md
Danktuary: Since `.reply()` replies with a "@User, text..." format, it's grammatically correct to have this lowercase

monbrey: To be changed in a future PR for v13 then?

Danktuary: Yeah, sure. There are a few other usages of `.reply()` around the guide that would need to be updated too

PR #666, opened 2021-05-13
refactor(oauth2): Update images, headers, and grammar

Danktuary:
- Images have been remade and cropped
- The "Putting it together" header has been renamed to "Implicit grant flow"
- The "OAuth2 flows" header has been renamed to "Authorization code grant flow"
- General grammar and small linting fixes have been included


nimit2801:
Hey @Danktuary 

Can you add more images for the support
1. Response we get on the redirected URL `localhost:53134/?token=asdasd`
2. Response we get on Sever CLI when we keep code in response type.
3. An example of how the URL looks, because there are two URL's we're changing.

Also an easy way yo simplify difference `code` and `token`. 
In this article we're assuming the user knows the difference between them.

Thanks for the PR @Danktuary

guide/oauth2/README.md
```diff
@@ -57,11 +57,11 @@ Although we're using express, there are many other possible alternatives to hand
...
 ### Getting an OAuth2 url
 
-Now that you have your web server up and running, it's time to get some information from Discord. Head over to [your Discord applications](https://discord.com/developers/applications/) and click "Create an application", where the following page will greet you:
+Now that you have your web server up and running, it's time to get some information from Discord. Open [your Discord applications](https://discord.com/developers/applications/), create or select an application, and head over to the "OAuth2" page.
 
-![Create an application page](./images/create-app-page.png)
+![OAuth2 application page](./images/oauth2-app-page.png)
 
-Take note of the `client id` field, the `client secret` field, and the "OAuth2" link on the left side of the page. Copy your client ID and secret into your `config.json` file; you'll need them later. For now, click on "OAuth2" and add a redirect url to `http://localhost:53134` like so:
+Take note of the `client id` and `client secret` fields. Copy your these values into your `config.json` file; you'll need them later. For now, add a redirect url to `http://localhost:53134` like so:
```
muhimur9049, line 64:
```suggestion
Take note of the `client id` and `client secret` fields. Copy these values into your `config.json` file; you'll need them later. For now, add a redirect url to `http://localhost:53134` like so:
```

nimit2801 (commented):
Hey @Danktuary

Can you add more images for the support?

1. The response we get on the redirected URL localhost:53134/?token=asdasd
2. The response we get on Sever CLI when we keep code in response type.
3. An example of how the URL looks because there are two URLs we're changing.

Also, an easy way to simplify different `codes` and `token`.
In this article, we're assuming the user knows the difference between them.

Thanks for the PR @Danktuary

Danktuary:
> 1. The response we get on the redirected URL localhost:53134/?token=asdasd

Good point; will do.

> 2. The response we get on Sever CLI when we keep code in response type.

This is already mentioned at the bottom of the "Authorization code grant section", i.e.:
```json
{
	"access_token": "an access token",
	"token_type": "Bearer",
	"expires_in": 604800,
	"refresh_token": "a refresh token",
	"scope": "identify"
}
```
However, it doesn't include an example of the Discord user object received after fetching through the `/api/users/@me` endpoint. I'll add that with a link to the Discord dev docs.

> 3. An example of how the URL looks because there are two URLs we're changing.

There's only one URL to change, and it's the mentioned under the "Getting an OAuth2 url" section, so I believe that when making references such as "change the URL", this should be understandable.
Technically speaking, there is the redirect URL, but that doesn't change and is referenced as "the redirect URL" throughout the page.

> Also, an easy way to simplify different `codes` and `token`.
> In this article, we're assuming the user knows the difference between them.

The difference is the OAuth2 flow you follow; this article explains the difference between both in the initial paragraph of the "Implicit grant flow" and "Authorization code grant flow" sections.

PraveshKunwar:
Good job @Danktuary , looks pretty good!

PR #676, opened 2021-06-07
Updated README.md

Spaxly:
Its more sufficient to import everything from the config file and not manually add everything!

Why should be merged and what it changes: 

For beginners that don't need to import specific items from the config file. If you import the whole file it would be easier as you would have to keep changing that to import more objects in from the file. That's what the changes are and why it should be merged.


almostSouji:
Surely a possibility, but I think we will stick with promoting the use of new language features. This is a great opportunity to learn about destructuring, which can come in very handy further down the line in your coding adventures!

PR #677, opened 2021-06-07
Add clarification about selecting redirect URL

SorraTheOrc:
Just an explicit pointer to something that was "hidden" in an image and this I missed it at first.


Danktuary:
I don't think this is necessary because once you start clicking the checkboxes, if no redirect URL is selected from the menu, the result at the bottom will read "Please enter a redirect uri". Asie

PR #681, opened 2021-06-15
Switch startsWith to IndexOf

LKNSI:
While the example is perfectly correct as it is, encouraging people to use IndexOf instead of startsWith is really a better choice for performance - and not just a few extra operations, but thousands of more operations per second.

Change checks if indexOf returns 0, or the first index of the string, which produces the same result.

https://www.measurethat.net/Benchmarks/Show/4797/1/js-regex-vs-startswith-vs-indexof

### DIFF
--- a/guide/command-handling/README.md
+++ b/guide/command-handling/README.md
@@ -15,7 +15,7 @@ client.once('ready', () => {
 });
 
 client.on('message', message => {
-	if ((message.content.indexOf(prefix) !== 0)|| message.author.bot) return;
+	if ((message.content.indexOf(prefix) !== 0) || message.author.bot) return;
 
 	const args = message.content.slice(prefix.length).trim().split(/ +/);
 	const command = args.shift().toLowerCase();
@@ -105,7 +105,7 @@ With your `client.commands` Collection setup, you can use it to retrieve and exe
 
 ```js {7-14}
 client.on('message', message => {
-	if (!message.content.startsWith(prefix) || message.author.bot) return;
+	if ((message.content.indexOf(prefix) !== 0) || message.author.bot) return;
 
 	const args = message.content.slice(prefix.length).trim().split(/ +/);
 	const command = args.shift().toLowerCase();
--- a/guide/command-handling/README.md
+++ b/guide/command-handling/README.md
@@ -15,7 +15,7 @@ client.once('ready', () => {
 });
 
 client.on('message', message => {
-	if (!message.content.startsWith(prefix) || message.author.bot) return;
+	if ((message.content.indexOf(prefix) !== 0)|| message.author.bot) return;
 
 	const args = message.content.slice(prefix.length).trim().split(/ +/);
 	const command = args.shift().toLowerCase();


LKNSI:
image

Keke27210:
The goal of this guide is to be understandable to most people, including beginners. Also, for something as critical as a Discord bot there will surely be 0.00% of noticeable performance difference between `String#startsWith` and `String#indexOf`.

Something like `if ("abc".startsWith("a"))` is more likely to be understood by the most, compared to something like `if ("abc".indexOf("a") === 0)`. Special mention to `if ("abc".indexOf("a"))` that would confuse people a lot, imo.

Now, the fact that `String#indexOf` is faster than `String#startsWith` is wrong. Or, at least, not completely true. It mostly depends on the implementation and many other factors. Here is a proof with the example you provided... where `indexOf` actually performs the worst compared to the rest.
I have no idea how node.js performs on its own, the example below comes from Safari 13 (JavascriptCore engine, not the best but an example of my sayings).

Capture d’écran 2021-06-15 à 22 51 33

kyranet:
Here comes the test results. The benchmarks say otherwise! String#startsWith out-performs String#indexOf by twice in the minimum time case:

image

Code run (requires Node.js v16.0.0 minimum):

<details>
<summary>How do I dropdown?</summary>
```typescript
import { bold, cyan } from 'colorette';
import { createHistogram, performance } from 'perf_hooks';

const reStart = /^test/;
const reEnd = /test$/;
const match = 'test';
const matchLength = match.length;
const data = [];
const possible = 'ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789';
const TOTAL_STRINGS = 10000;

function getRandomInt(max) {
	return Math.floor(Math.random() * max);
}

function makeRandomString(len) {
	let text = '';
	for (let i = 0; i < len; ++i) {
		text += possible.charAt(getRandomInt(possible.length));
	}

	return text;
}

while (data.length < TOTAL_STRINGS) {
	data.push(makeRandomString(getRandomInt(20)));
}

const regexStartHistogram = createHistogram();
const startsWithHistogram = createHistogram();
const regexEndHistogram = createHistogram();
const endsWithHistogram = createHistogram();
const indexOfStartHistogram = createHistogram();
const indexOfEndHistogram = createHistogram();

const regexStartResults = new Array(TOTAL_STRINGS);
const startsWithResults = new Array(TOTAL_STRINGS);
const regexEndResults = new Array(TOTAL_STRINGS);
const endsWithResults = new Array(TOTAL_STRINGS);
const indexOfStartResults = new Array(TOTAL_STRINGS);
const indexOfEndResults = new Array(TOTAL_STRINGS);

const regexStart = performance.timerify(
	() => {
		for (let i = 0; i < TOTAL_STRINGS; ++i) {
			regexStartResults[i] = reStart.test(data[i]);
		}
	},
	{ histogram: regexStartHistogram }
);

const startsWith = performance.timerify(
	() => {
		for (let i = 0; i < TOTAL_STRINGS; ++i) {
			startsWithResults[i] = data[i].startsWith(match);
		}
	},
	{ histogram: startsWithHistogram }
);

const regexEnd = performance.timerify(
	() => {
		for (let i = 0; i < TOTAL_STRINGS; ++i) {
			regexEndResults[i] = reEnd.test(data[i]);
		}
	},
	{ histogram: regexEndHistogram }
);

const endsWith = performance.timerify(
	() => {
		for (let i = 0; i < TOTAL_STRINGS; ++i) {
			endsWithResults[i] = data[i].endsWith(match);
		}
	},
	{ histogram: endsWithHistogram }
);

const indexOfStart = performance.timerify(
	() => {
		for (let i = 0; i < TOTAL_STRINGS; ++i) {
			indexOfStartResults[i] = data[i].indexOf(match) === 1;
		}
	},
	{ histogram: indexOfStartHistogram }
);

const indexOfEnd = performance.timerify(
	() => {
		for (let i = 0; i < TOTAL_STRINGS; ++i) {
			const str = data[i];
			indexOfEndResults[i] = str.indexOf(match) === str.length - matchLength;
		}
	},
	{ histogram: indexOfEndHistogram }
);

for (let n = 0; n < 1e4; n++) regexStart();
for (let n = 0; n < 1e4; n++) startsWith();
for (let n = 0; n < 1e4; n++) regexEnd();
for (let n = 0; n < 1e4; n++) endsWith();
for (let n = 0; n < 1e4; n++) indexOfStart();
for (let n = 0; n < 1e4; n++) indexOfEnd();

function display(name, h) {
	console.log(`- ${cyan(name)}:`);
	console.log(`  Min : ${bold((h.min / (1000 * 1000)).toFixed(3))} ms`);
	console.log(`  Max : ${bold((h.max / (1000 * 1000)).toFixed(3))} ms`);
	console.log(`  Mean: ${bold((h.mean / (1000 * 1000)).toFixed(3))} ms`);
	console.log(`    stddev: ${bold((h.stddev / (1000 * 1000)).toFixed(3))} ms`);
}

display('RegExp Start', regexStartHistogram);
display('RegExp End', regexEndHistogram);
display('String#startsWith', startsWithHistogram);
display('String#endsWith', endsWithHistogram);
display('String#indexOf Start', indexOfStartHistogram);
display('String#indexOf End', indexOfEndHistogram);

console.log(`regexStartResults.length: ${regexStartResults.length}`);
console.log(`startsWithResults.length: ${startsWithResults.length}`);
console.log(`regexEndResults.length: ${regexEndResults.length}`);
console.log(`endsWithResults.length: ${endsWithResults.length}`);
console.log(`indexOfStartResults.length: ${indexOfStartResults.length}`);
console.log(`indexOfEndResults.length: ${indexOfEndResults.length}`);
```
</details>

---

For future reference, please use a real benchmark suite.

Drahcirius:
Like keke mentioned, correctness is much more important in a tutorial than small micro-optimizations. And kyra has benchmark code that show otherwise. 

I suspect the reason the results are so different is because your benchmark is a victim of an optimization called [dead code elimination](https://en.wikipedia.org/wiki/Dead_code_elimination). 

I took your benchmark code and, added a condition so that there will be side effects, and as expected, the benchmarks skew the other way. 

https://www.measurethat.net/Benchmarks/Show/13467/0/js-regex-vs-startswith-vs-indexof-with-side-effects

This is why you shouldn't prematurely optimize. You were actually slowing down your code rather than speed it up with your trick.

LKNSI:
> This is why you shouldn't prematurely optimize. You were actually slowing down your code rather than speed it up with your trick.

Valid points, I see the reasoning behind that - makes sense. Thank you for explaining the discrepancies between the tests.

> For future reference, please use a real benchmark suite.

Regardless of not having the time to setup a full benchmarkjs for this, this was a valid benchmark suite running on the same engine for the target platform. The real issue was that of what was explained above, and I'm better off from learning that.

PR #683, opened 2021-06-20
refactor(Commandhandling): rewrite command handling pages to be based upon interactions

adrifcastr:

After talking to Crawl, the verdict fell to also rewrite the sections that incorporate the command handling to use interactions/slash commands.

Now there's two things on which I'd need @Danktuary 's input on:

- the `additional features` section seems now obsolete as interactions cover most of these features
 should this section be removed entirely? should only obsolete parts be removed?

- the `commands-with-user-input section` also contains a lot of obsolete content
should that be rewritten entirely to adapt interactions, or should the obsolete parts be hidden away with the `branch` tags? (There's a lot so it'd be like half the page)


code-samples/command-handling/file-setup/13/commands/prune.js
```diff
@@ -0,0 +1,26 @@
...
+			{
+				name: 'amount',
+				description: 'Number of messages to prune',
+				type: 'INTEGER',
+				required: true,
+			},
+		],
+	},
+	async run(interaction) {
+		const amount = interaction.options.get('amount');
```
Danktuary, line 15: Should this be `.get().value`?

code-samples/command-handling/file-setup/13/commands/prune.js
```diff
@@ -0,0 +1,26 @@
...
+
+		if (amount <= 1 || amount > 100) return interaction.reply({ content: 'You need to input a number between 1 and 99.', ephemeral: true });
```
Danktuary, line 17: Multiline this

code-samples/command-handling/file-setup/13/commands/prune.js
```diff
@@ -0,0 +1,26 @@
...
+		
+		await interaction.channel.bulkDelete(amount, true).catch(err => {
+			console.error(err);
```
Danktuary, line 20:
```suggestion
		await interaction.channel.bulkDelete(amount, true).catch(error => {
			console.error(error);
```

Danktuary (changes requested):
A few things to note for the code inside commands:
- Destructuring should be used wherever possible, e.g.:
```js
const { user } = get('target')
const { value: option } = get('input')
```
- Ternaries should be replaced with early return if statements, e.g.:
```js
if (!user) return reply('No user')
return reply('User exists')
```
- Commands should `return` at the end of their methods, not `await`
- If there's no real benefit to nesting the name and descriptions inside a `data` object, I'd prefer to keep them at the top level
- To keep things as they previously were, I prefer `execute` over `run`.

Of course, if there's anything you believe should stay as it is, I'm open to suggestions

Danktuary:
> the additional features section seems now obsolete as interactions cover most of these features
should this section be removed entirely? should only obsolete parts be removed?

> the commands-with-user-input section also contains a lot of obsolete content
should that be rewritten entirely to adapt interactions, or should the obsolete parts be hidden away with the branch tags? (There's a lot so it'd be like half the page)

I suppose neither of these would be needed with slash commands. Arguments are handled by slash commands, and the only applicable additional features I could see being useful would be cooldowns, command permissions, and maybe the dynamic help command.

almostSouji:
> obsolete parts be hidden away with the branch tags

nope, no branching.
this will all merge into the v13 guide overhaul, which is a clean slate in terms of branches, as v11/12 will be moved to it's own subdomain

adrifcastr:
> Of course, if there's anything you believe should stay as it is, I'm open to suggestions

Thanks, will address these changes in a bit.

The reason why I nested the command data is:

- that way people can easily grab it and register it, if there's other properties they want to export in their command files that would interfere with the structure of an ApplicationCommandData object, hence why I nested it down 

- renaming execute to run just felt more fluid while reading, will revert that back to how it was

guide/command-handling/README.md
```diff
@@ -6,34 +6,28 @@ Here's the base code we'll be using:
...
+	if (interacton.commandName === 'ping') {
+		await interaction.reply('Pong.');
+	} else if (interacton.commandName === 'beep') {
+		await interaction.reply('Boop!');
 	}
 	// ...
 });
 
 client.login(token);
 ```
```
almostSouji, line 30:
```suggestion

```
obsolete additional newline

guide/creating-your-bot/adding-more-commands.md
```diff
@@ -34,54 +36,41 @@ If you aren't familiar with some of this syntax, it may be ES6 syntax. If it doe
...
-		message.channel.send('Pong.');
-	} else if (message.content.startsWith(`${prefix}beep`)) {
-		message.channel.send('Boop.');
+	if (interacton.commandName === 'ping') {
+		await interaction.reply('Pong!');
+	} else if (interacton.commandName === 'beep') {
+		await interaction.reply('Boop!');
 	}
 });
 ```
```
almostSouji, line 50:
```suggestion

```
obsolete additional newline

almostSouji (changes requested):
newline nitpicks, else looking good

adrifcastr:
> newline nitpicks, else looking good

Applied 👍🏼

almostSouji (approved):

code-samples/command-handling/file-setup/13/commands/ping.js
```diff
@@ -0,0 +1,9 @@
+module.exports = {
+	data: {
+		name: 'ping',
+		description: 'Replies with Pong!',
+	},
+	async execute(interaction) {
+		return interaction.reply('Pong!');
+	},
+};
```
monbrey, line 9:
```suggestion
};

```

code-samples/command-handling/file-setup/13/commands/avatar.js
```diff
@@ -0,0 +1,18 @@
...
+		options: [
+			{
+				name: 'target',
+				description: 'The user who\'s avatar to show',
+				type: 'USER',
+			},
+		],
+	},
+	async execute(interaction) {
+		const { user } = interaction.options?.get('target');
```
monbrey, line 14:
```suggestion
		const { user } = interaction.options.get('target');
```
Do these example's assume the user has already validated `interaction#isCommand()` in the handler? If so, `CommandInteraction#options` isn't nullable, it will just be an empty Collection.

If we don't want to make that assumption, ignore these suggestions.

code-samples/command-handling/file-setup/13/commands/kick.js
```diff
@@ -0,0 +1,18 @@
...
+			{
+				name: 'target',
+				description: 'The member to kick',
+				type: 'USER',
+				required: true,
+			},
+		],
+	},
+	async execute(interaction) {
+		const { user } = interaction.options?.get('target');
```
monbrey, line 15:
```suggestion
		const { user } = interaction.options.get('target');
```

code-samples/command-handling/file-setup/13/commands/options-info.js
```diff
@@ -0,0 +1,18 @@
...
+		options: [
+			{
+				name: 'input',
+				description: 'The input to echo back',
+				type: 'STRING',
+			},
+		],
+	},
+	async execute(interaction) {
+		const { value } = interaction.options?.get('input');
```
monbrey, line 14:
```suggestion
		const { value } = interaction.options.get('input');
```

code-samples/command-handling/file-setup/13/commands/prune.js
```diff
@@ -0,0 +1,28 @@
...
+			{
+				name: 'amount',
+				description: 'Number of messages to prune',
+				type: 'INTEGER',
+				required: true,
+			},
+		],
+	},
+	async execute(interaction) {
+		const { value: amount } = interaction.options?.get('amount');
```
monbrey, line 15:
```suggestion
		const { value: amount } = interaction.options.get('amount');
```

code-samples/command-handling/file-setup/13/index.js
```diff
@@ -0,0 +1,32 @@
...
+	const command = require(`./commands/${file}`);
+	client.commands.set(command.data.name, command);
+}
+
+client.once('ready', () => {
+	console.log('Ready!');
+});
+
+client.on('interaction', async interaction => {
+	if (!interaction.isCommand()) return;
```
monbrey, line 20: Okay yeah, we do validate first so there's no need for optional chaining.

monbrey:
> A few things to note for the code inside commands:
> 
>     * Destructuring should be used wherever possible, e.g.:
> 
> 
> ```js
> const { user } = get('target')
> const { value: option } = get('input')
> ```

Wouldn't this throw a `Cannot destructure property 'user' of undefined` if 'target' isn't an option? It makes the validation on the next line a little confusing, since it implies that the option *did* exist but for some reason it *wasn't* a USER type / didn't resolve a User.

adrifcastr:
also rebased

guide/command-handling/README.md
```diff
@@ -95,33 +90,30 @@ for (const file of commandFiles) {
...
 
 ## Dynamically executing commands
 
-With your `client.commands` Collection setup, you can use it to retrieve and execute your commands! Inside your `message` event, delete your `if`/`else if` chain of commands and replace it with this:
-
-```js {7-14}
-client.on('message', message => {
-	if (!message.content.startsWith(prefix) || message.author.bot) return;
+With your `client.commands` Collection setup, you can use it to retrieve and execute your commands! Inside your `interaction` event, delete your `if`/`else if` chain of commands and replace it with this:
```
WiseDevHelper, line 99:
```suggestion
With your `client.commands` Collection setup, you can use it to retrieve and execute your commands! Inside your `interactionCreate` event, delete your `if`/`else if` chain of commands and replace it with this:
```

guide/creating-your-bot/adding-more-commands.md
```diff
@@ -1,25 +1,27 @@
...
+		await interaction.reply('Pong!');
 	}
 });
 ```
 
-Before doing anything else, make a property to store the prefix you've configured. Instead of `const config = ...`, you can destructure the config file to extract the prefix and token variables.
+Before doing anything else, make a property to store the token. Instead of `const config = ...`, you can destructure the config file to extract the token variable.
 
 ```js {1,3}
-const { prefix, token } = require('./config.json');
```
WiseDevHelper, line 22: Line 29 below could be replaced with the following (maybe also make it a tip?).

```
From now on, if you change the token in your `config.json` file, it'll change in your bot file as well!
```

guide/creating-your-bot/configuration-files.md
```diff
@@ -32,7 +30,7 @@ const config = require('./config.json');
 
 Next, copy your token from the `client.login('your-token-goes-here')` line and paste into the `config.json` file. Make sure to keep it between the double-quotes.
 
-Now you can simply do `client.login(config.token)` to login! If you want to use a different prefix than `!`, you can change that as well.
+Now you can simply do `client.login(config.token)` to login!
 
 ## Storing additional data
```
WiseDevHelper, line 35: The `prefix` variable must be removed here too.

guide/command-handling/README.md
```diff
@@ -95,33 +90,30 @@ for (const file of commandFiles) {
...
-		message.reply('there was an error trying to execute that command!');
+		await interaction.reply({ content: 'There was an error while executing this command!', ephemeral: true });
 	}
 });
 ```
 
-If there isn't a command with that name, you don't need to do anything further, so exit early with `return`. If there is, `.get()` the command, call its `.execute()` method, and pass in your `message` and `args` variables as its arguments. In case something goes wrong, log the error and report back to the member to let them know.
+If there isn't a command with that name, you don't need to do anything further, so exit early with `return`. If there is, `.get()` the command, call its `.execute()` method, and pass in your `interaction` variable as its argument. In case something goes wrong, log the error and report back to the member to let them know.
 
 And that's it! Whenever you want to add a new command, you make a new file in your `commands` directory, name it what you want, and then do what you did for the other commands.
```
WiseDevHelper, line 119: Below paragraph should be removed?

WiseDevHelper (changes requested):
A few corrections, else LGTM!

guide/creating-your-bot/adding-more-commands.md
```diff
@@ -34,96 +36,84 @@ If you aren't familiar with some of this syntax, it may be ES6 syntax. If it doe
...
-:::
-
 ## Displaying real data
 
 Let's start displaying some real data. For now, we'll be displaying basic member/server info.
 
 ### Server info command
 
-Make another if statement to check for commands using `server` as the command name. You've already interacted with the Message object via `message.channel.send()`. You get the message object, access the channel it's from, and send a message to it. Just like how `message.channel` gives you the message's _channel_, `message.guild` gives you the message's _server_.
+Make another if statement to check for commands using `server` as the command name. You get the interaction object, reply to the interaction just as before:
```
Danktuary, line 57:
```suggestion
Make another if statement to check for commands using `server` as the command name. You get the interaction object and reply to the interaction just as before:
```

guide/creating-your-bot/adding-more-commands.md
```diff
@@ -135,34 +125,36 @@ Set up another if statement and use the command name `user-info`.
...
 <DiscordMessages>
-	<DiscordMessage profile="user">
-		!user-info
-	</DiscordMessage>
 	<DiscordMessage profile="bot">
+		<template #interactions>
+			<DiscordInteraction profile="user" :command="true">user-info</DiscordInteraction>
+		</template>
 		Your username: User <br>
 		Your ID: 20833034795932416
```
Danktuary, line 152:
```suggestion
		Your ID: 123456789012345678
```
Old ID is probably one of my deleted applications

guide/creating-your-bot/adding-more-commands.md
```diff
@@ -135,34 +125,36 @@ Set up another if statement and use the command name `user-info`.
...
 	</DiscordMessage>
 </DiscordMessages>
 
 ::: tip
-`message.author` refers to the user who sent the message. For a full list of all the properties and methods for the author object (a member of the `User` class), check out <DocsLink path="class/User">the documentation page for it</DocsLink>.
+`interaction.user` refers to the user who sent the command. For a full list of all the properties and methods for the user object, check out <docs-link path="class/User">the documentation page for it</docs-link>.
```
Danktuary, line 157:
```suggestion
`interaction.user` refers to the user who sent the command. For a full list of all the properties and methods for the user object, check out <DocsLink path="class/User">the documentation page for it</DocsLink>.
```

Danktuary (changes requested):
> Wouldn't this throw a Cannot destructure property 'user' of undefined if 'target' isn't an option? It makes the validation on the next line a little confusing, since it implies that the option did exist but for some reason it wasn't a USER type / didn't resolve a User.

I think this can be solved via `const { user } = get('target') ?? {}` syntax

> The reason why I nested the command data is:
that way people can easily grab it and register it, if there's other properties they want to export in their command files that would interfere with the structure of an ApplicationCommandData object, hence why I nested it down

Makes sense, but it's just as easy to do `const { name, description, options } = command` while having additional properties in command files. Imo this is a simpler approach, as users won't have to nest this data when creating new commands.

Aside from the stuff mentioned above and the other few comments, looks good 🎉

adrifcastr:
> Aside from the stuff mentioned above and the other few comments, looks good 🎉

Adressed 👍🏼

code-samples/command-handling/file-setup/13/commands/prune.js
```diff
@@ -0,0 +1,26 @@
...
+		{
+			name: 'amount',
+			description: 'Number of messages to prune',
+			type: 'INTEGER',
+			required: true,
+		},
+	],
+
+	async execute(interaction) {
+		const { value: amount } = interaction.options.get('amount') ?? {};
```
Danktuary, line 14:
```suggestion
		const { value: amount } = interaction.options.get('amount');
```

code-samples/command-handling/file-setup/13/commands/kick.js
```diff
@@ -0,0 +1,17 @@
...
+		{
+			name: 'target',
+			description: 'The member to kick',
+			type: 'USER',
+			required: true,
+		},
+	],
+
+	async execute(interaction) {
+		const { user } = interaction.options.get('target') ?? {};
```
Danktuary, line 14:
```suggestion
		const { user } = interaction.options.get('target');
```

Danktuary (commented):
Required options should always have a value

Danktuary (approved):

PR #689, opened 2021-06-29
Suggestion: Eslint Rules

JPBM135:
Update the Eslint rules to support the ES2021 since the guide and the docs use the optional chaining and the null coalescing operator and in the ES6 rule it gives a warn

### DIFF
--- a/guide/preparations/setting-up-a-linter.md
+++ b/guide/preparations/setting-up-a-linter.md
@@ -64,8 +64,7 @@ Alternatively, if you don't want to go through everything one by one on your own
 	"extends": "eslint:recommended",
 	"env": {
 		"node": true,
-		"es6": true,
-		"es2021": true
+		"es6": true
 	},
 	"parserOptions": {
 		"ecmaVersion": 2021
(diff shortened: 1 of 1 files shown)


WiseDevHelper:
IMO, a few more rules should be added to the set:
```json
	"arrow-spacing": ["warn", { "before": true, "after": true }],
	"keyword-spacing": "error",
```

guide/preparations/setting-up-a-linter.md
```diff
@@ -64,10 +64,11 @@ Alternatively, if you don't want to go through everything one by one on your own
 	"extends": "eslint:recommended",
 	"env": {
 		"node": true,
-		"es6": true
+		"es6": true,
+		"es2021": true
```
Danktuary, line 68:
```suggestion
		"es6": true
```
Only one of these is needed in `env`

Danktuary:
> IMO, a few more rules should be added to the set:
> 
> ```json
> 	"arrow-spacing": ["warn", { "before": true, "after": true }],
> 	"keyword-spacing": "error",
> ```

These are good rules but I'll leave them out for now. I feel like this page could use a small revamp; I'll take care of that eventually.

PR #693, opened 2021-07-06
feat: v13 collectors

monbrey:
Figured I might as well get started on some updates - this updates the Collectors page to bring it in line with v13, and adds a section on interaction collectors.

### DIFF
--- a/guide/popular-topics/collectors.md
+++ b/guide/popular-topics/collectors.md
@@ -15,7 +15,7 @@ For now, let's take the example that they have provided us:
 ```js
 // `m` is a message object that will be passed through the filter function
 const filter = m => m.content.includes('discord');
-const collector = message.channel.createMessageCollector(filter, { time: 15000 });
+const collector = message.channel.createMessageCollector({ filter, time: 15000 });
 
 collector.on('collect', m => {
 	console.log(`Collected ${m.content}`);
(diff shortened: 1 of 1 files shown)


guide/popular-topics/collectors.md
```diff
@@ -26,17 +26,17 @@ collector.on('end', collected => {
...
 
-You can also allow more than one condition, as you would with any function. An alternative could be `m => m.content.includes('discord') && m.author.id === message.author.id`, assuming `message` is the name of what you receive in the `message` event. This function will only allow a message sent by the person who triggered the command *and* if the message content included "discord" in it.
+You can also allow more than one condition, as you would with any function. An alternative could be `m => m.content.includes('discord') && m.author.id === message.author.id`, assuming `message` is the name of what you receive in the `messageCreate` event. This function will only allow a message sent by the person who triggered the command *and* if the message content included "discord" in it.
```
almostSouji, line 31:
```suggestion
```

incorporated into rewrite above, more concise using MDN resources

guide/popular-topics/collectors.md
```diff
@@ -26,17 +26,17 @@ collector.on('end', collected => {
...
 
 After a message passes through, this will trigger the `collect` event for the `collector` you've created, which will then run the provided function. In this case, it will only log the collected message. Once the collector finishes, one way or another, it will run the `end` event. A collector can end in different ways, such as:
```
almostSouji, line 33:
```suggestion
If a message passes through the filter, it will trigger the `collect` event for the `collector` you've created. This message is then passed into the event listener as `collected` and the provided function is executed. In the above example, you simply log the message. Once the collector finishes collecting based on the provided end conditions the `end` event emits.

You can control when a collector ends by supplying additional option keys when creating a collector:
```
clarity

guide/popular-topics/collectors.md
```diff
@@ -26,17 +26,17 @@ collector.on('end', collected => {
...
 
-* Time running out
-* A certain number of messages passing the filter
-* A certain number of attempts to go through the filter altogether
+* Time running out: `time`
+* A certain number of messages passing the filter: `max`
+* A certain number of attempts to go through the filter altogether: `maxProcessed`
```
almostSouji, line 37:
```suggestion
* `time`: Amount of time in milliseconds the collector should run for
* `max`:  Number of messages to successfully pass the filter
* `maxProcessed`: Number of messages encountered (no matter the filter result)
```
clarity

guide/popular-topics/collectors.md
```diff
@@ -26,17 +26,17 @@ collector.on('end', collected => {
...
 
-Those options you pass as the second argument in `.createMessageCollector()`. The benefit of using this method over `.awaitMessages()` is that you can stop it manually by calling `collector.stop()`, should you have your own reason to interrupt the collecting early.
+Those are additional options you pass in the object to `.createMessageCollector()`. The benefit of using this method over `.awaitMessages()` is that you can stop it manually by calling `collector.stop()`, should you have your own reason to interrupt the collecting early.
```
almostSouji, line 39:
```suggestion
The benefit of using an event-based collector over `.awaitMessages()` (its promise-powered counterpart) is that you can do something after every collection, rather than just after the collector ended. You can also stop the collector manually by calling `collector.stop()`.
```
main benefit imo is that you can do stuff on every collect, added that

guide/popular-topics/collectors.md
```diff
@@ -128,9 +128,52 @@ const filter = (reaction, user) => {
...
 	});
 ```
+
+## Interaction collectors
+
+The third type of collector allows you to collect interactions; such as when users activate a slash command or click on a button in a message.
+
+### Basic message component collector
+
+Collecting interactions from message components works quite similarly to reaction collectors. In the following example, we check that the interaction came from a button, and that user interacting with the button shares the same id as the user who activated the slash command that caused the bot to send them.
```
almostSouji, line 144:
```suggestion
Collecting interactions from message components works similarly to reaction collectors. In the following example,  you will check that the interaction came from a button, and that the user clicking the button is the same user that initiated the command.
```
we -> you, clarity, obsolete "quite"

guide/popular-topics/collectors.md
```diff
@@ -128,9 +128,52 @@ const filter = (reaction, user) => {
...
+
+One important difference to note with interaction collectors however is that Discord expects a response to *all* interactions within 3 seconds - even ones that you don't want to collect. For this reason, you may wish to `.deferUpdate()` all interactions in your filter, or not use a filter at all and handle this behaviour in the `collect` event.
```
almostSouji, line 146:
```suggestion
One important difference to note with interaction collectors is that Discord expects a response to *all* interactions within 3 seconds - even ones that you don't want to collect. For this reason, you may wish to `.deferUpdate()` all interactions in your filter, or not use a filter at all and handle this behavior in the `collect` event.
```
AE, obsolete "however"

guide/popular-topics/collectors.md
```diff
@@ -128,9 +128,52 @@ const filter = (reaction, user) => {
...
+collector.on('end', collected => {
+	console.log(`Collected ${collected.size} interactions.`);
+});
+```
+
+### Await message component
+
+As before, this works similarly to the message component collector, except it is Promise-based.
+
+Unlike other Promisified collectors though, this method will only ever collect one interaction which passes the filter. If no interactions are collected before the time runs out, the Promise will reject. This is to align with Discord's requirement that actions should immediately receive a response. In this example, we choose to `.deferUpdate()` all interactions in the filter.
```
almostSouji, line 168:
```suggestion
Unlike other Promise-based collectors, this method will only ever collect one interaction that passes the filter. If no interactions are collected before the time runs out, the Promise will reject. This behavior aligns with Discord's requirement that actions should immediately receive a response. In this example, you will use `.deferUpdate()` on all interactions in the filter.
```
we -> you, grammar, clarity

almostSouji (changes requested):
- Trying to make the section a bit easier to read and comprehend (maybe even for non-native speakers and translator users)
- we -> you, whenever possible
- grammar nits
- BE -> AE nits

almostSouji (approved):

PR #694, opened 2021-07-06
feat: v13 embeds

monbrey:
Updated to v13 syntax

### DIFF
--- a/guide/popular-topics/embeds.md
+++ b/guide/popular-topics/embeds.md
@@ -79,7 +79,7 @@ const exampleEmbed = new Discord.MessageEmbed()
 	.setTimestamp()
 	.setFooter('Some footer text here', 'https://i.imgur.com/wSTFkRM.png');
 
-channel.send(exampleEmbed);
+channel.send({ embeds: [exampleEmbed] });
 ```
 
 ::: tip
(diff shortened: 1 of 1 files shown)


guide/popular-topics/embeds.md
```diff
@@ -103,29 +103,6 @@ if (message.author.bot) {
...
-	.attachFiles(['../assets/discordjs.png'])
-	.setImage('attachment://discordjs.png');
-
-channel.send(exampleEmbed);
-```
-
-::: warning
-If the images don't display inside the embed but outside of it, double-check your syntax to make sure it's as shown above.
-:::
-
```
iCrawl, line 128: This shouldn't be removed, but rather be updated how to do it now IMO.

monbrey: It was essentially duplicated on the page, once for the builder and once for the object. I've removed one of them and merged in into one section with two examples since both now require it to be attached in MessageOptions the same way.

monbrey: https://github.com/discordjs/guide/pull/694/files#diff-d54649caf99cc64d1693c6dc2f19c47d0055059cc4fd3edc2f909dc943d521bdR187

iCrawl: Right, I didn't see that. 👍

guide/popular-topics/embeds.md
```diff
@@ -207,6 +184,20 @@ You can then reference and use the images inside the embed itself with `attachme
 If you plan to attach the same image repeatedly, consider hosting it online and providing the URL in the respective embed field instead. This also makes your bot respond faster since it doesn't need to upload the image with every response depending on it.
 :::
 
+#### Using the MessageEmbed builder
+
+```js
+const file = new Discord.MessageAttachment('../assets/discordjs.png');
+
```
almostSouji, line 191:
```suggestion
```

almostSouji (approved):

Danktuary (changes requested):
The "Attaching Images" header should also go from `###` to `##`, because otherwise all of these subsections be listed under the "Using an embed object" section.

PR #709, opened 2021-07-14
fix(Content): Some Miscellaneous changes

SinisterDeveloper:

* List errors based on frequency
Rearrange a few Errors in `popular-topics/errors.md` based on the frequency of questions asked regarding them.

* Add simple eval command(supporting await) in `faq.md`

* A few words have been enhanced

### DIFF
--- a/guide/keyv/README.md
+++ b/guide/keyv/README.md
@@ -63,10 +63,10 @@ Keyv exposes a familiar [Map](https://developer.mozilla.org/en-US/docs/Web/JavaS
 
 ## Application
 
-Although Keyv can assist in any scenario where you need key-value data, we will focus on setting up a per-guild prefix configuration using Sqlite.
+Although Keyv can assist in any scenario where you involve key-value data, we will focus on setting up a per-guild prefix configuration using Sqlite.
 
 ::: tip
-This section will still work with any provider supported by Keyv. We recommend PostgreSQL for larger applications.
+This section will still work with any provider supported by Keyv. We recommend [PostgreSQL](https://www.postgresql.org/) for larger applications.
 :::
 
 ### Setup
@@ -101,7 +101,7 @@ client.on('message', async message => {
 			if (message.content.startsWith(guildPrefix)) prefix = guildPrefix;
 		}
 
-		// if we found a prefix, setup args; otherwise, this isn't a command
+		// if we found a prefix, setup args. Otherwise, this isn't a command
 		if (!prefix) return;
 		args = message.content.slice(prefix.length).trim().split(/\s+/);
 	} else {
--- a/guide/popular-topics/errors.md
+++ b/guide/popular-topics/errors.md
@@ -131,12 +131,12 @@ You can debug these messages in different ways:
 ### Cannot send messages to this user.
 
 This error throws when the bot attempts to send a DM message to a user but cannot do so. A variety of reasons may cause this error:
-- The bot and the user do not share a guild (often, people attempt to dm the user after kicking or banning them).
+- The bot and the user do not share a guild (often, people attempt to DM the user after kicking or banning them).
 - The bot tries to DM another bot.
 - The user has blocked the bot.
-- The user has disabled dms in the privacy settings.
+- The user has disabled DMs in the privacy settings.
 
-In the case of the last two reasons, the error is not preventable, as the Discord API does not provide a way to check if you can send a user a dm until you attempt to send one. The best way to handle this error is to add a `.catch()` where you try to dm the user and either ignore the rejected Promise or do what you want because of it.
+In the case of the last two reasons, the error is not preventable, as the Discord API does not provide a way to check if you can send a user a DM until you attempt to send one. The best way to handle this error is to add a `.catch()` where you try to DM the user and either ignore the rejected Promise or do what you want because of it.
 
 ### An invalid token was provided.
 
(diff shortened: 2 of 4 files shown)


guide/popular-topics/errors.md
```diff
@@ -120,6 +120,24 @@ In this particular example, you can see you are trying to access a piece of data
...
+
+This error is caused by spawning a large number of event listeners, usually for the client. The most common cause of this is nesting your event listeners instead of separating them. The way to fix this error is to make sure you do not nest your listeners; it is **not** to use `emitter.setMaxListeners()` as the error suggests.
+
+You can debug these messages in different ways:
+- Through the [CLI](https://nodejs.org/api/cli.html#cli_trace_warnings): `node --trace-warnings index.js`
+- Through the [`process#warning` event](https://nodejs.org/api/process.html#process_event_warning): `process.on('warning', console.warn);`
+
+### Cannot send messages to this user.
+
+This error throws when the bot attempts to send a DM message to a user but cannot do so. A variety of reasons causes this:
```
WiseDevHelper, line 133:
```suggestion
This error throws when the bot attempts to send a DM message to a user but cannot do so. A variety of reasons may cause this error:
```

guide/keyv/README.md
```diff
@@ -101,7 +101,7 @@ client.on('message', async message => {
 			if (message.content.startsWith(guildPrefix)) prefix = guildPrefix;
 		}
 
-		// if we found a prefix, setup args; otherwise, this isn't a command
+		// if we found a prefix, setup args. Otherwise, this isn't a command
```
WiseDevHelper, line 104: Isn't this already correct?
```suggestion
		// if we found a prefix, setup args; otherwise, this isn't a command
```

guide/popular-topics/errors.md
```diff
@@ -120,6 +120,24 @@ In this particular example, you can see you are trying to access a piece of data
...
+- The bot and the user do not share a guild (often, people attempt to dm the user after kicking or banning them).
+- The bot tries to DM another bot.
+- The user has blocked the bot.
+- The user has disabled dms in the privacy settings.
+
+In the case of the last two reasons, the error is not preventable, as the Discord API does not provide a way to check if you can send a user a dm until you attempt to send one. The best way to handle this error is to add a `.catch()` where you try to dm the user and either ignore the rejected Promise or do what you want because of it.
```
WiseDevHelper, line 139: `dm` -> `DM` to enforce consistency.
```suggestion
- The bot and the user do not share a guild (often, people attempt to DM the user after kicking or banning them).
- The bot tries to DM another bot.
- The user has blocked the bot.
- The user has disabled DMs in the privacy settings.

In the case of the last two reasons, the error is not preventable, as the Discord API does not provide a way to check if you can send a user a DM until you attempt to send one. The best way to handle this error is to add a `.catch()` where you try to DM the user and either ignore the rejected Promise or do what you want because of it.
```

guide/popular-topics/faq.md
```diff
@@ -75,6 +75,23 @@ if (<message>.author.id === '<id>') {
...
+const code = args.join(' ');
+
+const evaluated = await eval(`( async () => {
+  ${code}
+})()`);
+```
+::: tip
+You can find a more detailed version of the eval command [here](https://anidiots.guide/examples/making-an-eval-command).
+:::
+
```
WiseDevHelper, line 94: Would need @Danktuary or @almostSouji's input on this one.

WiseDevHelper (changes requested):

guide/keyv/README.md
SinisterDeveloper: How would it be? Isn't that a bit incorrect grammar

SinisterDeveloper:
@WiseDevHelper Committed 2 changes which you suggested, although I'm not so sure about the `; otherwise` part

guide/keyv/README.md
```diff
@@ -63,10 +63,10 @@ Keyv exposes a familiar [Map](https://developer.mozilla.org/en-US/docs/Web/JavaS
 
 ## Application
 
-Although Keyv can assist in any scenario where you need key-value data, we will focus on setting up a per-guild prefix configuration using Sqlite.
+Although Keyv can assist in any scenario where you involve key-value data, we will focus on setting up a per-guild prefix configuration using Sqlite.
```
almostSouji, line 66: need is more specific than involve, and semantically correct here, no need to change

guide/keyv/README.md
almostSouji: initial was correct

guide/popular-topics/faq.md
almostSouji, line 94: no eval

almostSouji (changes requested):
Thanks for your contribution, but I don't agree with most of these changes.

- Some fixes aren't really fixes
- Eval is not going into the guide. It's dangerous and serves almost no use case, since most debugging use cases are better done with proper debugging tools. We also don't interlink other guides to not depend on them updating their content, if necessary.
- I don't agree with the new proposed order of errors, They were semantically grouped before

Generally speaking, additions of topics should be in their own PR so they can be handled and regarded separately from chores like new links and grammar. Grouping adding controversial content as "miscellaneous" is a bit of a stretch as well.

SinisterDeveloper:
Alright no problem. I will be closing this Pull Request. Have a good day

PR #711, opened 2021-07-14
feat: Sapphire guide

kyranet:

As discussed internally with the team, we are replacing Commando with Sapphire, and this takes the first step.

> **Note**: Please keep in mind that while current release is v1.0.2, this guide has been written for v2.0.0, which will release shortly after discord.js v13, and features a lot of abstractions oriented for more novice users.

Also, this is my first time contributing to this repository, so please let me know if I did something wrong. I'll also test the changes once I finish writing all the pages 😅

### DIFF
--- a/guide/sapphire/README.md
+++ b/guide/sapphire/README.md
@@ -1,0 +1,64 @@
+# Getting started with Sapphire
+
+When installing Sapphire Framework, you need to install both discord.js and the framework, you can do this by running the following command:
+
+```sh-session
+npm install discord.js @sapphire/framework
+```
+
+::: warning
+You need at least Node.js version 14.0 to use Sapphire.
+:::
+
+## Creating your index.js file
+
+While it doesn't have to be called `index.js`, this file is the main file for your bot, which will handle the bot's setup and login. You can also place it inside a folder, such as `src`.
+
+::: warning
+Make sure that the `main` property in your `package.json` points to the right path, such as `index.js` or `src/main.js`.
+:::
+
+The first thing you have to do is require Sapphire. Contrary to what you may think, you may **not** need to require discord.js in this file, and the Sapphire client extends discord.js's, so everything from Client is available in SapphireClient!
+
+```js
+const { SapphireClient } = require('@sapphire/framework');
+```
+
+The next step is to create a new SapphireClient, which is where you can customize Sapphire's behavior, as well as discord.js's. Just like in Client, you must define the `intents`:
+
+```js
+const client = new SapphireClient({ intents: ['GUILDS', 'GUILD_MESSAGES'] });
+```
+
+Sapphire uses mention prefix (`@bot command`), but you can optionally define a default prefix (or prefixes) by defining `defaultPrefix` in the options, as well as `regexPrefix`.
+
+You can also define the base directory Sapphire will scan directories at by defining a path to `baseUserDirectory`, this is an advanced option, and will always be set to the directory of the main file, defined in your package.json.
+
+By default, Sapphire will register the directory the main file is at, joined by the store's name, as such, if the root directory is `src`, it will register `src/commands` as one of the command directories.
+
+Last but certainly not least, log the bot in.
+
+```js
+client.login('your-token-goes-here');
+```
+
+::: danger
+You should use environment variables or a `config.json` for your token instead of passing it directly!
+:::
+
+And there you have it! You've set up your `index.js` file! In the end, your file structure should look like this, along with whatever `.gitignore` or `config.json` you may have:
+
+```
+node_modules/
+src/
+  ├── commands/
+  │     └── ping.js
+  ├── listeners/
+  └── main.js
+package-lock.json
+package.json
+```
+
+## Resulting code
+
+<ResultingCode />
--- a/guide/sapphire/README.md
+++ b/guide/sapphire/README.md
@@ -1,6 +1,6 @@
 # Getting started with Sapphire
 
-When installing the Sapphire Framework, you need to install both discord.js and the framework, you can do this by running the following command:
+When installing the Sapphire Framework, you need to install both discord.js and the framework. You can do this by running the following command:
 
 ```sh-session
 npm install discord.js @sapphire/framework
@@ -10,7 +10,7 @@ npm install discord.js @sapphire/framework
 You need at least Node.js version 14.0 to use Sapphire.
 :::
 
-## Creating your index.js file
+## Creating an index.js file
 
 While it doesn't have to be called `index.js`, this file is the main file for your bot, which will handle the bot's setup and login. You can also place it inside a folder, such as `src`.
 
@@ -18,52 +18,45 @@ While it doesn't have to be called `index.js`, this file is the main file for yo
 Make sure that the `main` property in your `package.json` points to the right path, such as `index.js` or `src/index.js`.
 :::
 
-The first thing you have to do is require Sapphire. Contrary to what you may think, you may **not** need to require discord.js in this file, and the Sapphire client extends discord.js's, so everything from Client is available in SapphireClient!
+To begin, require `@sapphire/framework` and create a new `SapphireClient`. This is where you can customize Sapphire's behavior, as well as discord.js'. The Sapphire client extends discord.js', so everything from `Client` is available in `SapphireClient`!
 
 :::: code-group
 ::: code-group-item CommonJS
 ```js
 const { SapphireClient } = require('@sapphire/framework');
+
+const client = new SapphireClient({ intents: ['GUILDS', 'GUILD_MESSAGES'] });
+
+client.login('your-token-goes-here');
 ```
 :::
 ::: code-group-item ESM
 ```js
 import { SapphireClient } from '@sapphire/framework';
+
+const client = new SapphireClient({ intents: ['GUILDS', 'GUILD_MESSAGES'] });
+
+client.login('your-token-goes-here');
 ```
 :::
 ::::
 
-The next step is to create a new SapphireClient, which is where you can customize Sapphire's behavior, as well as discord.js's. Just like in Client, you must define the `intents`:
 
-```js
-const client = new SapphireClient({ intents: ['GUILDS', 'GUILD_MESSAGES'] });
-```
+Sapphire uses mention prefix (`@bot command`), but you can optionally define a default prefix (or prefixes) with the `defaultPrefix` option, as well as `regexPrefix` if you are familiar with regexes.
 
-Sapphire uses mention prefix (`@bot command`), but you can optionally define a default prefix (or prefixes) by defining `defaultPrefix` in the options, as well as `regexPrefix` if you are familiar with regexes.
+There is also the advanced option `baseUserDirectory`, which allows you to define the base directory for Sapphire to scan. By default, Sapphire will register this as the directory where your `main` file is at, joined by the store's name. As such, if the root directory is `src`, it will register `src/commands` as one of the command directories.
 
-You can also define the base directory Sapphire will scan directories at by defining a path to `baseUserDirectory`, this is an advanced option, and will always be set to the directory of the main file, defined in your package.json.
-
-By default, Sapphire will register the directory the main file is at, joined by the store's name, as such, if the root directory is `src`, it will register `src/commands` as one of the command directories.
-
-Last but certainly not least, log the bot in.
-
-```js
-client.login('your-token-goes-here');
-```
 
 ::: danger
 You should use environment variables or a `config.json` for your token instead of passing it directly!
 You can read more about why you should [here](/preparations/setting-up-a-bot-application.html#keeping-your-token-safe).
 :::
 
-And there you have it! You've set up your `index.js` file! In the end, your file structure should look like this, along with whatever `.gitignore` or `config.json` you may have:
+And that's it for your `index.js` file! In the end, your file structure should look like this, along with whatever `.gitignore` or `config.json` files you may have:
 
-```
+```:no-line-numbers
 node_modules/
 src/
-  ├── commands/
-  │     └── ping.js
-  ├── listeners/
   └── index.js
 package-lock.json
 package.json
--- a/guide/sapphire/README.md
+++ b/guide/sapphire/README.md
@@ -15,7 +15,7 @@ You need at least Node.js version 14.0 to use Sapphire.
 While it doesn't have to be called `index.js`, this file is the main file for your bot, which will handle the bot's setup and login. You can also place it inside a folder, such as `src`.
 
 ::: warning
-Make sure that the `main` property in your `package.json` points to the right path, such as `index.js` or `src/main.js`.
+Make sure that the `main` property in your `package.json` points to the right path, such as `index.js` or `src/index.js`.
 :::
 
 The first thing you have to do is require Sapphire. Contrary to what you may think, you may **not** need to require discord.js in this file, and the Sapphire client extends discord.js's, so everything from Client is available in SapphireClient!
@@ -30,7 +30,7 @@ The next step is to create a new SapphireClient, which is where you can customiz
 const client = new SapphireClient({ intents: ['GUILDS', 'GUILD_MESSAGES'] });
 ```
 
-Sapphire uses mention prefix (`@bot command`), but you can optionally define a default prefix (or prefixes) by defining `defaultPrefix` in the options, as well as `regexPrefix`.
+Sapphire uses mention prefix (`@bot command`), but you can optionally define a default prefix (or prefixes) by defining `defaultPrefix` in the options, as well as `regexPrefix` if you are familiar with regexes.
 
 You can also define the base directory Sapphire will scan directories at by defining a path to `baseUserDirectory`, this is an advanced option, and will always be set to the directory of the main file, defined in your package.json.
 
@@ -44,6 +44,7 @@ client.login('your-token-goes-here');
 
 ::: danger
 You should use environment variables or a `config.json` for your token instead of passing it directly!
+You can read more about why you should [here](/preparations/setting-up-a-bot-application.html#keeping-your-token-safe).
 :::
 
 And there you have it! You've set up your `index.js` file! In the end, your file structure should look like this, along with whatever `.gitignore` or `config.json` you may have:
@@ -54,7 +55,7 @@ src/
   ├── commands/
   │     └── ping.js
   ├── listeners/
-  └── main.js
+  └── index.js
 package-lock.json
 package.json
 ```
--- a/guide/sapphire/README.md
+++ b/guide/sapphire/README.md
@@ -1,6 +1,6 @@
 # Getting started with Sapphire
 
-When installing Sapphire Framework, you need to install both discord.js and the framework, you can do this by running the following command:
+When installing the Sapphire Framework, you need to install both discord.js and the framework, you can do this by running the following command:
 
 ```sh-session
 npm install discord.js @sapphire/framework
--- a/guide/sapphire/README.md
+++ b/guide/sapphire/README.md
@@ -20,9 +20,18 @@ Make sure that the `main` property in your `package.json` points to the right pa
 
 The first thing you have to do is require Sapphire. Contrary to what you may think, you may **not** need to require discord.js in this file, and the Sapphire client extends discord.js's, so everything from Client is available in SapphireClient!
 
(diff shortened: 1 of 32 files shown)


alexthemaster:
Not to sound super obnoxious but this will be legendary for Discord.js

guide/sapphire/README.md
```diff
@@ -0,0 +1,64 @@
...
+By default, Sapphire will register the directory the main file is at, joined by the store's name, as such, if the root directory is `src`, it will register `src/commands` as one of the command directories.
+
+Last but certainly not least, log the bot in.
+
+```js
+client.login('your-token-goes-here');
+```
+
+::: danger
+You should use environment variables or a `config.json` for your token instead of passing it directly!
```
WiseDevHelper, line 46: A reference to [this page](https://discordjs.guide/preparations/setting-up-a-bot-application.html#keeping-your-token-safe) here maybe?

```suggestion
You should use environment variables or a `config.json` for your token instead of passing it directly!
You can read more about why you should [here](/preparations/setting-up-a-bot-application.html#keeping-your-token-safe).
```

guide/sapphire/README.md
```diff
@@ -0,0 +1,64 @@
...
+::: warning
+You need at least Node.js version 14.0 to use Sapphire.
+:::
+
+## Creating your index.js file
+
+While it doesn't have to be called `index.js`, this file is the main file for your bot, which will handle the bot's setup and login. You can also place it inside a folder, such as `src`.
+
+::: warning
+Make sure that the `main` property in your `package.json` points to the right path, such as `index.js` or `src/main.js`.
```
WiseDevHelper, line 18:
```suggestion
Make sure that the `main` property in your `package.json` points to the right path, such as `index.js` or `src/index.js`.
```

guide/sapphire/README.md
```diff
@@ -0,0 +1,64 @@
...
+const { SapphireClient } = require('@sapphire/framework');
+```
+
+The next step is to create a new SapphireClient, which is where you can customize Sapphire's behavior, as well as discord.js's. Just like in Client, you must define the `intents`:
+
+```js
+const client = new SapphireClient({ intents: ['GUILDS', 'GUILD_MESSAGES'] });
+```
+
+Sapphire uses mention prefix (`@bot command`), but you can optionally define a default prefix (or prefixes) by defining `defaultPrefix` in the options, as well as `regexPrefix`.
```
WiseDevHelper, line 33:
```suggestion
Sapphire uses mention prefix (`@bot command`), but you can optionally define a default prefix (or prefixes) by defining `defaultPrefix` in the options, as well as `regexPrefix` if you are familiar with regexes.
```

WiseDevHelper (changes requested):
Some small suggestions.

WiseDevHelper:
Additionally, I think the `Blurple` theme will look great for this section.

guide/sapphire/args.md
```diff
@@ -0,0 +1,163 @@
...
+	async run(message, args) {
+		const text = await args.pick('string');
+		await message.channel.send(text);
+	}
+};
+```
+
+The way Sapphire's arguments work, is by "consuming" the parameters one by one on success, similar to how an iterator works. We'll go into more details and examples with another command.
+
+In this command, we're using `args.pick` to read a single parameter from the user's input. The input type is `string`, which means it'll pick the current parameter and return it as-is.
```
Tylertron1998, line 43: `
```suggestion
In this command, we're using `args.pick()` to read a single parameter from the user's input. The input type is `string`, which means it'll pick the current parameter and return it as-is.
```

small change, but would help reinforce the fact this is a method - and makes it easier to locate what you mean in the code example above.

guide/sapphire/args.md
```diff
@@ -0,0 +1,163 @@
...
+
+If you run this with `@bot say Hello there!`, it'll send `Hello` back. What happened!? As I said before, each word is a parameter, so `Hello there!` is converted to `["Hello", "there!"]`, and `args.pick` reads only one. But fear not! There are alternatives:
```
Tylertron1998, line 45: same here (use brackets to show it's a method)

guide/sapphire/args.md
```diff
@@ -0,0 +1,163 @@
...
+
+- **Quoted parameters**:
+  If instead of sending `@bot say Hello there!`, we sent `@bot say "Hello there!"` or `@bot say “Hello there!”`, wrapping the content with quotes, the entire content inside the quotes will be read as a single parameter (without the quotes, of course).
+
+- **`args.rest` as-is**:
```
Tylertron1998, line 50: same here (use brackets to show it's a method)

guide/sapphire/args.md
```diff
@@ -0,0 +1,163 @@
...
+  };
+  ```
+
+  By default, `Command` defines 3 pairs of valid quotes:
+
+  - `"` and `"` (double quotes)
+  - `“` and `”` (iOS smart quotes)
+  - `「` and `」` (CJK corner brackets)
+
+  This is troublesome if we want to send quotes in the content, as those would be excluded from `rest`, so we have set quotes to an empty array. This way, when you do `@bot say Hello "there!"`, your bot will send `Hello "there!"`.
```
Tylertron1998, line 73: please put "quotes" in backticks -- to show it's "code" (i.e. it's a field of the class)

guide/sapphire/args.md
```diff
@@ -0,0 +1,163 @@
...
+		});
+	}
+
+	run(message, args) {
+		// ...
+	}
+};
+```
+
+Now let's fill the logic for the command inside the `run` method. Remember, we want to get two numbers, and add them, so we will need to run `args.pick` twice:
```
Tylertron1998, line 97: same as above (use brackets to indicate it's a method)

guide/sapphire/first-command.md
```diff
@@ -0,0 +1,80 @@
...
+
+Before you do anything, at the start of your file, you're going to need to require Sapphire Framework again, specifically its Command class.
+
+```js
+const { Command } = require('@sapphire/framework');
+```
+
+Commands are classes that are exported in any of the following ways:
+
+```js
```
Megatank58, line 17: Maybe this can have only commonjs as most parts of guide follow common js and having two of them *might* confuse the user on which to use

guide/sapphire/args.md
```diff
@@ -0,0 +1,163 @@
...
+		const a = await args.pick('number');
+		const b = await args.pick('number');
+		return message.channel.send(`The result is... ${a + b}!`);
+	}
+};
+```
+
+Now, if you run `@bot add 1 2`, the bot will reply with `The result is... 3!`. But what is going on in here? What is `a` and what is `b`?
+
+As mentioned before, `args.pick` consumes parameters one by one, so at first, we will have two parameters: `'1'` and `'2'`. When calling `args.pick`, the `'1'` is consumed and parsed using the name of the parameter, in this case `'number'`, so it's parsed as the number `1`. Then we do the same with the next parameter, which will be `'2'`. As a result, `a` is assigned to the number `1`, and `b` is assigned to the number `2`.
```
Tylertron1998, line 113: same as above (use brackets to indicate it's a method)

guide/sapphire/first-command.md
```diff
@@ -0,0 +1,80 @@
...
+// CommonJS
+module.exports = class MainExportedFoo extends Piece {};
+exports.default = class DefaultExportedFoo extends Piece {};
+exports.myCommand = class CustomExportedFoo extends Piece {};
+
+// ECMAScript Modules
+export default class ExportedDefaultFoo extends Piece {}
+export class ExportedFoo extends Piece {}
+```
```
Megatank58, line 26: Maybe this should only have the CommonJS way as that is used mostly through the guide and the user might be confused on which to choose?

kyranet: @Danktuary @almostSouji Sapphire's loader supports both CJS and ESM (even mixed within the same project), as well as it supports multiple exports within the same file, which differs from Commando, so I don't really have a good reference at how should we display the possibilities.

Furthermore, all of Sapphire (as well as discord.js) support ESM exports, so the ESM integration here is full.

Any suggestions regarding how to make this less confusing/easier to understand?

Danktuary: For now, use the built-in `CodeGroup` and `CodeGroupItem` components: https://vuepress2.netlify.app/reference/default-theme/markdown.html#custom-containers

They're not used anywhere else yet, so see how well it works with ESLint (e.g. test with a missing semicolon). If it doesn't lint, try adding newlines before and after the codeblocks.

kyranet: It lints correctly, and it also works beautifully, thanks!
image

YorkAARGH (changes requested):
Nothing major, just a couple of inconsistencies I found whilst following these guides, so far they're good if an idiot like me can follow them!

guide/sapphire/first-command.md
```diff
@@ -0,0 +1,94 @@
...
+
+Don't let this scare you; it's straightforward.
+
+- `context` is an object that contains some file metadata required by the `Command` and `Piece` (which `Command` extends) require in order to function.
+- `name` is, by default, the name of the file without the extension, e.g. `ping.js` becomes `ping`, so we don't need to define it.
+- `aliases` are other ways users can call the command. You can have as many as you want!
+- `description` is the help text displayed when somebody uses a help command.
+
+There are many more properties you can use, but their sections will explain those.
```
WiseDevHelper, line 62: Something like:
```suggestion
There are many more properties you can use. The upcoming sections will explain those.
```

guide/sapphire/args.md
kyranet: What do you mean?

Tylertron1998: "This is troublesome if we want to send quotes in the content" -> " This is troublesome if we want to send `quotes` in the content"

guide/sapphire/first-command.md
```diff
@@ -0,0 +1,94 @@
...
+```
+
+Don't let this scare you; it's straightforward.
+
+- `context` is an object that contains some file metadata required by the `Command` and `Piece` (which `Command` extends) require in order to function.
+- `name` is, by default, the name of the file without the extension, e.g. `ping.js` becomes `ping`, so we don't need to define it.
+- `aliases` are other ways users can call the command. You can have as many as you want!
+- `description` is the help text displayed when somebody uses a help command.
+
+There are many more properties you can use. The upcoming sections will explain those.
```
Danktuary, line 62:
```suggestion
An overview of what's defined in the constructor:

- `context`: an object that contains file metadata required by the `Command` and `Piece` classes (which `Command` extends) in order to function.
- `name`: by default, the name of the file without the extension, i.e. `ping.js` becomes `ping`, so there's no need to define it.
- `aliases`: other ways users can call the command. You can have as many as you want!
- `description`: the text displayed when the help command is used.

There are many other properties available, explained in upcoming sections.
```

guide/sapphire/first-command.md
```diff
@@ -0,0 +1,94 @@
...
+
+## Creating your run method
+
+The next thing you're going to need is a `run` method. This should go right below the command's constructor. Inside, you'll run the command's logic:
```
Danktuary, line 66:
```suggestion
## Creating the `run` method

Commands have a `run` method, which executes the command logic. Define this below the command's constructor:
```

guide/sapphire/first-command.md
```diff
@@ -0,0 +1,94 @@
...
+
+<!-- eslint-disable constructor-super -->
+
+```js {6-8}
```
Danktuary, line 70:
```suggestion
```js {7-9}
```

guide/sapphire/first-command.md
```diff
@@ -0,0 +1,94 @@
...
+:::
+::: code-group-item ESM
+```js
+export default class DefaultFoo extends Piece {}
+export class Foo extends Piece {}
+```
+:::
+::::
+
+Because its spread usage, we will be using `module.exports` in all the guides, so let's create the class and set `module.exports` to it!
```
Danktuary, line 42:
```suggestion
`module.exports` will be used for these pages. Let's begin populating the command class:
```

guide/sapphire/first-command.md
```diff
@@ -0,0 +1,94 @@
...
+};
+```
+
+<!-- eslint-enable constructor-super -->
+
+As you can see, the `run` method is simply the code you want the bot to run when somebody uses the command. This code can be anything you can do in core discord.js, as Sapphire Framework is simply an extension.
+
+<!-- TODO(kyranet): Add a mention for editable commands once I make the plugin for it. -->
+
+Now fire up the bot as usual and use your command! It should automatically be `@bot ping` to use it.
```
Danktuary, line 90:
```suggestion
Any discord.js code can be executed here, since Sapphire Framework is an extension. The command can be triggered with `@bot ping` or `@bot pong`.

<!-- TODO(kyranet): Add a mention for editable commands once I make the plugin for it. -->
```

Danktuary (changes requested):

guide/sapphire/args.md
```diff
@@ -0,0 +1,163 @@
...
+		const a = await args.pick('number');
+		const b = await args.pick('number');
+		return message.channel.send(`The result is... ${a + b}!`);
+	}
+};
+```
+
+Now, if you run `@bot add 1 2`, the bot will reply with `The result is... 3!`. But what is going on in here? What is `a` and what is `b`?
+
+As mentioned before, `args.pick()` consumes parameters one by one, so at first, we will have two parameters: `'1'` and `'2'`. When calling `args.pick()`, the `'1'` is consumed and parsed using the name of the parameter, in this case `'number'`, so it's parsed as the number `1`. Then we do the same with the next parameter, which will be `'2'`. As a result, `a` is assigned to the number `1`, and `b` is assigned to the number `2`.
```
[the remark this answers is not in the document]
kyranet: Yeah, that reads A LOT better, thank you 😅

guide/sapphire/built-in-arguments.md
```diff
@@ -0,0 +1,35 @@
...
+- `guildThreadChannel`: Resolves to a `ThreadChannel` instance given a thread channel mention or ID.
+- `guildVoiceChannel`: Resolves to a `VoiceChannel` instance given a voice channel mention or ID.
+- `hyperlink` / `url`: Resolves to a `URL` instance given a valid URL.
+- `integer`: Resolves to a safe integer.
+- `member`: Resolves to a `GuildMember` instance given a member mention or ID.
+- `message`: Resolves to a `Message` instance given a message link or ID.
+- `number`: Resolves to any number.
+- `role`: Resolves to a `Role` instance given a role mention or ID.
+- `string`: Resolves to a string containing the parameter.
+- `user`: Resolves to a `User` instance given a user mention or ID.
```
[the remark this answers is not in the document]
kyranet: Nice catch on that `number` entry, thanks!

guide/sapphire/built-in-arguments.md
```diff
@@ -0,0 +1,35 @@
...
+
+:::tip
+They can be used anytime using the `Args` class, and can be accessed via `container.stores.get('arguments')`.
```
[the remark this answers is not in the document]
kyranet: Is the "the" omissible here? It doesn't ring in my head as well as "with the `Args` class".

guide/sapphire/first-command.md
```diff
@@ -0,0 +1,94 @@
...
+
+:::: code-group
+::: code-group-item CommonJS
+```js
+const { Command } = require('@sapphire/framework');
+```
+:::
+::: code-group-item ESM
+```js
+import { Command } from '@sapphire/framework';
```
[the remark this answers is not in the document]
kyranet: I think ESLint complains due to duplicated identifiers in here, since exported members in ESM are accessible with their name, unlike in CommonJS.

guide/sapphire/first-command.md
kyranet: Perhaps "required by the `Command` and `Piece` classes (which `Command` extends)" should be a little rewritten here, as the note in parenthesis is meanat to be next to "Piece", unless that's valid.

guide/sapphire/first-command.md
kyranet: Shouldn't this be "since Sapphire Framework is an extension **of it**"? Or is it redundant?

WiseDevHelper: > Shouldn't this be "since Sapphire Framework is an extension **of it**"? Or is it redundant?
I think both are correct, but this is better?

guide/sapphire/first-command.md
WiseDevHelper: So,
```suggestion
An overview of what's defined in the constructor:
`context`: an object containing file metadata that the `Command` and `Piece` (which is extended by `Command`) classes require to function.
- `name`: by default, the name of the file without the extension, i.e. `ping.js` becomes `ping`, so there's no need to define it.
- `aliases`: other ways users can call the command. You can have as many as you want!
- `description`: the text displayed when the help command is used.
There are many other properties available, explained in upcoming sections.
```
?

kyranet: Maybe "which is extended by the former"?

guide/sapphire/built-in-arguments.md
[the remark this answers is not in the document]
Danktuary: You're right, updated my suggestion

guide/sapphire/first-command.md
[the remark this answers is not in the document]
Danktuary: I suppose the other exports could be commented out, leaving only the default export, or an `<!-- eslint-skip -->` could be used.

guide/sapphire/first-command.md
Danktuary: I updated it to "an object that contains file metadata required by the `Piece` class (which `Command` extends)". This seems more straightforward and less wordy

guide/sapphire/first-command.md
Danktuary: You're right, updated my suggestion

guide/sapphire/first-command.md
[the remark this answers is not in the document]
kyranet: Commenting out also works, in a way it also shows the other possibilities. 👍🏼

guide/sapphire/args.md
```diff
@@ -0,0 +1,147 @@
...
+		return message.channel.send(`The result is ${a + b}!`);
+	}
+};
+```
+
+Each time `args.pick('<type>')` is called, it consumes the arguments one by one and parses them as the `<type>` given. So `@bot add 1 2` would return `The result is 3!`.
+
+## Optional arguments
+
+Sapphire offers two options regarding regarding optional arguments. Let's make a `pow.js` command that takes two numbers, where the `base` is required, and the `exponent` defaults to 2.
```
muhimur9049, line 103:
```suggestion
Sapphire offers two options regarding optional arguments. Let's make a `pow.js` command that takes two numbers, where the `base` is required, and the `exponent` defaults to 2.
```

guide/sapphire/args.md
```diff
@@ -0,0 +1,147 @@
...
+  		const base = await args.pick('number');
+  		const exponent = args.finished ? 2 : await args.pick('number');
+  		return message.channel.send(`The result is ${Math.pow(base, exponent)}!`);
+  	}
+  };
+  ```
+
+  This is slightly more verbose, but it defaults `exponent` to `2` **only** if there were no parameters left to parse, so `@bot pow 4` will return `The result is 16!`, while `@bot pow 4 hello!` will return an error.
+
+Each pattern has its use-cases. The former is specially helpful for handling several optional parameters, although you can also discriminate all errors from `args` by checking `error.identifier`. This way you can tell if a number failed to parse for reasons such as it's not a number, it doesn't meet the minimum and/or maximum (if set), among other possible options.
```
muhimur9049, line 135:
```suggestion
Each pattern has its use-cases. The former is especially helpful for handling several optional parameters, although you can also discriminate all errors from `args` by checking `error.identifier`. This way you can tell if a number failed to parse for reasons such as it's not a number, it doesn't meet the minimum and/or maximum (if set), among other possible options.
```
Yes, `specially` can be used in many different cases. However, it is being incorrectly used in this case, so it's better to use `especially`.

almostSouji:
A Sapphire guide can now be found at https://www.sapphirejs.dev/docs/Guide/getting-started/getting-started

PR #712, opened 2021-07-14
fix(ShardingManager): renamed `launch` event to `shardCreate`

Scrip7:
The `ShardingManager` EventEmitter does not have an event called `launch`,
so I replaced it with the correct one which is `shardCreate`.

https://github.com/discordjs/discord.js/blob/master/src/sharding/ShardingManager.js#L163


WiseDevHelper:
Hello there,
The file you have changed is for `v11` which does have the [`launch` event](https://discord.js.org/#/docs/main/v11/class/ShardingManager?scrollTo=e-launch).

PR #727, opened 2021-07-25
fix(sidebar): remove outdated voice-related links

J-Human:
Except the 1st one, all of these result to a 404 because of the voice rewrite PR making way for a dedicated section for voice. It removes the outdated sidebar links.

(I realized this was committed directly to `v13-prep` on my end so I'll re-fork again when needed.)

### DIFF
--- a/guide/.vuepress/sidebar.ts
+++ b/guide/.vuepress/sidebar.ts
@@ -159,11 +159,6 @@ export default {
 			text: 'Voice',
 			children: [
 				'/voice/',
-				'/voice/understanding-voice.md',
-				'/voice/the-basics.md',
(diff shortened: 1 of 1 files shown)


guide/.vuepress/sidebar.ts
```diff
@@ -159,11 +159,6 @@ export default {
 			text: 'Voice',
 			children: [
```
Danktuary, line 160: This entire object should be removed, since there's a nav bar button for the voice pages now

J-Human: Ok.

J-Human:
Perfect.

PR #744, opened 2021-08-01
fix: async function in advanced deployment example

SuperchupuDev:
Changes the advanced deployment code in the guide to use an async function, since TLA only works in es6 modules, and it would throw an error otherwise, and also for consistency with discordjs/discord.js#6250

### DIFF
--- a/guide/interactions/registering-slash-commands.md
+++ b/guide/interactions/registering-slash-commands.md
@@ -188,18 +188,20 @@ const commands = client.commands.map(({ execute, ...data }) => data);
 
 const rest = new REST({ version: '9' }).setToken(token);
 
-try {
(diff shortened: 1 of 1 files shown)


guide/interactions/registering-slash-commands.md
```diff
@@ -188,18 +188,20 @@ const commands = client.commands.map(({ execute, ...data }) => data);
...
+		await rest.put(
+			Routes.applicationGuildCommands(CLIENT_ID, GUILD_ID),
+			{ body: commands },
+		);
+
+		console.log('Sucessfully reloaded application (/) commands.');
+	} catch (error) {
+		console.error(error);
+	}
+})()
```
Danktuary, line 204:
```suggestion
})();
```

PR #721, opened 2021-07-19
feat: v13 `Code Samples`.

WiseDevHelper:
**Please describe the changes this PR makes and why it should be merged:**

_(Split from #708)_

[Demo](https://deploy-preview-721--discordjs-guide.netlify.app/popular-topics/permissions.html)

This PR updates the `Code Samples` of the guide for the `v13` release. It also updates other files to maintain consistency.

**Status**
- [x] Update `Code Samples`.
  - [x] Additional Info.
    - [x] Rest API.
  - [x] Command Handling. 
    - [x] ~~Adding features.~~ (Removed. Refer https://github.com/discordjs/guide/pull/683#issuecomment-866417742)
    - [x] File setup.
  - ~~Commando~~ (To be removed in #711) 
  - [x] Creating your bot.
    - [x] Getting your bot up and running.
    - [x] Configuration files.
    - [x] Adding more commands. 
    - [x] ~~Commands with user input.~~ (Removed. Refer https://github.com/discordjs/guide/pull/683#issuecomment-866417742)
  - [x] Event Handling.
  - [x] Keyv.
  - [x] Miscellaneous.
    - [x] Parsing mention arguments.
  - OAuth. (Does not require updating)
  - [x] Popular topics.
    - [x] Canvas.
    - [x] Permissions. (#705)
    - [x] Reactions. (#729)
    - [x] Webhooks. (#700)
  - [x] Sequelize.
    - [x] Storing data.
    - [x] Currency system. 
  - [x] Sharding. (#695)

Please let me know if there is more to change. I'll be happy to work on them.

PR #752, opened 2021-08-03
chore: Replace intents + permission strings with flags

PenPow:
This PR replaces all instances of strings being used for permission checks with the respective static flag.

I updated the `Gateway Intents` and `Permissions` pages, along with the resulting code section for the latter.

Let me know of any changes or suggestions. I was unsure of whether I should show
```js
const { Permissions } = require('discord.js');
```
at the top of every codeblock where I used permissions (and the same for intents), but in the end I did show requiring it. If this should be changed please let me know.

Intent Flags were done in #721 but whether the Gateway Intents page being missed or whether it was intentional is anyone's guess, so I added them there too.


code-samples/popular-topics/permissions/13/index.js
```diff
@@ -11,22 +11,22 @@ client.on('interactionCreate', interaction => {
 	if (!interaction.isCommand()) return;
 	if (!interaction.channel.permissionsFor(client.user).has('SEND_MESSAGES')) return;
```
Danktuary, line 12:
```suggestion
	if (!interaction.channel.permissionsFor(client.user).has(Permissions.FLAGS.SEND_MESSAGES)) return;
```

code-samples/popular-topics/permissions/13/index.js
```diff
@@ -11,22 +11,22 @@ client.on('interactionCreate', interaction => {
...
 
-	const botPerms = ['MANAGE_MESSAGES', 'KICK_MEMBERS', 'MANAGE_ROLES', 'MANAGE_CHANNELS'];
+	const botPerms = [Permissions.FLAGS.MANAGE_MESSAGES, Permissions.FLAGS.KICK_MEMBERS, Permissions.FLAGS.MANAGE_ROLES, Permissions.FLAGS.MANAGE_CHANNELS];
```
Danktuary, line 14:
```suggestion
	const botPerms = [
		Permissions.FLAGS.MANAGE_MESSAGES,
		Permissions.FLAGS.KICK_MEMBERS,
		Permissions.FLAGS.MANAGE_ROLES,
		Permissions.FLAGS.MANAGE_CHANNELS,
	];
```
Lint because it becomes long for a single line

Danktuary (commented):
Just these, the rest all looks good. Thanks for the PR! 🎉

PR #757, opened 2021-08-04
Update embeds.md

akhmed-codes:
Fixed markdown display of link


Danktuary:
The syntax currently used (`[Text](https://site.com 'hover text')`) is correct; removing an apostrophe would be invalid.

PR #759, opened 2021-08-04
fix: Implement environment variables in config file

SinisterDeveloper:

I feel the Algolia API key should not be a part of the main code, pretty sure Algolia has a limit for the number of requests per month. I don't think Google Analytics Id should be a part of code either.

This would require the guide to add  Netlify environment variables for the Algolia API key as well as Google Analytics ID

### DIFF
--- a/guide/.vuepress/config.ts
+++ b/guide/.vuepress/config.ts
@@ -54,7 +54,7 @@ if (process.env.NODE_ENV === 'production') {
 		[
 			'@vuepress/plugin-docsearch',
 			{
-				apiKey: process.env.API_KEY,
+				apiKey: process.env.ALGOLIA_DOCSEARCH_API_KEY,
 				indexName: 'discordjs',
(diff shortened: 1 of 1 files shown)


SinisterDeveloper:
Its a small pull request, not many changes, however I thought it would be more ideal to have the environmental is their correct place

guide/.vuepress/config.ts
```diff
@@ -54,14 +54,14 @@ if (process.env.NODE_ENV === 'production') {
 		[
 			'@vuepress/plugin-docsearch',
 			{
-				apiKey: 'c8d9361fb8403f7c5111887e0edf4b5e',
+				apiKey: process.env.API_KEY,
```
Danktuary, line 57:
```suggestion
				apiKey: process.env.ALGOLIA_DOCSEARCH_API_KEY,
```

PR #761, opened 2021-08-04
Use the dev version

theblobscp:
This will properly install the version 13


Megatank58:
The guide is for v13 not dev version so after v13 releases this would be needed to changed again

theblobscp:
True, but for now will it still be outdated?

PenPow:
> True, but for now will it still be outdated?

Yes, but this guide is for V13 when it releases, which hopefully _fingers crossed_should be fairly soon (all dependant on discord). There is no reason to make a PR changing this, when it will shortly be reverted.

theblobscp:
Ok, I see your point

PR #769, opened 2021-08-05
fix: Change interaction handler sample code

SinisterDeveloper:

I think this is a much better way to execute the Slash Commands. Instead of reading collection twice, this is a more simplified way


guide/command-handling/README.md
```diff
@@ -107,10 +107,12 @@ With your `client.commands` Collection setup, you can use it to retrieve and exe
 client.on('interactionCreate', async interaction => {
 	if (!interaction.isCommand()) return;
 
-	if (!client.commands.has(interaction.commandName)) return;
+	let command = client.commands.get(interaction.commandName);
```
DaStormer, line 110:
```suggestion
	const command = client.commands.get(interaction.commandName);
```
Use `const` since `command` isn't being re-assigned.

DaStormer (changes requested):
To be honest, I think the current way is fine, but if we go with this, the paragraph under the code block needs to be adjusted accordingly.

Also, this messes up the code block line highlighting, but I can't suggest a change since it's outside the diff.

guide/command-handling/README.md
SinisterDeveloper: Alright

PenPow (approved):

guide/command-handling/README.md
```diff
@@ -107,10 +107,12 @@ With your `client.commands` Collection setup, you can use it to retrieve and exe
 client.on('interactionCreate', async interaction => {
```
WiseDevHelper, line 107: The line above should be 
```
```js {4-13}
```

WiseDevHelper (changes requested):

guide/command-handling/README.md
SinisterDeveloper: Alright will fix

almostSouji (changes requested):
- [x]  trailing spaces as per eslint
- [x]  text needs to be adapted, order of operations isn't accurate anymore

WiseDevHelper (approved):

WiseDevHelper:
This needs(?) a rebase.

guide/command-handling/README.md
```diff
@@ -107,28 +107,28 @@ for (const file of commandFiles) {
...
 	}
 });
 ```
 
-If there isn't a command with that name, you don't need to do anything further, so exit early with `return`. If there is, `.get()` the command, call its `.execute()` method, and pass in your `interaction` variable as its argument. In case something goes wrong, log the error and report back to the member to let them know.
+First, fetch the command in the Collection with that name and assign it to the variable `command`. If the command doesn't exist, it will return `undefined` and you exit early with `return`. If it does exist, call the command's `.execute()` method, and pass in your `interaction` variable as its argument. In case something goes wrong, log the error and report back to the member to let them know.
 
-And that's it! Whenever you want to add a new command, you make a new file in your `commands` directory, name it what you want, and then do what you did for the other commands.
+And that's it! Whenever you want to add a new command, you [register a command](/interactions/registering-slash-commands.md), make a new file in your `commands` directory, name it with what you did for the Slash Command, and then do what you did for the other commands.
```
Danktuary, line 131:
```suggestion
And that's it! Whenever you want to add a new command, you [register a command](/interactions/registering-slash-commands.md), make a new file in your `commands` directory, name it the same as the slash command, and then do what you did for the other commands.
```

Danktuary (commented):
`/code-samples/command-handling/file-setup/13/index.js` needs to be updated to reflect the code changes

Danktuary:
Since I'll be rewriting this page in #826, I'll merge this now. Thanks for the PR. 😄

PR #799, opened 2021-08-11
I think you meant or

A7mooz:


Danktuary:
"nor" is correct here, it's being used with "cannot"

PR #809, opened 2021-08-13
Add explicit variable definitions for client and guild id

Renal-Of-Loon:
This PR aims to add the definition of GUILD_ID and CLIENT_ID to registering slash commands.

Although these could just be hardcoded in the relevant place, the (faulty) half definitions suggested would present a more explicit "do thing here" to the users.

If it is scanned over and ignored, the error should (I would hope) be pretty obvious to anyone who encounters it.

### DIFF
--- a/guide/interactions/registering-slash-commands.md
+++ b/guide/interactions/registering-slash-commands.md
@@ -27,6 +27,10 @@ const fs = require('fs');
 const commands = [];
 const commandFiles = fs.readdirSync('./commands').filter(file => file.endsWith('.js'));
 
+// Define values used for slash command registration
+const CLIENT_ID = ; // Place your client/bot id here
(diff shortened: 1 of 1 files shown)


guide/interactions/registering-slash-commands.md
```diff
@@ -27,6 +27,10 @@ const fs = require('fs');
 const commands = [];
 const commandFiles = fs.readdirSync('./commands').filter(file => file.endsWith('.js'));
 
+// Define values used for slash command registration
+const clientId = '12345678'; // Place your client/bot id here
+const guildId = '12345678'; // Place your guild id here
```
Danktuary, line 32:
```suggestion
// Place your client and guild ids here
const clientId = '123456789012345678';
const guildId = '876543210987654321';
```

PR #813, opened 2021-08-14
fix: add missing `}` in diff

QSmally:
A curly bracket was missing in the change, as there are two opening-brackets yet only one closing bracket


hellobafu:
Shouldn't there be a space behind it?

QSmally:
> Shouldn't there be a space behind it?

Yeah, you're right - my bad

kyranet (approved):

PR #827, opened 2021-08-17
Make await explanation accurate

cubedhuang:
This comment previously said "Since the image takes time to load, you should await it," but that's not what await actually does - I changed it to be more accurate.

### DIFF
--- a/guide/popular-topics/canvas.md
+++ b/guide/popular-topics/canvas.md
@@ -90,7 +90,7 @@ client.on('interactionCreate', async interaction => {
 	// ...
 	const context = canvas.getContext('2d');
 
-	// Since the image takes time to load, you should await it
+	// Using Canvas.loadImage to get the background returns a promise, so it should be awaited
 	const background = await Canvas.loadImage('./wallpaper.jpg');
 
 	// This uses the canvas dimensions to stretch the image onto the entire canvas
--- a/guide/popular-topics/canvas.md
+++ b/guide/popular-topics/canvas.md
@@ -90,7 +90,7 @@ client.on('interactionCreate', async interaction => {
 	// ...
 	const context = canvas.getContext('2d');
 
-	// Using Canvas.loadImage to get the background returns a promise, so it should be awaited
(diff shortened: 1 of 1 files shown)


monbrey:
Both are accurate for different reasons. async operations return Promises, because they take time to complete.

guide/popular-topics/canvas.md
```diff
@@ -90,7 +90,7 @@ client.on('interactionCreate', async interaction => {
 	// ...
 	const context = canvas.getContext('2d');
 
-	// Since the image takes time to load, you should await it
+	// Using Canvas.loadImage to get the background returns a promise, so it should be awaited
```
Danktuary, line 93:
```suggestion
	// Use node-canvas's `loadImage()` utility method to load images
```
`async`/`await` and `.then()`/`.catch()` is used all around the guide, so we don't need to keep mentioning "hey, Promises needs to be `await`ed". Since the `loadImage()` function is specific to the node-canvas package and not the native Canvas spec, the comment should reflect this instead.

An alternative to this approach: remove the comment entirely and mention this (with a [link](https://github.com/Automattic/node-canvas#loadimage)) in the paragraph above this codeblock.

wizo06: Alternatively, we could include a resource link in the introduction page that explains how promises work. Then, we can simply mention something like "if you don't know how the `await` keyword works, refer to <link>" across the guide.

WiseDevHelper: We already have a whole page on `async`, `await` and Promises [which you can find here.](https://discordjs.guide/additional-info/async-await.html#how-do-promises-work)

wizo06: 👍🏻 perfect. we can use that page as reference instead then. From what i understand, the guide is meant to teach how to use the library, not JS. So it makes perfect sense to simply use the `await` keyboard in our examples without explaining much, and then include a little note for something like "refer to this <link> if you don't know how `await` keyword works".

PR #830, opened 2021-08-18
fix(interactions): Remove Number Option

soyabn09:

`addNumberOption` and `getNumber` are no longer methods on `@discordjs/builders`


soyabn09:
There is this but not merged yet. https://github.com/discordjs/builders/pull/23

adrifcastr:
> `addNumberOption` and `getNumber` are no longer methods on `@discordjs/builders`

This is incorrect as those methods weren't ever there so they cannot no longer be there. PR 23 that adds this is open and thus the corresponding PR on the guide was also opened, but merged prematurely.

Danktuary:
Since PR 23 has two approvals and seems like it will be merged, there's no need to remove this.

PR #833, opened 2021-08-22
feat(rest-api): use native URLSearchParams class for Urban Dictionary sample

J-Human:
See https://nodejs.org/api/querystring.html. The query-string module is considered legacy and therefore not recommended to use, even if it has not yet been deprecated. The `URLSearchParams` class can also achieve the same thing in this context and is tested to work (v16.6.x).

### DIFF
--- a/code-samples/additional-info/rest-api/13/index.js
+++ b/code-samples/additional-info/rest-api/13/index.js
@@ -20,7 +20,7 @@ client.on('interactionCreate', async interaction => {
 		interaction.reply({ files: [file] });
 	} else if (commandName === 'urban') {
 		const term = interaction.options.getString('term');
-		const query = new URLSearchParams({ term }).toString();
+		const query = new URLSearchParams({ term });
 
 		const { list } = await fetch(`https://api.urbandictionary.com/v0/define?${query}`).then(response => response.json());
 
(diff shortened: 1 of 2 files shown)


code-samples/additional-info/rest-api/13/index.js
```diff
@@ -21,7 +20,7 @@ client.on('interactionCreate', async interaction => {
 		interaction.reply({ files: [file] });
 	} else if (commandName === 'urban') {
 		const term = interaction.options.getString('term');
-		const query = querystring.stringify({ term });
+		const query = new URLSearchParams({ term }).toString();
```
Danktuary, line 23:
```suggestion
		const query = new URLSearchParams({ term });
```

guide/additional-info/rest-api.md
```diff
@@ -94,21 +94,20 @@ Urban Dictionary's API is available at https://api.urbandictionary.com/v0/define
...
 
 ```js {1,5-11}
-const querystring = require('querystring');
 // ...
 	// ...
 	if (commandName === 'urban') {
 		const term = interaction.options.getString('term');
-		const query = querystring.stringify({ term });
+		const query = new URLSearchParams({ term }).toString();
```
Danktuary, line 102:
```suggestion
		const query = new URLSearchParams({ term });
```

Danktuary (commented):
`.toString()` isn't necessary, since `query` is inside a template literal

PR #846, opened 2021-08-29
refactor: production build uses webpack and shiki

DET171:
Webpack can bundle code that targets older browsers while vite cannot. Shiki also has higher fidelity than prismjs, as mentioned in the VuePress docs.
P.S. The VuePress 2 docs also use the same configuration as stated above.

### DIFF
--- a/guide/.vuepress/config.ts
+++ b/guide/.vuepress/config.ts
@@ -2,9 +2,12 @@ import path from 'path';
 import { defineUserConfig } from 'vuepress-vite';
 import type { DefaultThemeOptions, ViteBundlerOptions } from 'vuepress-vite';
 import sidebar from './sidebar';
+const { ALGOLIA_DOCSEARCH_API_KEY, GOOGLE_ANALYTICS_ID, NODE_ENV } = process.env;
+
+const isProd = NODE_ENV === 'production';
 
 const config = defineUserConfig<DefaultThemeOptions, ViteBundlerOptions>({
-	bundler: '@vuepress/vite',
+	bundler: isProd ? '@vuepress/webpack' : '@vuepress/vite',
 	templateDev: path.join(__dirname, 'templates', 'index.dev.html'),
 	templateSSR: path.join(__dirname, 'templates', 'index.ssr.html'),
 	lang: 'en-US',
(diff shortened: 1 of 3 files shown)


DET171:
Removed line numbers for code blocks as well as it looks wrong and bad to on mobile (the line numbers are already a bit off if the code blocks is a few hundred lines long).
Let’s wait for VuePress to fix this first?

almostSouji:
I have absolutely no idea what you mean by 
> it looks wrong and bad to on mobile
Looks good on my end

DET171:
This:
B69B1F22-1303-45DD-91A1-A265B314FB3F

PenPow:
I can't reproduce on Android (Longer lines just require horizontal scroll).

However, I can reproduce on IOS (iPhone 8).

Danktuary:
@DET171 @PenPow Could you guys provide your iOS and Safari versions?

Line numbers are a feature we'd like to keep; I'll try to implement a fix for iOS/Safari.

PenPow:
> Could you guys provide your iOS and Safari versions?

IOS Version 14.7.1 (Safari version ~= 14, as IOS doesn't give a precise version)

DET171:
> Could you guys provide your iOS and Safari versions?
iOS 14.7.1
Safari 14 as well

DET171:
Helloooo?

Danktuary:
I've given it some thought and decided that this probably isn't necessary.

Vite targets browsers with [esmodule](https://caniuse.com/es6-module)/[dynamic import](https://caniuse.com/es6-module-dynamic-import) compatibility, which is about 93% global according to caniuse, which includes browser versions from 3 years ago or so.
If browser compatibility is a concern, [@vitejs/plugin-legacy](https://github.com/vitejs/vite/tree/main/packages/plugin-legacy) could be included, but to me, this is fine for the guide's purpose and target audience.

PrismJS seems to work fine for us too, as far as I've seen. Shiki doesn't seem needed, and also renders some things differently (e.g. tab indents as 8 spaces).

---

I've pushed a commit to try to fix word wrapping in `pre` tags for Safari browsers (e0c61960e5878d2476e516be32151f187498a792). If it's still bugging for you, let me know.

PR #851, opened 2021-09-01
Added Intents.FLAGS.GUILD_MESSAGES as an option

GNVageesh:
Added `Intents.FLAGS.GUILD_MESSAGES` as an intent option so that we can send messages through the BOT

With only `Intents.FLAGS.GUILD`, the BOT isn't able to send messages back to the Server. Adding the `Intents.FLAGS.GUILD_MESSAGES` as an option intent gets the job done

Kindly consider this change,
Thanking you
@GNVageesh

### DIFF
--- a/guide/creating-your-bot/command-handling.md
+++ b/guide/creating-your-bot/command-handling.md
@@ -10,7 +10,7 @@ Here are the base files and code we'll be using:
 const { Client, Intents } = require('discord.js');
 const { token } = require('./config.json');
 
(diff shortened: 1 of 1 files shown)


J-Human:
This isn't needed, the bot doesn't send messages; it uses slash commands. And the statement regarding not being able to send messages without the intent is false, we *can* send messages but we won't *receive* them.

GNVageesh:
oh..ok sir... sorry for pull request

PR #864, opened 2021-09-09
fix: outdated code in async-await section and reactions page

Megatank58:
**Please describe the changes this PR makes and why it should be merged:**
Changed all `interaction.reply(string, { fetchReply: true })` => `interaction.reply({ content: string, fetchReply: true })`
Changed `replyMessage.delete({ timeout: 10000 })` => `setTimeout(() => replyMessage.delete(), 10000)`

PR #865, opened 2021-09-09
fix(reactions): fix interaction.reply() params

BaumianerNiklas:

In the `popular-topics/reactions` section, there are multiple instances of `interaction.reply()` that wrongly use 2 parameters instead of just one. This PR fixes those.

### DIFF
--- a/guide/popular-topics/reactions.md
+++ b/guide/popular-topics/reactions.md
@@ -60,7 +60,7 @@ client.on('interactionCreate', async interaction => {
 	const { commandName } = interaction;
 
 	if (commandName === 'react') {
-		const message = await interaction.reply('You can react with Unicode emojis!', { fetchReply: true });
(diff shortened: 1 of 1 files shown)


WiseDevHelper:
Doesn't #864 fix this?

BaumianerNiklas:
As far as I can tell that PR only fixes similar issues in the Async/Await section of the guide, while this one only fixes them in the Reactions section. There may also still be other sections of the guide with these issues though

Megatank58:
I can have these changes in my pr along with the async page if you would like or both pr can be merged, also only these two pages have the outdated code afaik

BaumianerNiklas:
I dont really mind, if you want to you can add these changes to your PR and I'll close this one. I took a quick look at some other pages in the Popular Topics section when making this PR but didn't notice anything that needed to be changed

Megatank58:
Hm alright then I'll just include these changes in the same pr

Megatank58:
I have included these changes in #864

PR #869, opened 2021-09-15
clarify custom Client.commands property usage

itsravenous:
Primarily to avoid confusion by consumers using typescript. I was motivated to add this after experiencing this issue https://github.com/discordjs/discord.js/issues/6638. I appreciate the guide is meant to be as accessible as possible, and focuses on JS rather than TS. I think this could still be helpful though - anecdotally more and more people are adopting TS. There's at least [a few others](https://stackoverflow.com/questions/62860164/stuck-with-adding-variable-to-discord-client-object-typescript) that have experienced this.

On the other hand, I understand if this feels like extra noise in a guide that's supposed to be focused on getting people up and running, so no worries if it doesn't feel like a good addition.

### DIFF
--- a/guide/creating-your-bot/command-handling.md
+++ b/guide/creating-your-bot/command-handling.md
@@ -175,6 +175,10 @@ for (const file of commandFiles) {
 }
 ```
 
+::: tip
+The `commands` property doesn't exist on the `Client` class by default, and it's not anything special (you could in fact call it whatever you like). We suggest adding it to the client like this so that you have access to it wherever you use your client.
+:::
+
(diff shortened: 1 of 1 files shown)


Danktuary:
This is valid, but I think that instead of it being a tip below, it can be included as a sentence in one of the paragraphs above.

It might fit well below the first codeblock in the "Reading command files" section, e.g.:
"We recommend attaching a `.commands` property to your `client` instance so that you can access your commands in other files"

itsravenous:
Sure thing! Thanks for the review. I’ll switch it up - probably be next week.

PR #883, opened 2021-09-24
fix: set node-fetch install version to 2.6.5

Megatank58:
node-fetch v3 allows only esm code, to use require the version must be 2.6.5 (latest subversion of 2)

### DIFF
--- a/guide/additional-info/rest-api.md
+++ b/guide/additional-info/rest-api.md
@@ -13,17 +13,17 @@ To install node-fetch, run the following command:
 :::: code-group
 ::: code-group-item npm
 ```sh:no-line-numbers
-npm install node-fetch
+npm install node-fetch@2.6.5
 ```
 :::
 ::: code-group-item yarn
 ```sh:no-line-numbers
-yarn add node-fetch
+yarn add node-fetch@2.6.5
 ```
 :::
 ::: code-group-item pnpm
 ```sh:no-line-numbers
-pnpm add node-fetch
+pnpm add node-fetch@2.6.5
 ```
 :::
 ::::


WiseDevHelper:
> Also should there be a tip on why we are installing 2.6.5 and not the latest?

I think yes, there should.

Megatank58:
> > Also should there be a tip on why we are installing 2.6.5 and not the latest?
> 
> I think yes, there should.
> Also could the command be `npm i node-fetch@2` instead?

I'll add the tip, also what difference do they have i suppose both work the same

WiseDevHelper:
> I'll add the tip, also what difference do they have i suppose both work the same

I believe both work the same, so just `2` would be better than `2.6.5` IMO

Danktuary:
`node-fetch@^2.6.5` should be used, and a tip stating that node-fetch v3 doesn't work with `require()` syntax would be good

Megatank58:
> `node-fetch@^2.6.5` should be used, and a tip stating that node-fetch v3 doesn't work with `require()` syntax would be good

Alright I'll keep the version and add a tip

guide/additional-info/rest-api.md
```diff
@@ -13,21 +13,25 @@ To install node-fetch, run the following command:
...
 ::: code-group-item pnpm
 ```sh:no-line-numbers
-pnpm add node-fetch
+pnpm add node-fetch@2.6.5
 ```
 :::
 ::::
 
+::: tip
+We're installing version 2.6.5 of node-fetch because version 3+ doesn't works with `require()` syntax
```
WiseDevHelper, line 32:
```suggestion
We're installing version 2.6.5 of `node-fetch` because versions 3+ do not support `require()` syntax.
```

guide/additional-info/rest-api.md
```diff
@@ -13,21 +13,25 @@ To install node-fetch, run the following command:
...
 ::: code-group-item pnpm
 ```sh:no-line-numbers
-pnpm add node-fetch
+pnpm add node-fetch@2.6.5
 ```
 :::
 ::::
 
+::: tip
+We're installing version 2.6.5 of `node-fetch` because versions 3+ do not support `require()` syntax.
```
almostSouji, line 32: Wrong "we" usage. You're not referring to the guide writers specifically here (we aren't installing anything on the reader's system). I'd propose

```suggestion
Newer versions of `node-fetch` do not support the [CommonJS](https://nodejs.org/api/modules.html#modules_modules_commonjs_modules) `require()` syntax.
```

...or similar.

almostSouji (changes requested):

almostSouji (approved):

SpaceEEC (approved):
There is also a `cjs` tag, which might be of interest:
https://www.npmjs.com/package/node-fetch?activeTab=versions

Megatank58:
Sorry for late reply, should the `cjs` tag be used here or is `2.6.5` fine?

PR #892, opened 2021-10-02
Fix Typo

apidev234:
Collectors Had a Typo, Unverified By The Team [ Mybe ]


monbrey:
I don't see this as a typo. The class name is `Collector` and that the linked portion of the text.

PR #900, opened 2021-10-04
Make sure webhook only returns the bots webhook

KK-Designs:
In the guide, I followed all instructions to send a message through a webhook. Then I got the error `[WEBHOOK_TOKEN_UNAVAILABLE]: This action requires a webhook token, but none is available.` I spent 3-5 days trying to figure out why it was happening, and then I found out why. When you fetch a channels webhooks, there a realistic chance that a webhook from another bot is the first one. So when you call `webhooks.first()` it gets the other bots webhook and gives an error. The PR uses a `.find()` function to only filter out webhooks from the bot itself and resolving the error.


guide/popular-topics/webhooks.md
```diff
@@ -157,6 +157,21 @@ client.once('ready', async () => {
 client.login(token);
 ```
 
+::: warning
+If your bot isn't the first webhook in the channel, it will give `[WEBHOOK_TOKEN_UNAVAILABLE]: This action requires a webhook token, but none is available.`
+:::
+
+So, to fix this, we will only try to find our bot with the following code:
```
GodderE2D, line 164:
```suggestion
If your bot isn't the first webhook in the channel, it will throw `[WEBHOOK_TOKEN_UNAVAILABLE]: This action requires a webhook token, but none is available.`

To fix this, we will only try to find our bot's webhook with the following code:
:::
```

Also, please include the codeblock and "The code above reads..." *inside* the warning.

guide/popular-topics/webhooks.md
```diff
@@ -157,6 +157,21 @@ client.once('ready', async () => {
...
+
+```diff
+- const webhooks = await channel.fetchWebhooks();
+- const webhook = webhooks.first();
++ const webhooks = await channel.fetchWebhooks();
++ const webhook = webhooks.find(wh => wh.owner.id === client.user.id);
+```
+
+The code above reads all of the channel's webhooks and only gives the webhooks by the bot.
```
GodderE2D, line 173:
```suggestion
The code above fetches all of the channel's webhooks and only return a webhook created by the bot.
```

guide/popular-topics/webhooks.md
KK-Designs: Oh ok i will thanks

WiseDevHelper:
Fixed in [release `13.2.0`](https://github.com/discordjs/discord.js/releases/tag/13.2.0).

KK-Designs:
Thats not exactly what I'm talking about, My PR is to change the guide to only fetch webhooks from the Bot

KK-Designs:
Alright, commited change

almostSouji (changes requested):
Any webhook that the bot knows a token for can be executed. (Created by a user or the logged in bot)

A safer, more inclusive, and preferable version of this guard would check for [`Webhook#token`](https://discord.js.org/#/docs/main/stable/class/Webhook?scrollTo=token) instead.

This needs to be reflected in both code sample and guide section explanation.

KK-Designs:
Something like this?
```js
const webhook = webhooks.find(wh => wh.token);
```

almostSouji:
yep, and might also want to handle the case that none is found (a log should suffice for guide purposes).

KK-Designs:
I committed my changes. Can you review them @almostSouji? I also merged the repo to the latest.

KK-Designs:
@almostSouji Can you check the PR and changes?

guide/popular-topics/webhooks.md
```diff
@@ -125,7 +125,7 @@ webhookClient.send({
 });
 ```
 
-Example using a Webhook:
+To make sure the resulting webhook can be executed by your bot, you need to find a webhook that is managable by your bot. To do so, you can check if a token of the webhook exists.
```
almostSouji, line 128: In my opinion this sounds quite clunky and doesn't read all too well.

> Try to find a webhook your bot knows the token for. This makes sure your bot can execute the webhook lateron.

I find much easier to read, as an example. Feel free to rephrase or use a different way to phrase it altogether, of course.

almostSouji (changes requested):

guide/popular-topics/webhooks.md
KK-Designs: So for example:
```diff
- To make sure the resulting webhook can be executed by your bot, you need to find a webhook that is manageable by your bot. To do so, you can check if a token of the webhook exists.
+ Try to find a webhook your bot knows the token for. This makes sure your bot can execute the webhook later on.
```
Like that? I don't want to make an unnecessary commit.

almostSouji: sure, if you want to rephrase it, feel free

KK-Designs: Ok, I will do so.

almostSouji (approved):

PR #951, opened 2021-11-20
fix(misc): make parsing mention arguments more resilient

AlexandrosGounis:
fixes USERS_PATTERN regex return

This expression
```ts
const matches = mention.match(USERS_PATTERN);
```
may return the match without the actual id. This might cause confusion, as `matches[1]` does not exist.
We need to `exec` on that regex to include the capture groups: 
```ts
USERS_PATTERN.exec(mention)
```
which always returns the id found, if any

### DIFF
--- a/guide/miscellaneous/parsing-mention-arguments.md
+++ b/guide/miscellaneous/parsing-mention-arguments.md
@@ -89,7 +89,7 @@ const { MessageMentions: { USERS_PATTERN } } = require('discord.js');
 
 function getUserFromMention(mention) {
 	// The id is the first and only match found by the RegEx.
-	const matches = mention.match(USERS_PATTERN);
+	const matches = USERS_PATTERN.exec(mention);
 
 	// If supplied variable was not a mention, matches will be null instead of an array.
 	if (!matches) return;
--- a/guide/miscellaneous/parsing-mention-arguments.md
+++ b/guide/miscellaneous/parsing-mention-arguments.md
@@ -89,7 +89,7 @@ const { MessageMentions: { USERS_PATTERN } } = require('discord.js');
 
 function getUserFromMention(mention) {
 	// The id is the first and only match found by the RegEx.
-	const matches = mention.matchAll(USERS_PATTERN).next().value
+	const matches = mention.matchAll(USERS_PATTERN).next().value;
 
 	// If supplied variable was not a mention, matches will be null instead of an array.
 	if (!matches) return;
--- a/guide/miscellaneous/parsing-mention-arguments.md
+++ b/guide/miscellaneous/parsing-mention-arguments.md
@@ -89,7 +89,7 @@ const { MessageMentions: { USERS_PATTERN } } = require('discord.js');
 
 function getUserFromMention(mention) {
 	// The id is the first and only match found by the RegEx.
-	const matches = USERS_PATTERN.exec(mention);
+	const matches = mention.matchAll(USERS_PATTERN).next().value
 
 	// If supplied variable was not a mention, matches will be null instead of an array.
 	if (!matches) return;


almostSouji (changes requested):
Please provide an example of a failing input text in the PR description. I currently don't understand why this is needed.

almostSouji:
Ah, so essentially the global flag makes match ignore capture groups but just return full matches, accordingly not returning an id. I see.

>  If the g flag is used, all results matching the complete regular expression will be returned, but capturing groups will not.
>  if the g flag is not used, only the first complete match and its related capturing groups are returned. In this case, the returned item will have additional properties as described below.
<https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/match>

guide/miscellaneous/parsing-mention-arguments.md
```diff
@@ -89,7 +89,7 @@ const { MessageMentions: { USERS_PATTERN } } = require('discord.js');
 
 function getUserFromMention(mention) {
 	// The id is the first and only match found by the RegEx.
-	const matches = mention.match(USERS_PATTERN);
+	const matches = USERS_PATTERN.exec(mention);
```
almostSouji, line 92: This is currently prone to the [statefulness of RegExp#exec](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/RegExp/exec), meaning every second execution of the function will not yield any match. The patterns provided by [discord.js](https://github.com/discordjs/discord.js/blob/stable/src/structures/MessageMentions.js#L221) are saved on the object as instance, not provided as fresh instances through getters.

> JavaScript RegExp objects are stateful when they have the global or sticky flags set (e.g. /foo/g or /foo/y). They store a lastIndex from the previous match. Using this internally, exec() can be used to iterate over multiple matches in a string of text (with capture groups), as opposed to getting just the matching strings with String.prototype.match(). 

The following shows the issue:

```js
const USERS_PATTERN = /<@!?(\d{1,19})>/g;

function getUserFromMention(mention) {
  const matches = USERS_PATTERN.exec(mention);
  if (!matches) return;
  return matches[1];
}

console.log("---");
console.log("id: ", getUserFromMention("<@1>")); // 1
console.log("id: ", getUserFromMention("<@2>")); // undefined
console.log("id: ", getUserFromMention("<@3>")); // 3
console.log("id: ", getUserFromMention("<@4>")); // undefined
console.log("id: ", getUserFromMention("<@5>")); // 5
```

almostSouji (changes requested):

guide/miscellaneous/parsing-mention-arguments.md
AlexandrosGounis: I would suggest having a single RegExp instance if this is going to be used more than once... then simply reset that last index:

```js
import { MessageMentions } from "discord.js";

const USERS_REGEX = new RegExp(MessageMentions.USERS_PATTERN);

function getUserFromMention(mention) {
  USERS_REGEX.lastIndex = 0;
  const matches = USERS_PATTERN.exec(mention);
  if (!matches) return;
  return matches[1];
}

or create a new RegExp within that function, then no need to reset, (bad for GC)
```

almostSouji: That's rather hacky, I'd prefer using <https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/matchAll> over this

AlexandrosGounis:
@almostSouji, I updated my suggestion.
That's right, but to me they all feel hacky and instantiate an iterator under the hood, I am not entirely convinced this is better than even having a RegExp (which won't fail on empty input either)

For what it's worth, it's better than what we have right now:
```ts
mention.matchAll(USERS_PATTERN).next().value // hacky
[...mention.matchAll(USERS_PATTERN)][0] // also hacky
```

guide/miscellaneous/parsing-mention-arguments.md
```diff
@@ -89,7 +89,7 @@ const { MessageMentions: { USERS_PATTERN } } = require('discord.js');
 
 function getUserFromMention(mention) {
 	// The id is the first and only match found by the RegEx.
-	const matches = mention.match(USERS_PATTERN);
+	const matches = mention.matchAll(USERS_PATTERN).next().value
```
almostSouji, line 92: missing semi colon (CI red)

almostSouji (changes requested):
ci red, otherwise ready to merge

guide/miscellaneous/parsing-mention-arguments.md
WiseDevHelper:
```suggestion
	const matches = mention.matchAll(USERS_PATTERN).next().value;
```

almostSouji (approved):

Issue #983, opened 2021-12-31
Enclose the three pages of slash commands in one


vitorlops:
**Is your feature request related to a problem? Please describe.**
When people see the Interactions item in the sidebar, perhaps will be better if there are Slash commands, Buttons and Select menu pages, not three for only one topic.

**Describe the solution you'd like**
I think if the three slash commands pages (Registering slash commands, Replying to slash commands and Slash command permissions) came together, forming a single page, it would be clearer.

**Describe alternatives you've considered**
Here is a picture of what i think looks better:
![image](https://user-images.githubusercontent.com/79790819/147836814-e47eeaf9-ed39-4419-8ecf-7439cae6b971.png)

**Additional notes**
If it really be applied, there will need to change the files that links to that pages.

PR #985, opened 2022-01-01
fix(interactions): Enclosing the slash commands pages in one unique

vitorlops:
Closes the issue #983. That makes the Interactions sidebar item clear (I guess).


almostSouji (changes requested):
Has conflicts that need to be resolved, apart from that: looks good to me

PR #986, opened 2022-01-02
Update Deprecated Method in Embeds Discord.js Guide (setFooter)

Aerosphia:
Updated the old, deprecated `setFooter` method with the new method.


Aerosphia:
nevermind, was already reported

PR #993, opened 2022-01-09
Add axios and crypto-js to useful packages

silktg:
I added two very useful packages to the guide: axios and crypto-js.

Axios facilitates http(s) requests of all kinds. (similar to node-fetch, but better [at least imho]).

Crypto-JS is a simple package that helps encrypt and secure user data for at rest remote storage (which is required by the Discord Terms of Service).

### DIFF
--- a/guide/miscellaneous/useful-packages.md
+++ b/guide/miscellaneous/useful-packages.md
@@ -359,7 +359,7 @@ function decryptData(data){
 
 You can now send the `secureData` to the database, and it will be encrypted at rest. When you need to retrieve the data, you should retrieve the data from your database, and then decrypt it using a function similar to the one demonstrated above.
 
-**Please Note:** In this demonstration I only encrypted Strings, however, it is entirely possible to encrypt any data type. 
+**Please Note:** In this demonstration is only encrypted Strings, however, it is entirely possible to encrypt any data type. 
 
 :::warning
 Please ensure that the passphrase you use to encrypt user data is **strong**. It should consist of at least eight characters, and should include letters, numbers and symbols.
--- a/guide/miscellaneous/useful-packages.md
+++ b/guide/miscellaneous/useful-packages.md
@@ -263,3 +263,107 @@ Official documentation: [https://www.i18next.com](https://www.i18next.com)
 i18next is an internationalization-framework for JavaScript. It is beneficial to translate your bot's user-facing messages into various languages based on the server it is used in.
 
 Covering an entire use case example for internationalization would be out of this guide's scope and requires some more explanation as to how the system operates. Please refer to the official documentation linked above for an in-depth usage guide.
+
+## axios
+
+:::tip
+Axios is used to make HTTP(S) requests to remote servers. The documentation can be located [Here](https://axios-http.com/docs/intro)
+:::
+Example:
+<!-- eslint-skip -->
+```js
+const axios = require('axios')
+
+axios({
+	method: 'post',
+	url: 'https://example.com/v1',
+	headers:
+	{
+		'Content-Type:':'application/json',
+		'User-Agent':'MyBot'
+	},
+	data{
+		"item":"one"
+	}
+}).then(response =>{
+	console.log(response)
+})
+
+```
+
+:::warning
+If the Axios request fails for one or more reasons, it will fire an an uncaught error. For this reason, it is recommended that you place Axios requests inside of a "try","catch","finally" sequence to prevent your application from crashing.
+
+An example of this is below
+:::
+<!-- eslint-skip -->
+```js
+
+const axios = require('axios')
+
+try{
+axios({
+	method: 'post',
+	url: 'https://example.com/v1',
+	headers:
+	{
+		'Content-Type:':'application/json',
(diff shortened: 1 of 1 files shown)


guide/miscellaneous/useful-packages.md
```diff
@@ -263,3 +263,107 @@ Official documentation: [https://www.i18next.com](https://www.i18next.com)
 i18next is an internationalization-framework for JavaScript. It is beneficial to translate your bot's user-facing messages into various languages based on the server it is used in.
 
 Covering an entire use case example for internationalization would be out of this guide's scope and requires some more explanation as to how the system operates. Please refer to the official documentation linked above for an in-depth usage guide.
+
+## axios
+
+:::tip
+Axios is used to make HTTP(S) requests to remote servers. The documentation can be located [Here](https://axios-http.com/docs/intro)
```
vitorlops, line 270:
```suggestion
Axios is used to make HTTP(S) requests to remote servers. The documentation can be found [here](https://axios-http.com/docs/intro).
```

guide/miscellaneous/useful-packages.md
```diff
@@ -263,3 +263,107 @@ Official documentation: [https://www.i18next.com](https://www.i18next.com)
...
+		"item":"one"
+	}
+}).then(response =>{
+	console.log(response)
+})
+
+```
+
+:::warning
+If the Axios request fails for one or more reasons, it will fire an an uncaught error. For this reason, it is recommended that you place Axios requests inside of a "try","catch","finally" sequence to prevent your application from crashing.
```
vitorlops, line 295:
```suggestion
If the Axios request fails for one or more reasons, it will fire an an uncaught error. For this reason, it is recommended that you place Axios requests inside of a `try`, `catch` and `finally` sequence to prevent your application from crashing.
```

guide/miscellaneous/useful-packages.md
```diff
@@ -263,3 +263,107 @@ Official documentation: [https://www.i18next.com](https://www.i18next.com)
...
+
+
+```
+
+## crypto-js
+
+:::tip
+Crypto-JS is used to encrypt and decrypt data. This is helpful if you are storing user data in a remote database, and it is not encrypted at rest by default (ie. mongodb, firebase, aws, etc.) - It is always good practice to store user data encrypted at rest, and it is **required** by Discord's Terms of Service. 
+
+The documentation can be found [Here](https://cryptojs.gitbook.io/docs/)
```
vitorlops, line 334:
```suggestion
The documentation can be found [here](https://cryptojs.gitbook.io/docs/).
```

guide/miscellaneous/useful-packages.md
```diff
@@ -263,3 +263,107 @@ Official documentation: [https://www.i18next.com](https://www.i18next.com)
...
+	let decrypted = bytes.toString(CryptoJS.enc.Utf8)
+	return decrypted
+}
+
+
+```
+
+You can now send the `secureData` to the database, and it will be encrypted at rest. When you need to retrieve the data, you should retrieve the data from your database, and then decrypt it using a function similar to the one demonstrated above.
+
+**Please Note:** In this demonstration I only encrypted Strings, however, it is entirely possible to encrypt any data type.
```
vitorlops, line 362:
```suggestion
**Please Note:** In this demonstration is only encrypted Strings, however, it is entirely possible to encrypt any data type. 
```

guide/miscellaneous/useful-packages.md
```diff
@@ -263,3 +263,107 @@ Official documentation: [https://www.i18next.com](https://www.i18next.com)
...
+
+
+```
+
+You can now send the `secureData` to the database, and it will be encrypted at rest. When you need to retrieve the data, you should retrieve the data from your database, and then decrypt it using a function similar to the one demonstrated above.
+
+**Please Note:** In this demonstration I only encrypted Strings, however, it is entirely possible to encrypt any data type. 
+
+:::warning
+Please ensure that the passphrase you use to encrypt user data is **strong**. It should consist of at least eight characters, and should include letters, numbers, and symbols.
```
vitorlops, line 365:
```suggestion
Please ensure that the passphrase you use to encrypt user data is **strong**. It should consist of at least eight characters, and should include letters, numbers and symbols.
```

vitorlops (changes requested):
Fixing punctuation, using lowercase words in the middle of some sentences and taking out the `I`.

guide/miscellaneous/useful-packages.md
silktg: > 

This change contains incorrect phrasing. I recommend:

"In the demonstration above, only Strings were encrypted." If a change is required.

vitorlops: yeah, that is better, i just changed because of the `I`.

almostSouji:
<https://github.com/nodejs/undici>, being written for node.js from scratch, by the node dev team seems to be the go-to request library we'd recommend.

While storing user data at rest is indeed required, this should probably be done with the database, where the data rests, instead.
Unsure if I see much usefulness for it in a beginners bot guide.

PR #996, opened 2022-01-13
match is the wrong method here

justmedev:
use exec to actually get the capturing group.

Changed `.match(regex)` on the string to `.exec(mention)` on the regex since `.match` doesn't return the capturing group.

### DIFF
--- a/guide/miscellaneous/parsing-mention-arguments.md
+++ b/guide/miscellaneous/parsing-mention-arguments.md
@@ -76,7 +76,7 @@ Here is how the RegEx works:
  4. `\d+` means the RegEx will look for multiple digits, which will be the ID.
  5. The parentheses around the `\d+` create a capture group, which allows you to get the ID out of the mention.
 
-Using the `.match()` method on strings, you can get the capture group's values, i.e., the mention's ID.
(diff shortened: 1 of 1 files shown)


guide/miscellaneous/parsing-mention-arguments.md
```diff
@@ -76,7 +76,7 @@ Here is how the RegEx works:
  4. `\d+` means the RegEx will look for multiple digits, which will be the ID.
  5. The parentheses around the `\d+` create a capture group, which allows you to get the ID out of the mention.
 
-Using the `.match()` method on strings, you can get the capture group's values, i.e., the mention's ID.
+Using the `.exec()` method on the regex passing the the `mention` string, you can get the capture group's values, i.e., the mention's ID.
```
vitorlops, line 79:
```suggestion
Using the `.exec()` method on the RegEx passing the the `mention` string, you can get the capture group's values, i.e., the mention's ID.
```
Maybe keep the RegEx with the uppercase letters.

almostSouji:
See here https://github.com/discordjs/guide/pull/951#discussion_r769119618 as to why just swapping the method will not work here. Additionally, this is covered by #951 already

PR #997, opened 2022-01-14
Updated embeds.md

Shadyyey:
Changed embeds.md because of footer url beign deprecated to work with the current discord.js version.


Jiralite:
The footer `iconURL` is not deprecated. How the method currently is currently used is fine.

Shadyyey:
Yes, it is decapretated. Now if you check the website you will see it, https://discord.js.org/#/docs/main/stable/class/MessageEmbed?scrollTo=setFooter

Shadyyey:
You are probably on older versions of discord.js, so you don't see the error popping up. But if you were on 13.6.0 or 13.5.1 you would get the error.

Shadyyey:
You would need to use the options parameter instead of url.

PR #1000, opened 2022-01-18
chore: update links

almeidx:
The old links point to archived repositories


guide/whats-new.md
```diff
@@ -55,7 +55,7 @@ All content has been updated to use discord.js v13 syntax. The v12 version of th
 ### Updated
 
 - Commando: Removed, to be replaced with a guide on [Sapphire](https://github.com/discordjs/guide/pull/711)
```
WiseDevHelper, line 57:
```suggestion
- Commando: Removed, to be replaced with [Sapphire](https://www.sapphirejs.dev/) 
```

vitorlops: I don't think it's necessary, because the it's related with this PR specifically.

PR #1020, opened 2022-02-02
author & footer icon_url => iconURL

Larsundso:
Please describe the changes this PR makes: 
> Changed Author and Footer icon_url to iconURL

and why it should be merged:
> For an up-to-date Guide

### DIFF
--- a/guide/popular-topics/embeds.md
+++ b/guide/popular-topics/embeds.md
@@ -114,7 +114,7 @@ const exampleEmbed = {
 	url: 'https://discord.js.org',
 	author: {
 		name: 'Some name',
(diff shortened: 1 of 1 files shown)


megatank58:
`icon_url` does works in raw embeds (objects) so this change isn't needed

Larsundso:
That is true however when accessing `icon_url` on Discord.js initiated Embeds nothing is returned
so it would be better to have something in a beginner-friendly guide which supports both

almostSouji:
This section very deliberately uses the raw embed structure properties as documented by discord upstream (https://discord.com/developers/docs/resources/channel#embed-object-embed-footer-structure).

PR #1025, opened 2022-02-05
Allowing to use import()

DaInfLoop:

Adds additional notes to show how to use the `import()` function to import node-fetch without having to install the cjs tag.


almostSouji:
As you already state in your PR await import is not recommended for this use case

PR #1036, opened 2022-02-17
Replace node-fetch with undici

DaInfLoop:
Make sure the node-fetch install command is the cjs tag version, otherwise the code on line 195-233 will throw errors!

### DIFF
--- a/code-samples/oauth/simple-oauth-webserver/index.js
+++ b/code-samples/oauth/simple-oauth-webserver/index.js
@@ -1,4 +1,4 @@
-const fetch = require('node-fetch');
+const { request } = require('undici');
 const express = require('express');
 const { clientId, clientSecret, port } = require('./config.json');
 
--- a/code-samples/oauth/simple-oauth-webserver/index.js
+++ b/code-samples/oauth/simple-oauth-webserver/index.js
@@ -2,6 +2,17 @@ const fetch = require('node-fetch');
 const express = require('express');
 const { clientId, clientSecret, port } = require('./config.json');
 
+// Just make a helper function to get JSON response from request()
+async function getJSONResponse(body) {
+	let full_body = '';
+
+	for await (const data of body) {
+		full_body = `${full_body}${data.toString()}`;
+	}
+
+	return JSON.parse(full_body);
+}
+
 const app = express();
 
 app.get('/', async ({ query }, response) => {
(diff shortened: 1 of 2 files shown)


xHyroM:
I'd probably swap node-fetch to undici straight away, because djs will be switching to undici as well.

DaInfLoop:
> I'd probably change node-fetch to undici straight away, because djs will be switching to undici as well.

I'll have a look at it now, then I'll edit the page based on it.

N0ahMC:
Is there a reason you used undici instead of axios? Just wondering, because I believe it's usually preferred.

suneettipirneni:
> Is there a reason you used undici instead of axios? Just wondering, because I believe it's usually preferred.

Axios is one of the slowest performing http request libraries which doesn't make it a great candidate for an api wrapper that makes a lot of requests at once. undici seems to be one of the fastest due to its many optimizations. In addition, it provides many spec compliant structures like FormData that are cross compatible with the web standards.

N0ahMC:
> > Is there a reason you used undici instead of axios? Just wondering, because I believe it's usually preferred.
> 
> Axios is one of the slowest performing http request libraries which doesn't make it a great candidate for an api wrapper that makes a lot of requests at once. undici seems to be one of the fastest due to its many optimizations. In addition, it provides many spec compliant structures like FormData that are cross compatible with the web standards. (And it'll be built into node soon)

Ahhh that makes sense, thanks!

almostSouji (commented):
Why not stick to the fetch API with #fetch and #json? Is there a reason I'm missing?

DaInfLoop:
Why not stay with it?

The fetch API part of undici is still experimental, so until it's stable,
may aswell use `request()`

almostSouji:
I see! So then this mostly needs to get off of CI red

DaInfLoop:
> I see! So then this mostly needs to get off of CI red

Tried to fix it, need approval for the ESLint action (which is why I asked for review)

DaInfLoop:
Just went to check deploy preview and saw most of the code highlighting had messed up, if anyone could fix it because I'm busy right now I'd be glad.

GodderE2D:
There's also a node-fetch occurrence here: https://github.com/discordjs/guide/blob/main/code-samples/additional-info/rest-api/13/index.js, would that also need to be changed to undici?

DaInfLoop:
Probably, when I get the chance I'll update that page too.

imranbarbhuiya:
Change the title and description as this pr no longer 3dits node-fetch install command.

almostSouji:
applied in https://github.com/discordjs/guide/commit/f8479528204893b19202903055799021a61ff03a

PR #1040, opened 2022-02-20
feat: replaces PR link with sapphire guide link

megatank58:


vitorlops:
Guess it is not necessary, see https://github.com/discordjs/guide/pull/1000#discussion_r786911151.
Because it's related with this PR in particular.

megatank58:
I don't think it makes much sense to have a link to a PR which is no longer being worked on

PR #1009, opened 2022-01-25
Fixed readDirSync path

lukas0008:
It's a simple fix where in the guide, while creating the command handler, `fs.readdirSync('./commands')` throws `Error: ENOENT: no such file or directory, scandir './commands'`

Added `path.join(__dirname, "./commands")` to command handler in the guide.

PR #1041, opened 2022-02-20
fix: error commands folder not found because wrong path

RobyCigar:
I build my app under `MyRootApp/src` folder, so if I only use './commands' it gives me an error, so I fix this by adding __dirname variable so that I can find the folder path dynamically


vitorlops:
I think https://github.com/discordjs/guide/pull/1009 already does that.

almostSouji:
Already covered by #1009

PR #1043, opened 2022-02-23
feat(faq): use optional chaining to access status

BenjammingKirby:

Since uncached presences would make `member.presence` return `null` i think this bit should be modified to use optional chaining and deal with that to avoid accessing null


guide/popular-topics/faq.md
```diff
@@ -246,7 +246,7 @@ A User represents a global Discord user, and a GuildMember represents a Discord
 ```js
 // First use guild.members.fetch to make sure all members are cached
 guild.members.fetch().then(fetchedMembers => {
```
almeidx, line 248:
```suggestion
guild.members.fetch({ withPresences: true }).then(fetchedMembers => {
```

PR #1045, opened 2022-02-27
feat: set my status every couple of minutes?

twisttaan:

A lot of new javascripters ask the same question,

> "How do I change my bot's status every couple of seconds"

This FAQ entry solves their question.

### DIFF
--- a/guide/popular-topics/faq.md
+++ b/guide/popular-topics/faq.md
@@ -81,6 +81,20 @@ client.user.setAvatar('URL or path');
 client.user.setActivity('activity');
 ```
 
+### How do I set my status every couple of minutes?
+
+```js
+client.once("ready", () => {
+    setInterval(() => {
+      const activities_list = [`activity1`, `activity2`, `activity3`];
+      const index = Math.floor(
+        Math.random() * (activities_list.length - 1) + 1
+      );
+      client.user!.setActivity(activities_list[index], { type: "LISTENING" });
+    }, 50000); // Please don't go faster then 50000ms
+});
+```
+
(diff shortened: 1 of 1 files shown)


guide/popular-topics/faq.md
```diff
@@ -81,6 +81,20 @@ client.user.setAvatar('URL or path');
...
+
+```js
+client.once("ready", () => {
+    setInterval(() => {
+      const activities_list = [`activity1`, `activity2`, `activity3`];
+      const index = Math.floor(
+        Math.random() * (activities_list.length - 1) + 1
+      );
+      client.user!.setActivity(activities_list[index], { type: "LISTENING" });
+    }, 50000); // Please don't go faster then 50000ms
```
itsUrcute, line 94: than*

monbrey:
This is not something we'd like to encourage users of the library to be doing.

guide/popular-topics/faq.md
twisttaan: thx

suneettipirneni:
This feels like a topic that's out of scope of discord.js. The guide telling the user how to set a presence is fine. However, explaining how to periodically change the presence seems to be a topic that's more reliant on JavaScript knowledge and built-ins like `setInterval`. If you know how to repeat a block of code for a given interval in JavaScript alone, changing your presence periodically is rather straightforward.

twisttaan:
Sorry guys I didn't know that this is a grey area, it's just a question I see a lot and a lot of other bots do it. I understand that we shouldn't be promoting shady stuff like this. I have learnt my mistake and will not do it again. 😔

PR #1052, opened 2022-03-02
Clarification: "Client ID" into "Bot's Client ID"

jakjus:
In fact, it is not "your client id" when you refer to the reader, but rather "your bot's client id". 

For someone just starting off with discord bot programming, it may not be obvious that slash commands are, at the smallest scope, registered guild-wise, but never user-wise. Without this knowledge, it may feel justifiable to use particular *user's client ID*, instead of *bot's client ID*, that is actually needed. 

This addition clears the confusion.

### DIFF
--- a/guide/creating-your-bot/creating-commands.md
+++ b/guide/creating-your-bot/creating-commands.md
@@ -45,12 +45,12 @@ pnpm add @discordjs/builders @discordjs/rest discord-api-types
 
 Below is a deployment script you can use. Focus on these variables:
 
-- `clientId`: Your client's id
+- `clientId`: Your bot's client id
 - `guildId`: Your development server's id
 - `commands`: An array of commands to register. The [slash command builder](/popular-topics/builders.md#slash-command-builders) from `@discordjs/builders` is used to build the data for your commands
 
 ::: tip
-In order to get your client and guild ids, open Discord and go to your settings. On the "Advanced" page, turn on "Developer Mode". This will enable a "Copy ID" button in the context menu when you right-click on a server icon, a user's profile, etc.
+In order to get your bot's client id and guild id, open Discord and go to your settings. On the "Advanced" page, turn on "Developer Mode". This will enable a "Copy ID" button in the context menu when you right-click on a server icon, a user's profile, etc.
 :::
 
 :::: code-group


guide/creating-your-bot/creating-commands.md
```diff
@@ -45,12 +45,12 @@ pnpm add @discordjs/builders @discordjs/rest discord-api-types
 
 Below is a deployment script you can use. Focus on these variables:
 
-- `clientId`: Your client's id
+- `clientId`: Your bot's client id
```
megatank58, line 48:
```suggestion
- `clientId`: Your bot's id
```
wouldn't bot's client be simply the bot itself :thinking:

guide/creating-your-bot/creating-commands.md
```diff
@@ -45,12 +45,12 @@ pnpm add @discordjs/builders @discordjs/rest discord-api-types
...
 - `guildId`: Your development server's id
 - `commands`: An array of commands to register. The [slash command builder](/popular-topics/builders.md#slash-command-builders) from `@discordjs/builders` is used to build the data for your commands
 
 ::: tip
-In order to get your client and guild ids, open Discord and go to your settings. On the "Advanced" page, turn on "Developer Mode". This will enable a "Copy ID" button in the context menu when you right-click on a server icon, a user's profile, etc.
+In order to get your bot's client id and guild id, open Discord and go to your settings. On the "Advanced" page, turn on "Developer Mode". This will enable a "Copy ID" button in the context menu when you right-click on a server icon, a user's profile, etc.
```
megatank58, line 53:
```suggestion
In order to get your bot's id and guild ids, open Discord and go to your settings. On the "Advanced" page, turn on "Developer Mode". This will enable a "Copy ID" button in the context menu when you right-click on a server icon, a user's profile, etc.
```
why change `guild ids` to `guild id` that doesn't seems relevant

guide/creating-your-bot/creating-commands.md
tipakA: There are some edge cases where the application has a different id than the bot account that's attached to it, though as far as i know it's more on the "ancient" side.

guide/creating-your-bot/creating-commands.md
tipakA: Because it *was* grammatically correct to say `ids` due to it originally including both *bot and guild* ids. Now it says "bot's client id and guild id", and would have to be something around "bot's client and guild ids", but that would in turn imply that bot has a guild.

Also, the variable this tip is under is a singular id.

almostSouji (changes requested):
Good catch, I see why this can be confusing. The developer applications page (where you get tokens from) has settled on the term "application ID" (some applications do not even have a bot), which I think we should use here.

"Application ID" also makes it very clear that we're not talking about the developer's user id

guide/creating-your-bot/creating-commands.md
almostSouji: bot ID should always be application ID, IIRC those were replaced at some point

jakjus:
I thought about application's ID as well, but the sidebar was mentioning "Bot" more often, e.g "Creating your first Bot". It felt to me like referring to "bot" instead of app during first chapters is reasonable. Nevertheless, going for application seems like the right direction.

Feels like we have to keep **client** word, because it is referred throughout the guide. It's not ideal.
In the OAuth2 section of the application, we can read [an application's] "Client ID", while in General Information it's just "Application ID". 

In [docs](https://discord.com/developers/docs/topics/oauth2#bot-authorization-flow-bot-auth-parameters) we read `bot's client_id` as `your app's **client** id` yet again. 

After all, I think it's fine now.

guide/creating-your-bot/creating-commands.md
tipakA: @almostSouji no, they were not, or at least not fully if that ever happened. The only bot i know that has such "issue" still has a different ID in the invite link than the bot account it uses. It doesn't even "redirect", and just gives you "unknown application".

Admittedly though, with the addition of the "Add to Server" button (which bot in question does use), i don't think this will be anyhow common problem (on top of the fact that you had to know how to "make" a bot invite link in order to actually bump into it in the first place).

PR #1059, opened 2022-03-09
feat: replace 'canvas' with '@napi-rs/canvas'

KhafraDev:

[@napi-rs/canvas](https://www.npmjs.com/package/@napi-rs/canvas) is similar to canvas but does not rely on postinstallation scripts (no node-gyp!!) which makes it much more useful for new developers or basically anyone that has ever tried debugging why a node-gyp error occurs. It supplies the same canvas api, along with some non-spec compliant apis such as `canvas.toBuffer(?mimeType)` that `canvas` has.

I'm unsure if this change is wanted but not having to deal with node-gyp anymore seems like a huge win IMO. 👍 

## Pros
- No post installation script, no build tools needed, etc...
- Less dependencies (faster start up for beginners, less time installing packages)
- Faster (doubt this really matters as this guide is aimed at beginners)

## Cons
- No `Canvas.loadImage` method, this is replaced by either using `readFile` from `fs/promises` for local files and with `undici` for third-party files. So I guess this PR depends on #1036?

### DIFF
--- a/code-samples/popular-topics/canvas/13/index.js
+++ b/code-samples/popular-topics/canvas/13/index.js
@@ -1,5 +1,7 @@
 const { Client, Intents, MessageAttachment } = require('discord.js');
-const Canvas = require('canvas');
+const { createCanvas, Image } = require('@napi-rs/canvas');
+const { readFile } = require('fs/promises');
+const { fetch } = require('undici');
 
 const client = new Client({ intents: [Intents.FLAGS.GUILDS] });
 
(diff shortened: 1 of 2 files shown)


code-samples/popular-topics/canvas/13/index.js
```diff
@@ -44,10 +48,12 @@ client.on('interactionCreate', async interaction => {
 		context.closePath();
 		context.clip();
 
-		const avatar = await Canvas.loadImage(interaction.user.displayAvatarURL({ format: 'jpg' }));
+		const avatarRes = await fetch(interaction.user.displayAvatarURL({ format: 'jpg' }));
```
DaInfLoop, line 51: As I had said in #1036, `undici#fetch` is experimental, so I wouldn't recommend it. But by all means go ahead.

KhafraDev: Doesn't really mean much. The experimental label was added way before basic spec features were added such as fetching data or blob urls. Nowadays the implementation is mostly complete, minus some browser-only things (origin, CORs, etc.). Plus, the example follows the spec and with the vast amount of tests undici's fetch has, this example will never break

KhafraDev: (resolved)

almostSouji (approved):

code-samples/popular-topics/canvas/13/index.js
```diff
@@ -44,10 +48,12 @@ client.on('interactionCreate', async interaction => {
 		context.closePath();
 		context.clip();
 
-		const avatar = await Canvas.loadImage(interaction.user.displayAvatarURL({ format: 'jpg' }));
+		const { body } = await request(interaction.user.displayAvatarURL({ format: 'jpg' }));
+		const avatar = new Image()
```
almostSouji, line 52:
```suggestion
		const avatar = new Image();
```

PR #1069, opened 2022-03-25
feat:  path construction

CarpentersKeys:
I found that these requires weren't working for me despite same relative path. This might cause fewer errors for people adapting the code.

Node.js v16.14.0.
discord.js  v13.6.0

**Please describe the changes this PR makes and why it should be merged:**

PR #1102, opened 2022-05-04
Update fs.readdirSync() setup.

Hades6578:
fs.readdirSync() no longer functions without invoking the path to command folder via fs.realpathSync(), when you do not, it returns "Error: ENOENT: no such file or directory, scandir './commands'. " Simply requires that extra line to work properly.

Took quite a while to attempt to get the code to find the file, originally deterred me from continuing to work on my bot.  I solved the problem through sheer luck.

### DIFF
--- a/guide/creating-your-bot/command-handling.md
+++ b/guide/creating-your-bot/command-handling.md
@@ -154,8 +154,7 @@ We recommend attaching a `.commands` property to your client instance so that yo
 [`fs`](https://nodejs.org/api/fs.html) is Node's native file system module. <DocsLink section="collection" path="class/Collection" /> is a class that extends JavaScript's native [`Map`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map) class, and includes more extensive, useful functionality.
 :::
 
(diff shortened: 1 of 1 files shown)


Hades6578:
Please let me know if this change is necessary or helpful.  If it's not relevant, I will remove this pull req.

megatank58:
#1069 already does this and a few more changes

Hades6578:
> #1069 already does this and a few more changes

I went back through the tutorial, and it does not appear as if the changes were actually pushed yet.  I also attempted to fix my error with the solution, but it produces a similar error.

megatank58:
The guide should then just have it directly from the starting instead of having an update or note somewhere and all occurrences would need to be changed

Hades6578:
> The guide should then just have it directly from the starting instead of having an update or note somewhere and all occurrences would need to be changed

Ok, how would I go about that? Or is that someone else's job?

Hades6578:
Rewrote the lines, and more smoothly integrated the modified code.  Should be ready for merge, but please check to make sure you like the way I wrote it.

almostSouji:
I am still not sure what this fixes? I cannot reproduce the issue you seem to be experiencing.

Hades6578:
> I am still not sure what this fixes? I cannot reproduce the issue you seem to be experiencing.

Strange, I had the standard EONET error. I wonder why it isn’t coming up for you.

almostSouji:
Until someone can reproduce and explain why exactly this is necessary, we won't implement this.

Hades6578:
> Until someone can reproduce and explain why exactly this is necessary, we won't implement this.

Will attempt to recreate circumstances and attach screenshots of output.

PR #1116, opened 2022-05-14
feat(slashcommands): add slash command permissions section

megatank58:

Adds the section to set the new default slash command permissions on the time of registering the commands.

### DIFF
--- a/guide/interactions/slash-commands.md
+++ b/guide/interactions/slash-commands.md
@@ -511,7 +511,7 @@ const data = new SlashCommandBuilder()
 	.setDescription('Ban a member!')
 	.addUserOption(option =>
 		option.setName('target').setDescription('The member to ban'))
-	.setDefaultMemberPermissions([Permissions.FLAGS.KICK_MEMBERS, Permissions.FLAGS.BAN_MEMBERS]);
+	.setDefaultMemberPermissions(new Permissions([Permissions.FLAGS.KICK_MEMBERS, Permissions.FLAGS.BAN_MEMBERS]).bitfield);
 ```
 
 And that's all you need to know on slash command permissions!
--- a/guide/interactions/slash-commands.md
+++ b/guide/interactions/slash-commands.md
@@ -507,7 +507,7 @@ const { SlashCommandBuilder } = require('@discordjs/builders');
 const { Permissions } = require('discord.js');
 
 const data = new SlashCommandBuilder()
-	.setName('Ban')
+	.setName('ban')
 	.setDescription('Ban a member!')
 	.addUserOption(option =>
 		option.setName('target').setDescription('The member to ban'))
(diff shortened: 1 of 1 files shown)


guide/interactions/slash-commands.md
```diff
@@ -476,3 +476,42 @@ await interaction.reply('Pong!');
...
+### Member permissions
+
+You can use the `setDefaultMemberPermissions` method to set the default permissions required by a member in order to run the command, setting it to `0` will prohibit anyone in a guild from using the command unless a specific overwrite is configured or the user has admin permissions.
+
+```js {9}
+const { SlashCommandBuilder } = require('@discordjs/builders');
+const { Permissions } = require('discord.js');
+
+const data = new SlashCommandBuilder()
+	.setName('Ban')
```
BenjammingKirby, line 510:
```suggestion
	.setName('ban')
```


slashie names can't have caps :O

guide/interactions/slash-commands.md
```diff
@@ -476,3 +476,42 @@ await interaction.reply('Pong!');
...
+```js {9}
+const { SlashCommandBuilder } = require('@discordjs/builders');
+const { Permissions } = require('discord.js');
+
+const data = new SlashCommandBuilder()
+	.setName('ban')
+	.setDescription('Ban a member!')
+	.addUserOption(option =>
+		option.setName('target').setDescription('The member to ban'))
+	.setDefaultMemberPermissions([Permissions.FLAGS.KICK_MEMBERS, Permissions.FLAGS.BAN_MEMBERS]);
```
BenjammingKirby, line 514: Just note that as of right now, this method does not take an array

vladfrangu: And never will! You can use `new Permissions([FLAGS.KickMembers, FLAGS.BanMembers].bitfield)` instead 😄

BenjammingKirby: ~~array.bitfield haha~~
i know the point of builders and how it is and why this wouldn't be possible on builders
but what could happen is djs can re-export builders where the method takes in a PermissionResolvable, right? then users can just import SlashCommandBuilder from djs itself

vladfrangu: I mean...yes, it could, but extending the whole class for one method only, especially for something that you really only do once doesn't feel right

BenjammingKirby, line 514:
```suggestion
	.setDefaultMemberPermissions(new Permissions([Permissions.FLAGS.KICK_MEMBERS, Permissions.FLAGS.BAN_MEMBERS]).bitfield);
```

monbrey: Wasn't the original point of builders to be able to be independent of discord.js, and usable for raw REST calls? Or at least only have a reliance on types?

Couldn't we represent this using something like 
```js
const { PermissionFlagsBits } = require("discord-api-types/v10");
// and
.setDefaultMemberPermissions(PermissionFlagsBits.KickMembers | PermissionFlagsBits.BanMembers)
```

monbrey: Maybe we'd have to, but I'd still prefer that as builders are supposed to be agnostic. Maybe it would be better if the builder method itself could accept an array of bits, but I don't think it does.

megatank58: It doesn't unfortunately, ill go with the bits method then and add a docs link to mdn with some pretext and a tip

lolollllo:
Builders now supports permissions v2, when will the PR be merged?

PR #1118, opened 2022-05-14
Fix prune.js in the code-samples

pickaxe828:
Fix the check number range logic for the `/prune` command
Located in `code-samples/creating-your-bot/command-handling/commands/prune.js`


almostSouji:
this needs a rebase, as it diverged from main when I fixed an erroneous deploy

almostSouji:
Thanks for finding the bug and suggesting a fix, applied in 6a90044

PR #1128, opened 2022-05-20
feat: add context menu

megatank58:
Co-authored-by: Ben  <88249114+BenjammingKirby@users.noreply.github.com>

Adds context menu page

Renaming branches apparently closes the PR so I make a new one

### DIFF
--- a/guide/interactions/context-menus.md
+++ b/guide/interactions/context-menus.md
@@ -37,7 +37,7 @@ You can get the targeted user by accessing the `targetUser` or `targetMember` pr
 client.on('interactionCreate', interaction => {
 	if (!interaction.isUserContextMenu()) return;
 	// Get the User's username from context menu
-	const username = interaction.targetUser.username;
-	console.log(username);
+	const name = interaction.targetUser.username;
+	console.log(name);
 });
 ```
--- a/guide/interactions/context-menus.md
+++ b/guide/interactions/context-menus.md
@@ -38,6 +38,6 @@ client.on('interactionCreate', interaction => {
 	if (!interaction.isUserContextMenu()) return;
 	// Get the User's username from context menu
 	const username = interaction.targetUser.username;
-	console.log(user);
+	console.log(username);
 });
 ```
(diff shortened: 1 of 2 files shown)


guide/interactions/context-menus.md
```diff
@@ -0,0 +1,43 @@
...
+const { ContextMenuCommandBuilder } = require('@discordjs/builders');
+
+const data = new ContextMenuCommandBuilder()
+	.setName('echo')
+	.setType('USER');
+```
+
+## Receiving context menu command interactions
+
+Context Menus are received via an interaction. You can check if a given interaction is a context menu by invoking the `isContextMenu()` method, you can use the `isMessageContextMenu()` and `isUserContextMenu()` methods to check for the specific type of context menu interaction:
```
vitorlops, line 23:
```suggestion
Context Menus are received via `interactionCreate` event. You can check if a given interaction is a context menu by invoking the `isContextMenu()` method, you can use the `isMessageContextMenu()` and `isUserContextMenu()` methods to check for the specific type of context menu interaction:
```

guide/interactions/context-menus.md
```diff
@@ -0,0 +1,43 @@
+# Context Menus
+
+Context Menus are application commands which appear on right clicking or tapping a user or a message.
```
vitorlops, line 3:
```suggestion
Context Menus are application commands which appear on right clicking or tapping an user or a message.
```

Mateo-tem: I think it's right as it is, just "a user". If I'm not wrong, "an" is used when the word starts with a vowel sound. It's a little bit confusing because "user" starts with "u", but when you pronounce "user" it starts with a "y".

vvito7: then just ignore it

guide/interactions/context-menus.md
megatank58: this was directly taken from https://discordjs.guide/interactions/modals.html#receiving-modal-submissions also if this change is necessary shouldn't it be `the interactionCreate event`

PR #1140, opened 2022-05-28
Update errors.md

FlatManDud:
a lot of people have this basic error


megatank58:
This is a basic JavaScript error and this occurs when you are trying to access the property of an object which is not present (null), the reason is inaccurate and this is too basic to be added in the guide 🤔

almostSouji (changes requested):
I think this is covered in the section "JavaScript errors" line 19ff.
If you think it's worth mentioning, feel free to add it as an example there
```
Example: `example1`, `example2`
```

almostSouji (approved):

PR #1144, opened 2022-06-01
Edit slash-commands.md to use discord.js (v14)

samast2010:
Updates slash-commands.md to require discord.js instead of @discordjs/builders, which has had its builders integrated into the main package (discord.js) in v14.


samast2010:
closed due to #1146 doing the things mentioned

Issue #1122, opened 2022-05-17
Raw payloads instead of builders


almostSouji:
### Is your feature request related to a problem? Please describe.

Advanced users may prefer not to use builders

### Describe the solution you'd like

Cover raw payloads as an option where builders are the only approach explained at the moment

### Describe alternatives you've considered

n/a

### Additional notes

n/a

PR #1149, opened 2022-06-06
feat(slashcommands): add JSON objects along with builders

megatank58:

#1122 Adding JSON objects as an alternative option to builders will allow advanced developers to ~~copy paste~~ understand code quickly and something that is more native to them.

One question is whether the builders page should also have the JSON objects for direct comparison with builders or should they not have them as the page is specifically made for builders?

### DIFF
--- a/guide/interactions/slash-commands.md
+++ b/guide/interactions/slash-commands.md
@@ -99,6 +99,9 @@ await rest.put(
 
 Application commands can have `options`. Think of these options as arguments to a function. You can specify them as shown below:
 
+:::: code-group
+
+::: code-group-item SlashCommandBuilder
 ```js {6-9}
 const { SlashCommandBuilder } = require('@discordjs/builders');
 
@@ -110,6 +113,27 @@ const data = new SlashCommandBuilder()
 			.setDescription('The input to echo back')
 			.setRequired(true));
 ```
+:::
+
+::: code-group-item JSON
+```js
+const data = {
+	name: 'echo',
+	description: 'Replies with your input!',
+	options: [
+		{
+			name: 'input',
+			description: 'The input to echo back',
+			type: 'STRING',
+			required: true,
+		},
+	],
+};
+
+```
+:::
+
+::::
 
 Notice how `.setRequired(true)` is specified within the options builder. Setting this will prevent the user from sending the command without specifying a value for this option!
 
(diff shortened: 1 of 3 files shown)


guide/interactions/slash-commands.md
```diff
@@ -110,6 +113,27 @@ const data = new SlashCommandBuilder()
...
+::: code-group-item JSON
+```js
+const data = {
+	name: 'echo',
+	description: 'Replies with your input!',
+	options: [
+		{
+			name: 'input',
+			description: 'The input to echo back',
+			type: 'STRING',
```
BenjammingKirby, line 127: just naming one out of the many, If you're actually going to use json, look at the docs, they don't take in strings, they take in numbers, the enum values
sure in v13, the applicationcommandmanager.create takes in the string i think but aren't you people discouraging this way of registering??

megatank58: Oh i completely missed that, you are right, the raw api takes numbers

BenjammingKirby:
I also disagree with having this in the first place, we don't need a raw json example for everything, that'd be ass to maintain and just unneeded stuff cluttering the guide imo. What I would rather do is recommend builders and link to the api docs and perhaps mention discord-api-types for typescript users. In the case of the "advanced users", that should be enough for them.

guide/interactions/slash-commands.md
```diff
@@ -110,6 +113,28 @@ const data = new SlashCommandBuilder()
...
+
+```js
+const data = {
+	name: 'echo',
+	description: 'Replies with your input!',
+	options: [
+		{
+			name: 'input',
+			description: 'The input to echo back',
+			type: 3,
```
BenjammingKirby, line 128: I recommended the usage of enums, is there a reason why those weren't used? They can be imported from discord-api-types
If you use rest, you need them for the routes, and rest uses it, builders uses it, so does discord.js
I don't think there's really a noticeable downside to using enums

almostSouji (changes requested):
Prefer enums over magic numbers, if you really feel the need to, you can note the value as a comment.

guide/interactions/slash-commands.md
```diff
@@ -181,6 +251,40 @@ const data = new SlashCommandBuilder()
...
+	options: [
+		{
+			name: 'user',
+			description: 'Info about a user',
+			type: ApplicationCommandOptionType.Subcommand,
+			options: [
+				{
+					name: 'target',
+					description: 'The user',
+					type: 'USER',
```
almeidx, line 272:
```suggestion
					type: ApplicationCommandOptionType.User,
```

almeidx (changes requested):

almostSouji:
This will need to be revisited after the semantic and structural rewrite that's currently in progress

megatank58:
It's probably better to remake this pr entirely (or rethink the idea as a whole) as the codebase has changed a lot.

PR #1151, opened 2022-06-13
change 'USER' to 2

ipwndwagons:
'user' doesnt work and it confuse me 😭

cause 	.setType('USER'); doesnt work

### DIFF
--- a/guide/interactions/context-menus.md
+++ b/guide/interactions/context-menus.md
@@ -15,7 +15,7 @@ const { ContextMenuCommandBuilder } = require('@discordjs/builders');
 
 const data = new ContextMenuCommandBuilder()
 	.setName('echo')
-	.setType('USER');
+	.setType(2);
 ```
 
 ## Receiving context menu command interactions
--- a/guide/interactions/context-menus.md
+++ b/guide/interactions/context-menus.md
@@ -12,10 +12,11 @@ To create a context menu you construct a new `ContextMenuCommandBuilder`. You ca
 
 ```js
 const { ContextMenuCommandBuilder } = require('@discordjs/builders');
+const { ApplicationCommandType } = require('discord-api-types/v9');
 
 const data = new ContextMenuCommandBuilder()
 	.setName('echo')
-	.setType(2);
+	.setType(ApplicationCommandType.User);
 ```
 
 ## Receiving context menu command interactions


guide/interactions/context-menus.md
```diff
@@ -15,7 +15,7 @@ const { ContextMenuCommandBuilder } = require('@discordjs/builders');
 
 const data = new ContextMenuCommandBuilder()
 	.setName('echo')
-	.setType('USER');
+	.setType(2);
```
thehairy, line 18:
```suggestion
	.setType(ApplicationCommandOptionTypes.USER);
```

guide/interactions/context-menus.md
```diff
@@ -15,7 +15,7 @@ const { ContextMenuCommandBuilder } = require('@discordjs/builders');
```
thehairy, line 15: idk if thats the right import tho
```suggestion
const { ApplicationCommandOptionTypes } = require('discord-api-types/v9')

```

thehairy (changes requested):
Should use ENUM instead.

megatank58:
this can also be left as it is until v14 rewrite, using enum here will be inconsistent with the rest of sections

almostSouji:
> this can also be left as it is until v14 rewrite, using enum here will be inconsistent with the rest of sections

as far as i am aware this is currently not valid with the version of builders the stable 13.8.0 uses, so it cannot be left as-is, even if it means introducing some inconsistency

guide/interactions/context-menus.md
megatank58: https://discord.js.org/#/docs/discord.js/stable/typedef/ApplicationCommandOptionType

Can't this be used instead here :thinking:

BenjammingKirby: builders exports a ContextMenuType enum right? just use that?

guide/interactions/context-menus.md
almeidx: `ApplicationCommandType.User`

guide/interactions/context-menus.md
thehairy: > https://discord.js.org/#/docs/discord.js/stable/typedef/ApplicationCommandOptionType
> 
> Can't this be used instead here 🤔

No, I have mistaken this for options, but its an ApplicationCommand type.

PR #1152, opened 2022-06-14
feat(autocomplete): remove `response` variable

megatank58:
The return type of AutocompleteInteraction#respond is void, this PR removes the response variable and no longer awaits the promise, hence also making the function synchronous.


almostSouji:
It's still best practice to handle and await promises, for example so you get a proper stack trace

almostSouji (changes requested):
It's still best practice to handle and await promises, for example so you get a proper stack trace

megatank58:
Reverted

PR #1163, opened 2022-06-28
chore: fix hyperlinks in .md file

lars-hs:


megatank58:
this will actually make the links go to https://discordjs.guide/guide/creating-your-bot/command-handling.html which is not where the section is

megatank58:
No this works completely fine in the discordjs.guide and doesn't needs changing at all, the links in MD files are according to the vuepress convention and not the GitHub one

lars-hs:
> No this works completely fine in the discordjs.guide and doesn't needs changing at all, the links in MD files are according to the vuepress convention and not the GitHub one

Oh that makes sense, I was taking this from reading the files on github, beginner mistale, thank you!

PR #1186, opened 2022-07-21
Added warning for discordjs v14 Intents

Stark-program:
Added a warning to let users know to require 'GatewayIntentBits' instead of 'Intents' when using discordjs v14+. 


Stark-program:
Awesome thank you!

PR #1146, opened 2022-06-03
Update guide to v14 code

lolollllo:
**Please describe the changes this PR makes and why it should be merged:**
- Updates all parts of the guide to v14 syntax.
- Updates all the code samples to v14 syntax.

PR #1193, opened 2022-07-22
Update to DIscord.js v14

Michaelg22:
Intents was changed to GatewayIntentBits for version 14 of DIscord.js.
The change was made to declared the bot intents.
simply an update from v13 to v14


megatank58:
#1146

PR #1195, opened 2022-07-22
feat(*): apply v14 changes (#1146 follow-up)

lolollllo:
**Please describe the changes this PR makes and why it should be merged:**

- Some changed in v14 were missed in #1146, this PR adds those changes

PR #1199, opened 2022-07-23
Update README.md

skyler252:
**Changed the part where it says if it outputs 16.6.0 to 16.9.0 since discord.js v14 requires 16.9.0 or higher.**


lolollllo:
Fixed in #1195

PR #1208, opened 2022-07-25
Update canvas guide to use loadImage

DevAndromeda:

**[@napi-rs/canvas](https://npm.im/@napi-rs/canvas)** supports the `loadImage` function since v0.1.25
and it is a lot easier to load image via `loadImage`:

```diff
- const imgData = getBufferSomehow(source);
- const image = new Canvas.Image();
- image.src = imgData;
- ctx.drawImage(image, ...);
+ const image = await Canvas.loadImage(source);
+ ctx.drawImage(image, ...)
```

`loadImage` function supports file path, url, data uri, streams, `Image` and buffer-like objects.

Edit:
I have replaced `canvas.toBuffer('image/png')` with `await canvas.encode('png')` because the `encode` method does the processing in a separate thread, without blocking the main event loop.

### DIFF
--- a/guide/popular-topics/canvas.md
+++ b/guide/popular-topics/canvas.md
@@ -136,7 +136,7 @@ client.on('interactionCreate', async interaction => {
 
 A bit plain, right? Fear not, for you have a bit more to do until you reach completion. Since this guide page's goal is focused more on actual code than design, let's place a basic square-shaped avatar for now on the left side of the image. In the interest of coverage, you will also make it a circle afterward.
 
-```js {5-12}
+```js {7-15}
 const { request } = require('undici');
 
 client.on('interactionCreate', async interaction => {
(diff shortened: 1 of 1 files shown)


guide/popular-topics/canvas.md
```diff
@@ -140,17 +136,13 @@ client.on('interactionCreate', async interaction => {
...
+```js {3-5}
 client.on('interactionCreate', async interaction => {
 	// ...
 	context.strokeRect(0, 0, canvas.width, canvas.height);
 
-	const { body } = await request(interaction.user.displayAvatarURL({ format: 'jpg' }));
-	const avatar = new Canvas.Image();
-	avatar.src = Buffer.from(await body.arrayBuffer());
-
+	const avatar = await Canvas.loadImage(interaction.user.displayAvatarURL({ format: 'jpg' }));
```
almeidx, line 144: This is actually a downgrade performance-wise, since `@napi-rs/canvas` uses the native `https` module under the hood
It might be a good idea to keep `undici.request()` and only update the `new Canvas.Image()` part

DevAndromeda: @almeidx I've made some changes, is it good now?

almeidx: Yes

almeidx (changes requested):

almeidx (approved):

PR #1216, opened 2022-07-27
Adding another code snippet

phyouthcenter1:
I noticed a code snippet that used config.json but no code snippet that was for the same thing that is for .env files so I added it.

### DIFF
--- a/guide/creating-your-bot/README.md
+++ b/guide/creating-your-bot/README.md
@@ -137,6 +137,8 @@ Open your code editor and create a new file. We suggest that you save the file a
 
 Here's the base code to get you started:
 
+:::: code-group
+::: code-group-item Config file
 ```js
 // Require the necessary discord.js classes
 const { Client, GatewayIntentBits } = require('discord.js');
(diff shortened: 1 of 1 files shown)


guide/creating-your-bot/README.md
```diff
@@ -153,6 +155,29 @@ client.once('ready', () => {
...
 ```
+:::
+::: code-group-item .env file
+```js
+const { Client, GatewayIntentBits } = require('discord.js');
+const dotenv = require('dotenv');
+
+dotenv.config();
+
+const { token } = process.env["DISCORD_TOKEN"]
```
almeidx, line 166: `process.env["DISCORD_TOKEN"]` would be a string, you can't destructure `token` out of it 
and besides, discord.js defaults the client token to that (`process.env.DISCORD_TOKEN`, if set), there's no need to pass it

almeidx (changes requested):

guide/creating-your-bot/README.md
phyouthcenter1: Yeah I figured that out like right before Is aw that reply lol. I had to find out the hard way when my code didnt work 💀

PR #1228, opened 2022-08-10
refactor(faq): use consistent examples

Squiddleton:
The original section for "How do I unban a user?" uses `<CommandInteractionOptionResolver>.get(...).value` instead of `<CommandInteractionOptionResolver>.getUser(...)` which is inconsistent with the nearby section regarding banning users.  This corrects that, and it rewrites the tip that was originally based on message commands instead of slash commands.

### DIFF
--- a/guide/popular-topics/faq.md
+++ b/guide/popular-topics/faq.md
@@ -21,12 +21,12 @@ guild.members.ban(user);
 ### How do I unban a user?
 
 ```js
-const id = interaction.options.get('target')?.value;
-guild.members.unban(id);
+const user = interaction.options.getUser('target');
+guild.members.unban(user);
 ```
 
 ::: tip
-Because you cannot ping a user who isn't in the server, you have to pass in the user id. To do this, we use a <DocsLink path="typedef/CommandInteractionOption" />. See [here](/interactions/slash-commands.html#parsing-options) for more information on this topic.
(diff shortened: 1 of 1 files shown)


Jiralite (approved):

guide/popular-topics/faq.md
```diff
@@ -21,12 +21,12 @@ guild.members.ban(user);
...
 ```js
-const id = interaction.options.get('target')?.value;
-guild.members.unban(id);
+const user = interaction.options.getUser('target');
+guild.members.unban(user);
 ```
 
 ::: tip
-Because you cannot ping a user who isn't in the server, you have to pass in the user id. To do this, we use a <DocsLink path="typedef/CommandInteractionOption" />. See [here](/interactions/slash-commands.html#parsing-options) for more information on this topic.
+The <DocsLink path="class/CommandInteractionOptionResolver?scrollTo=getUser" /> method parses user ids into <DocsLink path="class/User" /> instances even if the target is not in the server where the command is used.
```
almostSouji, line 29:
```suggestion
Discord validates and resolves user ids for users not on the server in user slash command options. To retrieve and use the full structure from the resulting interaction, you can use the <DocsLink path="class/CommandInteractionOptionResolver?scrollTo=getUser" /> method.
```

Semantical nit, it's not the getUser method that does anything special, it's discord that does the magic here. If the method did the resolving this would rely on the user cache, but in reality, retrieves it from the interaction payload

almostSouji (changes requested):

almostSouji (approved):

PR #1230, opened 2022-08-12
feat(AuditLogs): add `Who unbanned user?`

dager-mohamed:

Adding more info about Auditlogs like who unbanned user.

### DIFF
--- a/guide/popular-topics/audit-logs.md
+++ b/guide/popular-topics/audit-logs.md
@@ -143,3 +143,40 @@ client.on('guildBanAdd', async ban => {
 	}
 });
 ```
+## Who unbanned user?
+Similar to `guildBanAdd` but we will use `guildBanRemove` instead of `guildBanAdd` event.
+
+```js
+client.on('guildBanRemove', member => {
+	console.log(`${member.user.tag} was unbanned from the guild ${member.guild.name}`);
+});
+```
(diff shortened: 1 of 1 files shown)


guide/popular-topics/audit-logs.md
```diff
@@ -143,3 +143,40 @@ client.on('guildBanAdd', async ban => {
...
+});
+```
+
+In this example you will see what exactly happened, and who unbanned user, we will fetch the audit log with `MEMBER_BAN REMOVE`, The `guildBanRemove` listener then becomes:
+
+```js {2-5,7-8,10-11,13-15,17-23}
+client.on('guildBanRemove', async member => {
+	const fetchedLogs = await member.guild.fetchAuditLogs({
+		limit: 1,
+		type: 'MEMBER_BAN_REMOVE',
```
imranbarbhuiya, line 161:
```suggestion
		type: AuditLogEvent.MemberBanRemove,
```

imranbarbhuiya (changes requested):

almostSouji:
Where is the added value to the guide here, compared to banning? Isn't it just another type?
I don't plan on explaining each and every audit log type with an example; explaining it on the example of one should be enough if it's not adding any new nuance to it

PR #1235, opened 2022-08-23
Changed from request to fetch

Juoper:
Changed from using request to using fetch so it works again


almostSouji:
I'd prefer fixing the issue and implementing request properly over switching it to an experimental method.

PR #1237, opened 2022-08-28
Fix slash command guide being inaccurate

sahmyooI:
This caused me quite a bit of headache! `const data` is not defined under the scope of the file, so the external 'commands.js' file cannot read it
`this.data` fixes this.

### DIFF
--- a/guide/interactions/slash-commands.md
+++ b/guide/interactions/slash-commands.md
@@ -98,7 +98,7 @@ Application commands can have `options`. Think of these options as arguments to 
 ```js {6-9}
 const { SlashCommandBuilder } = require('discord.js');
 
-const data = new SlashCommandBuilder()
+this.data = new SlashCommandBuilder()
 	.setName('echo')
 	.setDescription('Replies with your input!')
 	.addStringOption(option =>
@@ -143,7 +143,7 @@ Specify them by using the `addChoices()` method from the slash command builder:
 ```js {10-14}
 const { SlashCommandBuilder } = require('discord.js');
 
-const data = new SlashCommandBuilder()
+this.data = new SlashCommandBuilder()
 	.setName('gif')
 	.setDescription('Sends a random gif!')
 	.addStringOption(option =>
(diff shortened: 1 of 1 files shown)


almostSouji:
It took me quite some time to understand the underlying problem here.

In the "handling commands" section we attach the data to be deployed to the respective export, so it is packed with its execution.
The slash command section provides more examples of using command builders, but defines them as a `const` for demonstration purposes.

It is expected that you still just use them as demonstrated in the other guide section. Just define it on the respective export, in the `data` key, as before.

We could potentially remedy this by using it in the same context as previous sections, for example

```js
module.exports = {
	data: new SlashCommandBuilder()
	    .setName('echo')
	    .setDescription('Replies with your input!')
	    .addStringOption(option =>
		    option.setName('input')
			    .setDescription('The input to echo back')
			    .setRequired(true));
	async execute(/* ... */) {
		// ...
	},
};
```

Overwriting them using `this` is unnecessary and this is not really a fix.

sahmyooI:
> It took me quite some time to understand the underlying problem here.
> 
> In the "handling commands" section we attach the data to be deployed to the respective export, so it is packed with its execution. The slash command section provides more examples of using command builders, but defines them as a `const` for demonstration purposes.
> 
> It is expected that you still just use them as demonstrated in the other guide section. Just define it on the respective export, in the `data` key, as before.
> 
> We could potentially remedy this by using it in the same context as previous sections, for example
> 
> ```js
> module.exports = {
> 	data: new SlashCommandBuilder()
> 	    .setName('echo')
> 	    .setDescription('Replies with your input!')
> 	    .addStringOption(option =>
> 		    option.setName('input')
> 			    .setDescription('The input to echo back')
> 			    .setRequired(true));
> 	async execute(/* ... */) {
> 		// ...
> 	},
> };
> ```
> 
> Overwriting them using `this` is unnecessary and this is not really a fix.

This should definitely be clearer in some way then. There's a hyperlink earlier on that links to this section of the guide, so clicking that, you would never have seen the earlier section which implied you use the module.exports. This makes it hard to understand and will give you errors if you are not following the guide step-by-step.

PR #1242, opened 2022-08-30
chore: Mention deleteMessageSeconds instead of deleteMessageDays

Jiralite:
`deleteMessageDays` is deprecated. Users should be using `deleteMessageSeconds` instead: https://github.com/discordjs/discord.js/pull/8326.

### DIFF
--- a/guide/additional-info/changes-in-v14.md
+++ b/guide/additional-info/changes-in-v14.md
@@ -280,7 +280,7 @@ The `threadMembersUpdate` event now emits the users who were added, the users wh
 
 ### GuildBanManager
 
-The `days` option when banning a user has been removed. Use `deleteMessageSeconds` instead.
+The `days` option when banning a user has been removed. Use `deleteMessageSeconds` instead which, as named, takes seconds instead of the number of days.
 
 ### Guild
 
--- a/guide/additional-info/changes-in-v14.md
+++ b/guide/additional-info/changes-in-v14.md
@@ -280,7 +280,7 @@ The `threadMembersUpdate` event now emits the users who were added, the users wh
 
 ### GuildBanManager
 
-Staring from 14.4.0, developers should utilise `deleteMessageSeconds` instead of `days` and `deleteMessageDays`:
+Starting from 14.4.0, developers should utilise `deleteMessageSeconds` instead of `days` and `deleteMessageDays`:
 
 ```diff
 <GuildBanManager>.create('123456789', {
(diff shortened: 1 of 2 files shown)


guide/additional-info/changes-in-v14.md
```diff
@@ -280,7 +280,7 @@ The `threadMembersUpdate` event now emits the users who were added, the users wh
 
 ### GuildBanManager
 
-The `days` option when banning a user has been renamed to `deleteMessageDays` to be more aligned to the API name.
+The `days` option when banning a user has been removed. Use `deleteMessageSeconds` instead which, as named, takes seconds instead of the number of days.
```
megatank58, line 283:
```suggestion
The `days` option when banning a user has been deprecated. Use `deleteMessageSeconds` instead which, as named, takes seconds instead of the number of days.
```

megatank58 (changes requested):

almostSouji:
blocking this until it actually works as advertised, which apparently it does not at the moment

guide/additional-info/changes-in-v14.md
Jiralite: Ah yeah, that's true. When this becomes available for 13.x and 14.x, it would already be deprecated for new users switching over from 13.x to 14.x - it's not a new deprecation. Maybe "The `days` option in version 14 still deprecated" or something along the lines of that?

megatank58: Eh i don't think it would be appropriate to mention that a property is still deprecated, should be probably fine to say that it has been deprecated, maybe add it in v13 changes too?

Jiralite: What about just being more thorough?

> Since 14.4.0, developers should utilise `deleteMessageSeconds` instead of `days` and `deleteMessageDays`:
> 
> ```diff
> <GuildBanManager>.create({
> -  days: 3
> -  deleteMessageDays: 3
> +  deleteMessageSeconds: 3 * 24 * 60 * 60
> });
> ```
> 
> `deleteMessageDays` (introduced with version 14) and `days` are both deprecated and will be removed in the future.

As for the version 12 to 13 update guide, I guess I could use an adjusted version of the above?

megatank58: seems good to me

Jiralite:
Seems resolved @almostSouji! https://github.com/discord/discord-api-docs/issues/5439

almostSouji (approved):

PR #1247, opened 2022-08-31
👾 replace right click -> right-click

Rnbsov:
I think it will be slightly more correct


megatank58:
Is this really necessary, looks just fine either way

Rnbsov:
> Is this really necessary, looks just fine either way

I guess it will look more beautiful anyway

megatank58:
I would argue both methods of writing it are common and this change is unnecessary

almostSouji:
Even the various dictionaries aren't in unison about this, we'll keep it as-is

PR #1252, opened 2022-09-01
We have changed links, etc.

gx1285:
Changed v13 notations, links, etc. to v14.
Changed stable documentation links to v12 documentation links.


megatank58:
If we redirect users directly from v12 to v14 wouldn't that make them miss a lot of changes between the v12-v14 since v14 only has v13-v14 changes documented

almostSouji:
13 is still supported and gets bugfixes and backports

PR #1263, opened 2022-09-14
feat(Interactions): expand and refactor commands, components and modals

monbrey:
The guides for commands, components, modals etc all read like an extra wordy version of documentation. They detail every method, option and parameter without doing a great job of building on concepts or sticking with consistent examples.

This draft PR is going to work on revising these sections to be more of a guide, explaining to users *why* things are being done a certain way and using consistent, "real-world" examples that are built upon continually with new concepts.

At this point, the draft PR is just so I can see the deploy preview - I'm open to ideas and suggestions on what people would like to see in these sections but am not looking for line-by-line reviews yet.


monbrey:
Closing this for now since I think it's going to change in scope pretty significantly (and I fixed running it locally so I don't need deploy previews)

PR #1276, opened 2022-09-22
Revised explanation of getting guild id

asummers1:
The previous explanation to get a guild id was unclear. I made the description more specific.

### DIFF
--- a/guide/creating-your-bot/creating-commands.md
+++ b/guide/creating-your-bot/creating-commands.md
@@ -32,7 +32,11 @@ Below is a deployment script you can use. Focus on these variables:
 - `commands`: An array of commands to register. The [slash command builder](/popular-topics/builders.md#slash-command-builders) from `discord.js` is used to build the data for your commands
 
 ::: tip
-In order to get your application's client id, go to [Discord Developer Portal](https://discord.com/developers/applications) and choose your application. Find the id under "Application ID" in General Information subpage. To get guild id, open Discord and go to your settings. On the "Advanced" page, turn on "Developer Mode". This will enable a "Copy ID" button in the context menu when you right-click on a server icon, a user's profile, etc.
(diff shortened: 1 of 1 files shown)


guide/creating-your-bot/creating-commands.md
```diff
@@ -32,7 +32,11 @@ Below is a deployment script you can use. Focus on these variables:
...
 
 ::: tip
-In order to get your application's client id, go to [Discord Developer Portal](https://discord.com/developers/applications) and choose your application. Find the id under "Application ID" in General Information subpage. To get guild id, open Discord and go to your settings. On the "Advanced" page, turn on "Developer Mode". This will enable a "Copy ID" button in the context menu when you right-click on a server icon, a user's profile, etc.
+In order to get your application's client id, go to [Discord Developer Portal](https://discord.com/developers/applications) and choose your application. Find the id under "Application ID" in General Information subpage. To get guild id, open Discord and go to your settings. On the "Advanced" page, turn on "Developer Mode". Then, exit that window, right-click the name of your guild, and click the "Copy ID" button.
+:::
+
+::: warning
+Note that your channel ids are not the same as your guild id. Make sure to get the id from the latter; otherwise, you will get a "Missing Access" error upon trying to run your file.
 :::
```
Jiralite, line 40: This feels pretty redundant, doesn't it? A channel is in no way a guild and I don't see how that can be confusing.

asummers1: Not necessarily to the layman who doesn't know what a guild is; the Discord UI from the user's end uses the term "server", not "guild".  I assumed until recently that they were the same as channels.

I think this could be very confusing to the newcomer. In fact, the article seems to agree with this by explicitly describing what a "guild" is in the later part of the section concerning server info commands. If we assume that the user doesn't know what a guild is until that section, then we should be more clear earlier on what exactly they should be obtaining when they want a "guild id".

almostSouji:
The commands section of the guide is currently being rewritten. If you want to suggest different wording, please do so in #1281

PR #1280, opened 2022-10-01
Add tip to show '!' is unnecessary to show a nickname

GodderE2D:

Added a tip to show that an exclamation mark after a user's mention is no longer needed if you want to show their nickname - Discord automatically shows their nickname if they have one now regardless if a `!` is present.

For reference that it's deprecated:

- https://github.com/discord/discord-api-docs/issues/4734

P.S. Now that I'm reading my own wording, some could be rephrased.


almostSouji:
After internal discussion, the mention argument section will be removed instead - slash commands are by design positional and don't have any of the caveats explained in this section and should be preferred by now.

PR #1282, opened 2022-10-05
fix: use `UserManager#send()` for sending a DM

Idris1401:

It's better to use `UserManager#send()` instead of fetching the user and then use `User#send()`.

### DIFF
--- a/guide/popular-topics/faq.md
+++ b/guide/popular-topics/faq.md
@@ -135,8 +135,8 @@ channel.threads.create({ name: 'Post name', message: { content: 'Message content
 <!-- eslint-skip -->
 
 ```js
-const user = await client.users.fetch('id');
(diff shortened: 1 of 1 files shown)


guide/popular-topics/faq.md
```diff
@@ -135,8 +135,8 @@ channel.threads.create({ name: 'Post name', message: { content: 'Message content
 <!-- eslint-skip -->
 
 ```js
-const user = await client.users.fetch('id');
-user.send('content');
+const message = await client.users.send('id', 'content');
+console.log(message);
```
almostSouji, line 139: the logging is inconsistent with the rest of the faq

almostSouji (changes requested):

almostSouji (approved):

PR #1286, opened 2022-10-06
fix: remove the usage of @discordjs/builders

samast20000:
Removes the usage of `@discordjs/builders` in `builders.md`


megatank58:
this change shouldn't be made, builders can be used without djs eg for making interaction only botd

almostSouji:
I agree that we should be importing them from discord.js across the guide - which should already be the case, but the change mentioning the package is valid

PR #1292, opened 2022-10-11
Numbers: Numerical vs written representation

KinectTheUnknown:
Description: The word "ten" should be spelled out in the last bullet of the "Embed Limits" page section since the section uses numbers frequently in the other bullet points.

Explanation: I believe that the original writer was following the rule where whole numbers from one to ten, but there's another rule that consistency is more important and overrides all other rules[^1]. This rule should only be applied to this specific case because the other numbers in the "Embed Limits" section are written using numerical representation, making the "ten" stand out awkwardly.

[^1]: https://www.scribendi.com/academy/articles/when_to_spell_out_numbers_in_writing.en.html


megatank58:
Or maybe all other numbers can be changed! /j

PR #1306, opened 2022-10-19
Fixed the wrong URL at Resulting Code

okunamayanad:
The Resulting Code button sends to the wrong link!
Fixed it and changed it to Command Handling resulting code


almostSouji:
Please review the other code samples introduced in the same PR, so we can change them all at once, if necssary

monbrey (approved):
Looks like it's just this one from what I can see, thanks!

PR #1314, opened 2022-10-20
A Note for STEP-BY-STEP Users

Tarvey:
A Note, for S-B-S Users thats quite hard for beginners

### DIFF
--- a/guide/slash-commands/advanced-creation.md
+++ b/guide/slash-commands/advanced-creation.md
@@ -22,6 +22,9 @@ const data = new SlashCommandBuilder()
 		option.setName('input')
 			.setDescription('The input to echo back'));
 ```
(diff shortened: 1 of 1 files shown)


jaw0r3k:
i think it's useless

megatank58:
This feels like basic JavaScript, if anything and if really, it should be added to the errors page and not here, you should not be adding random commas

Tarvey:
> This feels like basic JavaScript, if anything and if really, it should be added to the errors page and not here, you should not be adding random commas

random commas? if you follow step by step, in one of the code blocks, theres a comma, thats why

Jiralite:
This is far too basic.

monbrey:
So fix the code block if there's a comma that shouldn't be there?

Don't add a section to explain if it was wrong, make it right.

Issue #1318, opened 2022-10-21
Mention values property in Select menus chapter


Sheepolution:
### Is your feature request related to a problem? Please describe.

When reading the Select menus chapter to figure out how they work, I am not educated on how to get the value(s) of the selected option(s).

### Describe the solution you'd like

Mention that you can get the selected value(s) with `.values`.

zayKenyon:
Was going to comment how this have been automatically closed, but Souji explains why it wasn't: https://github.com/discordjs/guide/issues/959#issuecomment-1295604321

PR #1320, opened 2022-10-22
feat(select-menus): add access select menu values

AircraftOverviewer:
Addresses issue #1318 by adding a rudimentary section for accessing [SelectMenuInteraction#values](https://discord.js.org/#/docs/discord.js/main/class/SelectMenuInteraction?scrollTo=values) upon receiving a select menu interaction in the style of the guides

### DIFF
--- a/guide/interactions/select-menus.md
+++ b/guide/interactions/select-menus.md
@@ -195,36 +195,6 @@ client.on(Events.InteractionCreate, async interaction => {
 });
 ```
 
-## Accessing select menu interaction values
-
-After receiving your <DocsLink path="class/SelectMenuInteraction"/> you will be able to access the selected values from <DocsLink path="class/SelectMenuInteraction?scrollTo=values"/>. This will return an array of string values associated with the selected options in your select menu.
-
-```js {4,6}
-client.on(Events.InteractionCreate, async interaction => {
-	if (!interaction.isSelectMenu()) return;
-
-	const selected = interaction.values.join(', ');
-
-	await interaction.update(`The user selected ${selected}!`);
-});
-```
-
-The snippet below accesses the first index of the returned array to demonstrate how single-select menu values can be handled:
-
-```js {4,6-10}
-client.on(Events.InteractionCreate, async interaction => {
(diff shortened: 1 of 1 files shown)


guide/interactions/select-menus.md
```diff
@@ -195,6 +195,36 @@ client.on(Events.InteractionCreate, async interaction => {
...
+client.on(Events.InteractionCreate, async interaction => {
+	if (!interaction.isSelectMenu()) return;
+
+	const selected = interaction.values.join(', ');
+
+	await interaction.update(`The user selected ${selected}!`);
+});
+```
+
+The snippet below accesses the first index of the returned array to demonstrate how single-select menu values can be handled:
```
monbrey, line 212:
```suggestion
By default, select menus only accept a single selection. You can retrieve the selected value by accessing the first index of the returned array, as demonstrated in the snippet below:
```

guide/interactions/select-menus.md
```diff
@@ -195,6 +195,36 @@ client.on(Events.InteractionCreate, async interaction => {
 });
 ```
 
+## Accessing select menu interaction values
+
+After receiving your <DocsLink path="class/SelectMenuInteraction"/> you will be able to access the selected values from <DocsLink path="class/SelectMenuInteraction?scrollTo=values"/>. This will return an array of string values associated with the selected options in your select menu.
+
+```js {4,6}
```
monbrey, line 202:
```suggestion
In the case of a multi-select menu, the received <DocsLink path="class/SelectMenuInteraction?scrollTo=values"/> may contain more than one value, and should be handled accordingly:

```js {4,6}
```

monbrey (changes requested):
I think the examples would be better swapped around. Since single-select is the default behaviour for select menus we should probably show that first.

Creating multi-selects isn't covered until the next sub-heading, so I'd probably suggest either moving the multi-select receiving example into that section, or moving this whole section below it.

My suggestions for wording changes are based on this reordering, keeping your first paragraph as the intro to this section.

monbrey (approved):

PR #1328, opened 2022-10-29
Add correct link to Buttons page

mrlovett:
The "button components section" link has a self-reference to the current page due to an empty URL reference. Added a reference to /interactions/buttons.html.


monbrey:
Ahh, good catch, but this whole line is actually kind of incorrect. I must have written it in preparation for the buttons rewrite I was working on, but that hasn't been completed and merged in yet.

There isnt a good confirmation workflow on that page at all 😁

mrlovett:
If you're happy for this link to stay blank for now I'll close this PR 🙂

PR #1329, opened 2022-10-29
chore: use the same syntax as node-fetch

Idris1401:
I think it'll be better if the guide keeps the same syntax syntax as before when `node-fetch` was used.
Like: `const { file } = await fetch('https://aws.random.cat/meow').then(response => response.json())`


megatank58:
this doesn't really needs to be changed, the only issue here *could* be declaring two variables for a single job but i don't think this is really needed

PR #1330, opened 2022-10-30
Moved warning for better visibility.

Townsy45:
This page is mis-leading because it shows you the syntax but this does not work for typescript and you have to scroll down two monitors worth of the page to get a little warning at the end that you need something for typescript.

This has thrown me and others off because it is unclear, moving this to the top will stop confusion.

*You can stretch the page over two monitors and barely see the warning!*


monbrey (approved):
Seems fine - still planning to rework this section but I'll keep in mind making the TypeScript warning more prominent when I do.

PR #1333, opened 2022-11-02
Fix minor code issue.

ThisIsACoolBoy:
There is a small bracket mistake in section- Required Choices where another bracket was added. This can arise a few issues.

This would be useful for a few people who might not see small and minor mistakes that can arise to some issue, and that it will make the guide for advanced commands a bit more reliable. (at least that's what I think)


almostSouji:
It closes the `addStringOption()` call.

PR #1348, opened 2022-11-11
Add more example for sweeper element.

FabienBounoir:
the use of the sweeper is not documented enough, so I made an example of how the filter works in the sweeper.

### DIFF
--- a/guide/miscellaneous/cache-customization.md
+++ b/guide/miscellaneous/cache-customization.md
@@ -85,7 +85,7 @@ const client = new Client({
 		},
 		users: {
 			interval: 3600, // Every hour...
-			filter: user => user.bot, // Remove all bots.
+			filter: user => user.bot && user.id !== client.user.id, // Remove all bots.
 		},
 	},
 });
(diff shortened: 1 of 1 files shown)


jaw0r3k:
What about keeping but itself in cache?

FabienBounoir:
> What about keeping itself in cache?
I just submitted the idea to show explicitly how to manage the cache with the filter method in the sweeper

jaw0r3k:
But it can be better lol

FabienBounoir:
> 

yes

guide/miscellaneous/cache-customization.md
```diff
@@ -83,6 +83,10 @@ const client = new Client({
 			interval: 3600, // Every hour...
 			lifetime: 1800,	// Remove messages older than 30 minutes.
 		},
+		users: {
+			interval: 3600, // Every hour...
+			filter: user => user.bot, // Remove all bots.
```
megatank58, line 88:
```suggestion
			filter: user => user.bot && user.id !== client.user.id, // Remove all bots.
```

almostSouji (approved):

PR #1349, opened 2022-11-12
Mistake in embed mentions tip

vunsh:
Updates a mistake in a tip for mentions in embeds. Mentions only resolve in embed descriptions and field values, not titles.


megatank58:
Mentions in title actually do resolve on mobile devices, just not on desktop

PR #1351, opened 2022-11-14
fix(code-samples): remove unused line

V3N0M-Z:
Remove unused line:
```js
const { Op } = require('sequelize');
```


Jiralite:
In the first file you removed this from, it _is_ being used:

https://github.com/discordjs/guide/blob/27f4bb718adc74c1b240bebb0cbd4fe7cee58623/code-samples/sequelize/currency/14/app.js#L74

In the second file, all imports are retained as the rest of the has been omitted.

PR #1365, opened 2022-12-05
Removal of Number Intent Example

vunsh:
It is often advised that users do not use a number for intents. This is because users often use a number to represent "all intents", not realizing the number gets outdated with the introduction of other intents. The majority of people who are reading the guide at this point would benefit from listing out all their intents as strings/enums not only for readability, but for debugging purposes as well. This would help people figure out why they aren't collecting certain data, things they are missing, etc. I feel the blurb about bitfields directly under this code example is more than sufficient. The problem with having a code example is blind copy pastes, and there is no good reason to include it.

tldr: intent number bad coding practice

### DIFF
--- a/guide/popular-topics/intents.md
+++ b/guide/popular-topics/intents.md
@@ -44,7 +44,7 @@ myIntents.add(Intents.FLAGS.GUILD_PRESENCES, Intents.FLAGS.GUILD_MEMBERS);
 
 const client = new Client({ intents: myIntents });
 
-// other examples:
+// other example:
 
 const otherIntents = new Intents([Intents.FLAGS.GUILDS, Intents.FLAGS.DIRECT_MESSAGES]);
 otherIntents.remove([Intents.FLAGS.DIRECT_MESSAGES]);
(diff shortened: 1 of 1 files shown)


jaw0r3k:
is v13 still supported?

vunsh:
> is v13 guide still supported?
> 
> 

afaik yea

megatank58:
this literal example is for bitfields, this should not be removed, it displays the function itself

vunsh:
> this literal example is for bitfields, this should not be removed, it displays the function itself

Anyone with any knowledge of bitfields would be able to go to the docs or read the blurb above this code example and easily figure out implementation. The point is that it is very unlikely that someone who actually understands what bitfields represent and how to properly use them would be reading this portion of the guide. 

Additionally, I was motivated to make this pr when I saw someone use code from this example in a djs support channel. It is always better to just list out your intents so you are aware of what you have, and it makes sense to teach that in a beginner djs guide.

megatank58:
> Anyone with any knowledge of bitfields would be able to go to the docs or read the blurb above this code example and easily figure out implementation. The point is that it is very unlikely that someone who actually understands what bitfields represent and how to properly use them would be reading this portion of the guide. 

well everything can be looked up in the docs, the guide (from what i think) acts as a medium to show the commonly used methods that would generally be used or to expose the methods in a better practical manner. this change shouldn't happen as it shows the basic functionality of the method of bigints with intents itself, that's the point of this whole section. if something really needs to be done then a warning can be added but removing a whole section because someone in support quoted this code doesn't makes sense

PR #1369, opened 2022-12-08
Edited, where exclamation mark was after quote.

Neodevils:
**Please describe the changes this PR makes and why it should be merged:**
A miss typing, which would occur an error.

PR #1392, opened 2022-12-31
change a typo

Lux64:
on the interaction.reply() demonstration, there is a typo where an exclamation mark is after the string ends

idk how github works so im just leaving it like that


Jiralite:
Duplicate of #1369.

PR #1403, opened 2023-01-09
Replace maxcdn with cloudflare

CorwinDev:
As MaxCDN is down permanently I replaced it with cloudflares CDN


Jiralite (changes requested):
It seems the links are up again. Following the conversation in https://github.com/twitter/twemoji/issues/580, it seems the recommendation is to use jsDelivr.

CorwinDev:
So update to jsdeliver? Or just let the old urls be in there

Jiralite:
The links are up, but they all redirect to jsDelivr, so I'd modify them all to jsDelivr now.

Issue #1340, opened 2022-11-07
Incorrect information in Slash command permissions regarding permission requirements


Sheepolution:
### Describe the bug

In [this article](https://discordjs.guide/slash-commands/permissions.html#member-permissions) the following is stated:

> For a kick command however, we can allow members with either BanMembers or KickMembers to execute the command, so we'll list both flags here joined by the bitwise OR operator.

This is incorrect, as Discord does not support one-of permission requirements for Slash commands. The example `.setDefaultMemberPermissions(PermissionFlagsBits.BanMembers | PermissionFlagsBits.KickMembers);` will require the member to have both the ban BanMembers and KickMembers permission.

### To Reproduce

1. Create a slash command with `.setDefaultMemberPermissions(PermissionFlagsBits.BanMembers | PermissionFlagsBits.KickMembers);`
2. Give a member only the KickMembers permission.
3. They are not allowed to use the command, as they miss the BanMembers permission.

### Device (please complete the following information)

N.A.

monbrey:
Thanks, this makes sense, not sure what example I was going for when I wrote this. I'll review and make changes.

PR #1407, opened 2023-01-18
Fix slash commands permissions description

MichalD96:
[Using Default Permissions](https://discord.com/developers/docs/interactions/application-commands#application-command-permissions-object-using-default-permissions) made by Discord describes that all permissions separated by bitwise "or" operator are required to execute the command, in discord.js guide was that required is either one permission.

### DIFF
--- a/guide/slash-commands/permissions.md
+++ b/guide/slash-commands/permissions.md
@@ -28,7 +28,7 @@ const data = new SlashCommandBuilder()
 	.setDefaultMemberPermissions(PermissionFlagsBits.BanMembers);
 ```
 
-We can require user to have both `BanMembers` and `KickMembers` permissions to execute the command, so we'll list both flags here joined by the bitwise OR operator.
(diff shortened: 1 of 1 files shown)


MichalD96:
I just found there is an issue opened about this #1340

guide/slash-commands/permissions.md
```diff
@@ -28,7 +28,7 @@ const data = new SlashCommandBuilder()
 	.setDefaultMemberPermissions(PermissionFlagsBits.BanMembers);
 ```
 
-For a kick command however, we can allow members with either `BanMembers` or `KickMembers` to execute the command, so we'll list both flags here joined by the bitwise OR operator.
+You can require user to have both `BanMembers` and `KickMembers` permissions to execute the command, so we'll list both flags here joined by the bitwise OR operator.
```
megatank58, line 31:
```suggestion
For a kick commans however, you can require the user to have both `BanMembers` and `KickMembers` permissions to execute the command, so we'll list both flags here joined by the bitwise OR operator.
```
matches the phrasing used in the section above this

PR #1419, opened 2023-02-22
fix: Update expired link

JannekKn:
the old link to "the previous page" went to https://discordjs.guide/slash-commands/creating-commands.html 
that page does not exist (anymore).
This is the closest page that would make sense imo


jaw0r3k:
how ever your pr name should match the format
`fix(Deleting commands): Update expired link` or smth

PR #1439, opened 2023-03-30
Clarified Localizations

LifeRP-FiveM:
Added more details on how to use localizations correctly in order to make it work.


almostSouji:
This is not true; you generally do not need to quote object keys in JS, unless they do not fulfill the naming requirements for literals.
In JSON however, object keys need to be quoted.

Here, we define the keys within a JS context.

PR #1446, opened 2023-04-05
fix(cooldowns): command#data#name and relative timestamp fix

AmNobCop:
Updated how the message to inform the user of the amount of time they need to wait before using the command again looks. Now it looks better with the format of the unix timestamp. Also added more clarity to add the code to the InteractionCreate event, not just anywhere in the main file.
 
@Joeyk710 - Helped with changes

PR #1447, opened 2023-04-05
Update index.js

AmNobCop:
Updated how the message to inform the user of the amount of time they need to wait before using the command again looks. Now it looks better with the format of the unix timestamp. (https://github.com/discordjs/guide/pull/1446) 

Also updated ${command.name} to ${command.data.name} so it will not return undefined. This request is just to update the resulting code with the guide. 

@Joeyk710 - Helped with changes

### DIFF
--- a/code-samples/additional-features/cooldowns/index.js
+++ b/code-samples/additional-features/cooldowns/index.js
@@ -51,7 +51,7 @@ client.on(Events.InteractionCreate, async interaction => {
 
 		if (now < expirationTime) {
 			const expiredTimestamp = Math.round(expirationTime / 1000);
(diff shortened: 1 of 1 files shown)


Jiralite:
If `command.name` is possibly `undefined`, there's a lot more that needs updating.

jaw0r3k:
and you can use just one pr

joeyk710:
The current guide uses `command.data.name` so I'm not sure if `command.name` was originally used.

As shown from the guide for reference
```js
return interaction.reply({ content: `Please wait <t:${expiredTimestamp}:R> more second(s) before reusing the \`${command.data.name}\` command.`, ephemeral: true });
```

AmNobCop:
> If `command.name` is possibly `undefined`, there's a lot more that needs updating.

there's no name property in command.name, its inside data

Jiralite:
You should use the other pull request to update this code instead of creating multiple. I'll close this, and please commit to your other branch.

PR #1453, opened 2023-04-13
fix(command-deployment): folder pathing variable fix

ZoraDrip:
Hi! I'm going through the examples from scratch and ran into a few issues above:

- `foldersPath` was undefined
- I moved the REST() into the async, otherwise I got `TypeError: (intermediate value).setToken(...) is not a function`

I'm using Node 18.15.0 on Windows 11. After the above changes, the tutorial works again

# Screenshots of deploy errors

image

image

### DIFF
--- a/guide/creating-your-bot/command-deployment.md
+++ b/guide/creating-your-bot/command-deployment.md
@@ -51,7 +51,7 @@ const path = require('node:path');
 
 const commands = [];
 // Grab all the command files from the commands directory you created earlier
-const commandsPath = path.join(__dirname, 'commands');
+const foldersPath = path.join(__dirname, 'commands');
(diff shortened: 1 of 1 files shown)


Jiralite:
> * I moved the REST() into the async, otherwise I got `TypeError: (intermediate value).setToken(...) is not a function`

This will not make a difference.

Jiralite (approved):

monbrey:
> > * I moved the REST() into the async, otherwise I got `TypeError: (intermediate value).setToken(...) is not a function`
> 
> This will not make a difference.

Just to add, the reason you got this error is because you didn't have a semicolon `;` on the end of the previous line.

PR #1456, opened 2023-04-19
Archived Threads can be Locked.

Neodevils:
It has outdated information toast.

### DIFF
--- a/guide/popular-topics/threads.md
+++ b/guide/popular-topics/threads.md
@@ -90,8 +90,8 @@ await thread.setLocked(true); // locked
 await thread.setLocked(false); // unlocked
 ```
 
-::: warning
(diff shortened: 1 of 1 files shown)


Jiralite (commented):
I feel like the whole tip should be removed. There's no implication above this in the guide that archived threads cannot be locked, so there's no reason to doubt this?

Neodevils:
> I feel like the whole tip should be removed. There's no implication above this in the guide that archived threads cannot be locked, so there's no reason to doubt this?

Well, you are right. I changed it to info toaster cause there’s old developers would notice it that it has been updated on Discord API. But sure, do you want I remove it?

Jiralite (approved):

PR #1460, opened 2023-05-03
fix(reloading-commands): add categories to code samples

nanashi-03:
For the reloading commands part of the code, it failed to run if the commands were in categories as suggested previously by the guide thus the code couldn't get the commands through require. I added a category key to the exports object and appropriate code in the reload.js file to properly.


Idris1401:
fix the lint issues and after that it should be good imo

Jiralite (approved):

PR #1469, opened 2023-05-19
fix(cache-customization): fix sweeper filter

Squiddleton:
The `filter` for the cache sweeper example should be a function that returns another function, according to the library's typings.  This fixes the current example which was only a function that directly returns a boolean.


jaw0r3k:
it is not?

Squiddleton:
> it is not?

The current typing is:
`type GlobalSweepFilter<K, V> = () => ((value: V, key: K, collection: Collection<K, V>) => boolean) | null;`
As you can see, it can be either a function that returns `null` *or* a function that returns another function, the latter of which returns a boolean.

PR #1470, opened 2023-05-21
fix(cooldowns): delete expired cooldown reply

duanluan:
image


jaw0r3k:
Um it isn't a good idea
as even reserving bot won't delete it
also 10 minutes ago shows that end of the timeout was in the past

PR #1472, opened 2023-05-21
fix(reloading-commands): unable to find command files

duanluan:
`/reload command ping`

guide:
```
Error executing reload
Error: Cannot find module './ping.js'
Require stack:
- ……\commands\utility\reload.js
```

code-samples:
```
Error executing reload
Error: Cannot find module '../undefined/ping.js'
Require stack:
- ……\commands\utility\reload.js
```


jaw0r3k:
If command has `category` property right set, it should work

PR #1475, opened 2023-05-29
Update setting-up-a-bot-application.md

YadBro:
less letter l (L)


jaw0r3k (approved):

almostSouji (approved):

almostSouji:
Window "pane" is technically not incorrect to use here and this isn't really a typo. But I agree that in context "panel" is probably easier to understand and translate

PR #1485, opened 2023-06-24
Add new package to useful-packages.md

malezjaa:
- The package is simply useful and easy to use. Customization options are large and simple. (it's logging library)


jaw0r3k:
I don't think it is an useful package
Not much related to djs

malezjaa:
so? none of the packages are related to djs

PR #1486, opened 2023-07-13
Modifying example to search command subfolders

JonathanRJoyner:
This portion of the guide is confusing because it does not work with a previous section title ["Command Categories"](https://discordjs.guide/creating-your-bot/command-handling.html#command-categories) which has the user move commands in to subcommand folders.

The changes I've made allow for the full commands folder to be searched for a matching command. This will work no matter the path to the file inside the command folder.

### DIFF
--- a/guide/additional-features/reloading-commands.md
+++ b/guide/additional-features/reloading-commands.md
@@ -16,8 +16,8 @@ The reload command ideally should not be used by every user. You should deploy i
 
 ```js
 const { SlashCommandBuilder } = require('discord.js');
-const glob = require('glob');
-const path = require('path');
+const { glob } = require('glob');
+const { path } = require('path');
 
 
 module.exports = {
--- a/guide/additional-features/reloading-commands.md
+++ b/guide/additional-features/reloading-commands.md
@@ -14,6 +14,8 @@ This section assumes you followed the [Command Handling](/creating-your-bot/comm
 The reload command ideally should not be used by every user. You should deploy it as a guild command in a private guild.
 :::
 
+We will use the glob package to navigate our slash command file structure. This can be installed via npm by running `npm i glob` in the terminal.
+
 ```js
 const { SlashCommandBuilder } = require('discord.js');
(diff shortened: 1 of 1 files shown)


GodderE2D:
This would require users to download the glob package using npm. As this is not stated on the page, I would recommend adding a part for installing the package.

JonathanRJoyner:
Good point. Added a line about installing glob.

guide/additional-features/reloading-commands.md
```diff
@@ -14,8 +14,13 @@ This section assumes you followed the [Command Handling](/creating-your-bot/comm
 The reload command ideally should not be used by every user. You should deploy it as a guild command in a private guild.
 :::
 
+We will use the glob package to navigate our slash command file structure. This can be installed via npm by running `npm i glob` in the terminal.
```
megatank58, line 17:
```suggestion
You can use the glob package to navigate your slash command file structure. To install the package run `npm install glob` in the terminal.
```
wording and consistency https://github.com/discordjs/guide/blob/main/CONTRIBUTING.md#youyour-instead-of-weour

GodderE2D:
Wait, you're making this change to reflect the command categories page, right? So why should we use a different approach than that page? I'd say either use `fs` here, or also make the categories page use `glob`.

sdanialraza:
I don't think changes like these are needed especially when the new guide is in the works.

Jiralite:
> This portion of the guide is confusing because it does not work with a previous section title ["Command Categories"](https://discordjs.guide/creating-your-bot/command-handling.html#command-categories) which has the user move commands in to subcommand folders.

Just pointing out that this issue is resolved.

Jiralite:
Node.js is adding this [natively](https://nodejs.org/en/blog/announcements/v22-release-announce#glob-and-globsync), so it'd probably be a good idea to use this when it's stable. For now, keeping things as-is is likely best. Thank you for your contribution, nonetheless.

PR #1517, opened 2023-11-09
Update advanced-creation.md

SimonPlays15:
Added return statements to the SlashCommandOptions.

-> no return statements will cause an internal error

### DIFF
--- a/guide/slash-commands/advanced-creation.md
+++ b/guide/slash-commands/advanced-creation.md
@@ -19,7 +19,7 @@ const data = new SlashCommandBuilder()
 	.setName('echo')
 	.setDescription('Replies with your input!')
 	.addStringOption(option =>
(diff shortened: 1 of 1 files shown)


jaw0r3k:
That is just not how javascript works

almostSouji:
This was true, if we made a function body, denoted with `{}`
```js
() => {
 1 // 1 is not returned here, the function will evaluate as undefined, if called
}
```

However, here we use immediate return
```js
() => 1 // 1 is returned here
```

> The braces can only be omitted if the function directly returns an expression. If the body has additional lines of processing, the braces are required — and so is the return keyword. Arrow functions cannot guess what or when you want to return.
see more: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/Arrow_functions

PR #1524, opened 2023-11-21
fix(command-handling): Cleanup page

Jiralite:

- The utility folder is the only folder present when looping over command folders.
- Command categories are already present (per the point above). There is an entire repeated section duplicating code and explanation here.


code-samples/creating-your-bot/command-handling/index.js
```diff
@@ -23,16 +23,18 @@ for (const folder of commandFolders) {
 	}
 }
 
-client.once(Events.ClientReady, () => {
-	console.log('Ready!');
+client.once(Events.ClientReady, c => {
+	console.log(`Ready! Logged in as ${c.user.tag}`);
```
almostSouji, line 27: as said in the other PR i'm really not sure why we introduce this duplicate with this horrible naming convention

almostSouji (approved):

PR #1525, opened 2023-11-21
fix(command-deployment): Cleanup page

Jiralite:

- Important comment difference between guide and code sample regarding folders vs. files.


code-samples/creating-your-bot/command-deployment/index.js
```diff
@@ -23,16 +23,18 @@ for (const folder of commandFolders) {
 	}
 }
 
-client.once(Events.ClientReady, () => {
-	console.log('Ready!');
+client.once(Events.ClientReady, c => {
```
almostSouji, line 26: did we really not call that client across the board, `c`, really? that's unfortunate

monbrey: It's a tradeoff between shadowing the variable, and showing that the event emits the client instance. It becomes relevant when we do event handling as it's no longer shadowing, but isn't relevant at this point in the guide.

almostSouji: if it has no purpose I personally don't think we need it here and can instead introduce and explain it once it does make sense to have.
On the other hand, I can see the idea behind introducing the fact here, but if we go with that approach, let's find something nicer than `c` to establish verbose and understandable variable names as convention and best practice.

Jiralite: For TypeScript users, `c` here will be emitted as `Client<ready>` whereas the previously-defined `client` is of type `Client<boolean>`. This means you won't be able to access `client.application` even though it's definitely defined in the ready event, so you'll have to do a pointless `client.isReady()` check.

almostSouji: sounds good! everything is better than single characters as variable names and explaining why this is useful (and for who) is def. a good idea

Jiralite: I have opened #1528!

almostSouji (approved):

PR #1527, opened 2023-11-21
fix(additional-features): Cleanup pages

Jiralite:

- Preferred `const { cooldowns } = interaction.client;`.


almostSouji (approved):

Raigekijin:
events branch has been deleted during this commit, and part of multiple sections are now missing.

PR #1539, opened 2023-12-26
Small consistency error in code blocks

man-o-valor:
was written as both 60000 and 60_000 as the same line of code in different example blocks


jaw0r3k:
Well its more readable this way, you should change the other example to follor the `_` syntax

PR #1543, opened 2024-01-15
docs: Implicit Grant flow is deprecated danger warning

Rein1n:
Implicit grant flow is recommended by the guide for SPAs, however the IETF does not recommend using Implicit grant flow in any scenario, preferring the Authorization code grant.

### DIFF
--- a/guide/oauth2/README.md
+++ b/guide/oauth2/README.md
@@ -71,6 +71,10 @@ The `identify` scope will allow your application to get basic user information f
 
 ### Implicit grant flow
 
+::: danger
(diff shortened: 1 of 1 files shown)


monbrey:
Couple of notes on this. Personally, I'd rather just rework the guide not to use implicit grant at all. In many other areas of the guide, we moved away from showing people a "bad but easy" way to do things, then changing it all in the next section. Just show the right way from the start.

Otherwise if we do want to stick with a warning, using `[here]` as the masked text on a link is poor accessibility for screen readers. Prefer something descriptive like `for more information check [the OAuth2 RFC](link)`

Rein1n:
I can change the pr to remove implicit grant completely if that would be preferable.

If we decide to keep implicit grant, I'll change the masked text to make it more clear, thanks for the heads up on that.

PR #1570, opened 2024-07-16
Add clarity to clearly state that the ‘first’ audit log is the newest

jackfrojpg:
This PR adds a small bit of extra text to make it clear to readers that the ‘first’ audit log is the most recent. Readers may be confused without this addition as they might believe that the first log is the first created, and therefore the oldest, however this isn’t the case and should be made clear.

### DIFF
--- a/guide/popular-topics/audit-logs.md
+++ b/guide/popular-topics/audit-logs.md
@@ -32,7 +32,7 @@ const fetchedLogs = await guild.fetchAuditLogs({
 const firstEntry = fetchedLogs.entries.first();
 ```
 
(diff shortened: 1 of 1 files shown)


almostSouji:
This is an upstream (discord) implementation detail and not always the case, since:

> The returned list of audit log entries is ordered based on whether you use before or after. When using before, the list is ordered by the audit log entry ID descending (newer entries first). If after is used, the list is reversed and appears in ascending order (older entries first). Omitting both before and after defaults to before the current timestamp and will show the most recent entries in descending order by ID, the opposite can be achieved using after=0 (showing oldest entries).

<https://discord.com/developers/docs/resources/audit-log#get-guild-audit-log>

PR #1575, opened 2024-09-03
chore: Update required Node version texts

maxtull:
Update the required Node version from v16.11.0 to v18 (Discord.JS v14.16.0 minimum).

### DIFF
--- a/guide/additional-info/changes-in-v14.md
+++ b/guide/additional-info/changes-in-v14.md
@@ -2,7 +2,7 @@
 
 ## Before you start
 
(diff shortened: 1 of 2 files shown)


Jiralite (changes requested):
Let's say Node.js LTS. It's likely more maintainable than changing versions as we have always done. The guide is typically read by new-to-discord.js people, so advertising LTS is what we want (16 is too old).

almostSouji (approved):

Jiralite (approved):

PR #1598, opened 2024-12-19
Update embeds.md

shantismurf:
- Added information about Discord's text sanitization process to the "Embed limits" section.
- Corrected the url linking to embed limit API (was pointing to channel, not message)


Jiralite:
Thank you for the pull request!

It is too general to document escaping markdown to show properly. It is also not specific to embeds. I do not believe it is warranted enough to document this.

The quote you provided is almost 5 years out of date—it has changed since allowed mentions. This is not an embed limitation, but rather, general guidance around message content. Maybe notes is a better place to put it.

PR #1601, opened 2025-01-07
Altered code samples to reflect previously implemented individual event files

Pogoretskiy777:

This pull request updates the `additional-features/code-samples` directory to include the individual event listener files that were added in previous steps of the guide. This change makes reading the reference code more consistent, clearer, and easier for the user who has followed the previous steps in implementing individual event listener files in case they want to cross-reference their work.


almostSouji:
sorry for the late response and thanks for the contribution!
we are moving forward with a "legacy" version of the guide as-is and likely approach the guide differently with regards to specific handlers so this will unfortunately not land.

PR #1617, opened 2025-03-24
refactor(permissions): Note hiding of commands when setting default member permissions

miguelmikkey:
### updated `slash-commands/permissions.md`
Added a note regarding command visibility when using `setDefaultMemberPermissions` as adding permissions to a slash command will *hide* the command from users when looking for commands.

### DIFF
--- a/guide/slash-commands/permissions.md
+++ b/guide/slash-commands/permissions.md
@@ -10,7 +10,7 @@ It is not possible to prevent users with Administrator permissions from using an
 
 ## Member permissions
 
-You can use <DocsLink section="builders" path="SlashCommandBuilder:Class#setDefaultMemberPermissions" type="method" /> to set the default permissions required for a member to run the command. Setting it to `0` will prohibit anyone in a guild from using the command unless a specific overwrite is configured or the user has the Administrator permission flag. This also controls the command's visibility, ensuring that only users with the required permissions can access it.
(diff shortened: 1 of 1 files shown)


guide/slash-commands/permissions.md
```diff
@@ -10,7 +10,7 @@ It is not possible to prevent users with Administrator permissions from using an
 
 ## Member permissions
 
-You can use <DocsLink section="builders" path="SlashCommandBuilder:Class#setDefaultMemberPermissions" type="method" /> to set the default permissions required for a member to run the command. Setting it to `0` will prohibit anyone in a guild from using the command unless a specific overwrite is configured or the user has the Administrator permission flag.
+You can use <DocsLink section="builders" path="SlashCommandBuilder:Class#setDefaultMemberPermissions" type="method" /> to set the default permissions required for a member to run the command. Setting it to `0` will prohibit anyone in a guild from using the command unless a specific overwrite is configured or the user has the Administrator permission flag. This also controls the command's visibility, ensuring that only users with the required permissions can access it.
```
Jiralite, line 13: Seems adding this word basically states the same intention. What do you think?

```suggestion
You can use <DocsLink section="builders" path="SlashCommandBuilder:Class#setDefaultMemberPermissions" type="method" /> to set the default permissions required for a member to run the command. Setting it to `0` will hide and prohibit anyone in a guild from using the command unless a specific overwrite is configured or the user has the Administrator permission flag.
```

miguelmikkey: It does look cleaner, but i'm not sure if the message is the same. I tried to emphasize it just in case the user following the guide is a beginner and needs the extra information, but yeah, it looks fine to me.

almostSouji:
applied to the legacy portion of the next guide iteration

PR #1618, opened 2025-04-01
fix(command-handling): Add `MessageFlags` to base template

babofitos:

The code uses MessageFlags later on, but it's never imported in the example code.


heavylightdecomp:
This seems to be pretty important because it relates directly to code correctness. 

Looks good to me

Jiralite (approved):

PR #1629, opened 2025-06-15
Components refactor

almostSouji:

- Phrasing improvements
- Switch from "components v2" to "display components" (v2 is never mentioned apart from the flag name in upstream docs)
- Switch from "message component" to "interactive component" (components v2 could still be understood as "message components" as they are also part of the message, so making a distinction here seems appropriate)
- Be more explicit about warning regarding accidental notifications caused by difference to embed behavior
- Move caveats of using cv2 to the top as warning admonition

TODO:

- [ ] After merge: link in FAQ post in discord


PakjeBecel:
I used to reference the display components as "Components V2", since that is what basically everyone is calling them.

People might search for "How to use components v2 discord bot" rather than "How to use display components discord bot" as well, so in my opinion it should be called "Components V2" throughout the article.

People might think "Huh I ain't looking for display components, I need a guide for Components V2".

almostSouji:
i understand the idea and arguably better discovery! we do stick quite strictly to the official terminology across the library and try to stay consistent with that idea on the guide.

"components v2" was more of a working title and isn't really reflected apart from blogs/announcements and the flag name itself. in upstream it's always just referenced as components, which imo makes more sense. cv2 isn't really a replacement/new version of already existing components (buttons, selects, and action rows), which the "v2" suffix suggests, but an extension of the component system to move away from the fairly limited embed constraints to a more extendable and modular system. 

(maybe a new wave of components will be "cv3" - we will see ^^)

if you have other feedback, feel free to leave a review and thanks again for the original PR!

PakjeBecel:
Ah, that indeed makes a lot more sense to me now. Thanks for explaining! In this way I agree with the changes you have made.

scottlangille:
Looks like this PR broke the entire message-components section of the docs.

https://discordjs.guide/message-components (404)

Jiralite:
Where did you get that from?

almostSouji:
thanks! that makes sense - i think we need to refresh our search provider's crawl (algolia) - the display components section wont't surface at all either, since the search still has the old state.

almostSouji:
thanks again for the reminder - the guide has now been reindexed and search should work and not result in 404s now
(granted, with the updated wording, of course)

PR #1630, opened 2025-08-22
update voice guide encryption dependencies

jNullj:
Seems that we might need a few updates here for the voice guides.

- There is support for the DAVE protocol, some users might be interested and it should show as an option.
- Same goes for `generateDependencyReport` output in the guide.
- Added warning about `@snazzah/davey` being a dev dependency, this could confuse some devs.
- We now support `node:crypto`, updated the section about the requirement to install encryption library and a tip to verify the system support.

Support for DAVE protocol was added here https://github.com/discordjs/discord.js/issues/10735
Please note that I discovered this while I got repeated crashes due to discord not downgrading my connection to a non e2ee connection!
Discord mention this in a past blog post [here](https://discord.com/developers/docs/change-log#voice-endtoend-encryption-dave-protocol) and it seems like this is about to be a requirement.

### DIFF
--- a/guide/voice/README.md
+++ b/guide/voice/README.md
@@ -6,7 +6,11 @@
 
 ### Barebones
 
-To add voice functionality to your discord.js bot, you will need the `@discordjs/voice` package, as well as one of the encryption packages listed below. For example: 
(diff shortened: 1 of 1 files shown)


Jiralite:
Hey, you should update the guide in the monorepo (Same repository where you made https://github.com/discordjs/discord.js/pull/11061 at).

